PR #16, opened 2015-03-23
Compilation fixes to convtool and mksprite

sp1187:


DragonMinded:
Wouldn't it be better to fix the warnings instead of disable converting warnings into errors?

sp1187:
Mksprite fails to compile due to a missing #include <string.h>, which is easy to fix. Convtool fails due to an unused variable ret, and I thought the easiest way to remove the warning was to remove -Werror (which I admit was probably too drastic). The best solution would be actual error checking, but -Wno-unused-variable could also work. Personally, I can probably just patch the code locally myself until you decide which solution is the best, so I will close the issue.

PR #17, opened 2015-03-23
Improved directory structure by moving all MIPS libraries/includes

sp1187:
to $N64_INST/mips64-elf/{lib,include}.
Old project Makefiles incompatible, but can be trivially fixed
(see the "examples" folder).

This causes less issues when installing to /usr, since
MIPS/libdragon-related files will be installed in /usr/{lib,include} otherwise, 
where it can conflict with native libraries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 ROOTDIR = $(N64_INST)
-CFLAGS = -std=gnu99 -O2 -G0 -Wall -Werror -mtune=vr4300 -march=vr4300 -I$(CURDIR)/include -I$(ROOTDIR)/include -I$(ROOTDIR)/mips64-elf/include
+CFLAGS = -std=gnu99 -O2 -G0 -Wall -Werror -mtune=vr4300 -march=vr4300 -I$(CURDIR)/include -I$(ROOTDIR)/mips64-elf/include
 ASFLAGS = -mtune=vr4300 -march=vr4300
 N64PREFIX = $(N64_INST)/bin/mips64-elf-
 INSTALLDIR = $(N64_INST)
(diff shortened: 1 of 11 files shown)


DragonMinded:
Not sure why this is needed? The idea is that it shouldn't be installed into /usr, but /usr/mips64 or a similar directory. This leaves bin in the same place while moving lib/include which isn't normally how toolchains get installed as far as I know.

sp1187:
OK, I will try to explain a bit better.

Arch Linux usually installs the prefixed toolchain binaries in /usr/bin and architecture-specific libraries and includes in /usr/$ARCH (mips64-elf here). However, the default for the build script is to make another superfluous directory level so that the binaries get into /usr/mips64-elf/bin (not in $PATH) and architecture-specific libraries get into /usr/mips64-elf/mips64-elf/. Also, unlike Newlib, the libdragon files do not install into the architecture-specific directory, so if you try to install to /usr, you will get N64 libraries as the same directory as standard native libraries. I actually had weird linking issues when native and N64 files were on the same directory, which was the main reason for the patch. It also sort of makes sense if both newlib and libdragon stay in the same directories.

sp1187:
A simple script for converting an already existing installation to the proposed directory structure.

```sh
#!/bin/bash

set -u

mv -iv ${N64_INST}/lib/{libdragon.a,libdragonsys.a,libdragonpp.a,n64ld.x,n64ld_cpp.x,n64ld_exp_cpp.x,header} ${N64_INST}/mips64-elf/lib/
mv -iv ${N64_INST}/include/{n64sys.h,interrupt.h,dma.h,dragonfs.h,audio.h,display.h,console.h,mempak.h,controller.h,graphics.h,rdp.h,timer.h,exception.h,system.h,dir.h,libdragon.h} ${N64_INST}/mips64-elf/include/
```

For makefiles, you generally only need to change HEADERPATH to $(ROOTDIR)/mips64-elf/lib and remove all occurrences of -L$(ROOTDIR)/lib and -I$(ROOTDIR)/include.

sp1187:
I would like to discuss this pull request again.
I have build scripts (AUR) for Arch Linux ongoing which would simplify the installation significantly in that distribution, however this request needs to be accepted before the scripts can be used by others (unless I maintain parallel incompatible forks or patches).
Similar scripts or builds can then be made for other Linux distributions of course.

I hope I have established already that installing GCC cross-compilers in /usr is indeed common when installing system-wide and that /usr/mips64-elf would then be the correct place to put libraries, which is basically why Newlib does so.

Upgrading an old installation is fairly trivial, it can be done by simply move all files except the gcc folder from $N64_INST/{include,lib} to $N64_INST/mips64-elf/{include,lib} or by using the script in my previous post. Just reinstalling libdragon when upgrading the installation is another option. The project Makefile changes are also trivial.

DragonMinded:
I still don't understand why all this churn is necessary. The N64_INST should be set to /usr/mips64-elf or something similar, not /usr. If you set to /usr, of course the lib will go to /usr/lib and the includes will go to /usr/include. If we change this to assume /$ROOT/mips64-elf/ then we have to go back and change all the headers and include scripts (which this has done) instead of just setting the N64_INST directory to /usr/mips64-elf. I just don't see the problem in how it is, given that it already namespaces and you can control with one variable where it is installed.

sp1187:
The problem with N64_INST=/usr/mips64-elf is that the GCC toolchain will end up in the wrong place compared to the standard (and IMHO more elegant) setup. The mips64-elf-* binaries should be in /usr/bin (no $PATH changes required if you want to test the commands directly) and target-dependent files should be in /usr/mips64-elf, compare with official Linux distribution packages: [(1)](https://www.archlinux.org/packages/community/x86_64/arm-none-eabi-gcc/files), [(2)](https://www.archlinux.org/packages/community/x86_64/avr-gcc/files), [(3)](https://packages.debian.org/jessie/amd64/gcc-avr/filelist), [(4)](https://packages.debian.org/jessie/amd64/gcc-arm-none-eabi/filelist). The current convention makes this setup impossible while the proposed convention will support both ways.

I understand that updating the paths in old Makefiles and scripts may be somewhat painful, but the current convention seems to be simply incorrect about the path conventions and the compatibility fixes are trivial enough that I feel fixing this should be prioritized. Feel free to close this issue if you still disagree.

DragonMinded:
That makes a lot more sense.

PR #19, opened 2015-04-15
Portability fixes

parasyte:
This PR ensures the toolchain compiles and installs on Mac OS X. Also includes some niceties like adding a basic .gitignore to the tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,20 @@
+## Build files
+*.a
+build/
+build_gcc/
+tools/chksum64
+tools/dumpdfs/dumpdfs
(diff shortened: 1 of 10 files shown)


DragonMinded:
This all looks good. Have you been able to test it on a linux install to ensure no regressions?

parasyte:
I haven't. :confounded: But it is easy enough to test. I'll see how it goes, and submit additional patches if necessary.

DragonMinded:
I should accept this and then if there are breaks, we can accept fixes to it. Sorry for the delay.

PR #25, opened 2016-09-03
Fix bss-related compile error in inthandler.S with newer GAS versions

sp1187:
Without this fix, newer versions of GAS will abort the compilation with the message `.../src/inthandler.S:238: Error: attempt to store non-zero value in section `.bss'`


sp1187:
Are there any issues with this and my other pull request? I have tried the changes with most of the example programs, and they work just like before.

PR #27, opened 2016-09-26
Add support for callback-driven audio

mic-:


src/audio.c
```diff
@@ -147,7 +152,7 @@ static void audio_callback()
     {
         /* check if next buffer is full */
         int next = (now_playing + 1) % _num_buf;
-        if (!(buf_full & (1<<next)))
+        if ((!(buf_full & (1<<next))) && !_fill_buffer_callback)
```
DragonMinded: At this point, you verify that we don't have a callback, but what if somebody calls pause() without one? It looks like in that case the code below will set a pause callback anyway. Is that intended behavior?

src/audio.c
```diff
@@ -185,8 +194,10 @@ static void audio_callback()
  *            The frequency in Hz to play back samples at
  * @param[in] numbuffers
  *            The number of buffers to allocate internally
+ * @param[in] fill_buffer_callback
+ *            A function to be called when more sample data is needed
  */
-void audio_init(const int frequency, int numbuffers)
+void audio_init(const int frequency, int numbuffers, audio_fill_buffer_callback fill_buffer_callback)
```
DragonMinded: I don't see any updates to the examples. I forget, did any of the examples show off audio? If so, they won't compile anymore.

src/audio.c
mic-: True. I should probably add `&& _fill_buffer_callback` to the first `if` in `audio_pause`.

src/audio.c
mic-: There's the dfsdemo. So I guess I should update its call to `audio_init`. Or possibly create a new function called `audio_set_buffer_callback` or something like that. This is just the way I implemented it back when I originally made these changes, which was a while ago (when I had begun working on this: https://www.youtube.com/watch?v=EqrgPTFNXAA).

DragonMinded: Ooh nice, I like!

Possibly setting the fill buffer in another function instead of the init would be backwards compatible and also less confusing for somebody trying to use, so I like that idea.

src/audio.c
```diff
@@ -258,6 +256,13 @@ void audio_init(const int frequency, int numbuffers, audio_fill_buffer_callback
     _paused = false;
 }
 
+void audio_set_buffer_callback(audio_fill_buffer_callback fill_buffer_callback)
+{
+    disable_interrupts();
+    _fill_buffer_callback = _orig_fill_buffer_callback = fill_buffer_callback;
```
DragonMinded: Hate to nitpick, but setting this when paused would mean we would unpause, right? Probably low-pri but still nice to be correct.

PR #28, opened 2016-12-08
Change example Makefiles to compile the ROM in Z64 format

sp1187:
Makes the examples a little easier to test with CEN64.


sp1187:
I would appreciate some kind of response, especially considering that the PR has been up for nine months and other PRs have been merged during that time.

vieux:
@sp1187 why not keep both options available ?

DragonMinded:
Honestly this looks like churn for no good reason, which is why I wasn't paying attention to it. You've just renamed every output, but you could definitely write a local script that renames for you, or keep a branch if that makes a lot of difference.

vieux:
it also removed `-b` to make proper z64s

DragonMinded:
I didn't see that. I like the option to spit out both that was proposed by vieux. Could you make the all/copy targets require a v64 and z64 and then build both?

PR #31, opened 2017-09-04
Add in mempak header checksum during validation.

a-busman:


vieux:
I get compile error because the function is used before it's declaration.

https://github.com/vieux/libdragon/commit/1457713d631f68da8d63e5cb8ad3b0d8083dbd34 fixes the issue for me.

PR #40, opened 2018-07-21
RSP improvements

anacierdem:
Improved RSP syntax.
Added (almost?) all missing instructions.
Updated readme.


DragonMinded:
I wonder if we should abstract out instructions for compiling each of the demos, because it looks like a lot of copypasta. This looks good though!

PR #41, opened 2018-12-03
High resolutions, alignment

clbr:
The 64-byte req comes from docs, two new high-res modes tested on NTSC hw.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -57,14 +57,14 @@
 #define UNCACHED_ADDR(x)    ((void *)(((uint32_t)(x)) | 0xA0000000))
 
 /**
- * @brief Align a memory address to 16 byte offset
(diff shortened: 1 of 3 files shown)


src/display.c
```diff
@@ -352,8 +352,8 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
     {
         /* Set parameters necessary for drawing */
         /* Grab a location to render to */
-        buffer[i] = malloc( __width * __height * __bitdepth + 15 );
-        __safe_buffer[i] = ALIGN_16BYTE( UNCACHED_ADDR( buffer[i] ) );
+        buffer[i] = malloc( __width * __height * __bitdepth + 63 );
```
DragonMinded: This and the 63 above should probably be a constant, but whatever.

PR #58, opened 2019-05-07
Enable single entry C++ support.

### LINKED ISSUE #8: LD and startup.s files different for C and C++
There should be no reason these are different.  Need to look into merging C and C++ support back into one coherent package.

anacierdem:
Fixes #8


src/entrypoint.S
```diff
@@ -39,22 +39,6 @@ data_init:
 	jal data_cache_hit_writeback_invalidate
 	nop
 
-	/* fill .bss with 0s */
```
DragonMinded: What handles this for non-cpp files now? Is bss zero-filling considered unnecessary since C doesn't guarantee any particular value in a statically allocated variable?

src/entrypoint.S
```diff
@@ -39,6 +39,22 @@ data_init:
 	jal data_cache_hit_writeback_invalidate
 	nop
 
+	/* fill .bss with 0s */
```
DragonMinded: Okay, I see its back here.

n64.ld
```diff
@@ -28,8 +28,15 @@ MEMORY
 }
 
 SECTIONS {
+   /* Start address of code is 1K up in uncached, unmapped RAM.  We have
```
DragonMinded: Good documentation!

Makefile
```diff
@@ -1,5 +1,5 @@
 ROOTDIR = $(N64_INST)
-CFLAGS = -std=gnu99 -O2 -Wall -Werror -mtune=vr4300 -march=vr4300 -I$(CURDIR)/include -I$(ROOTDIR)/mips64-elf/include
+CFLAGS = -std=gnu99 -O2 -Wall -Werror -DLIBDRAGON_VERSION_MAJOR=$(LIBDRAGON_VERSION_MAJOR) -DLIBDRAGON_VERSION_MINOR=$(LIBDRAGON_VERSION_MINOR) -DLIBDRAGON_VERSION_REVISION=$(LIBDRAGON_VERSION_REVISION) -mtune=vr4300 -march=vr4300 -I$(CURDIR)/include -I$(ROOTDIR)/mips64-elf/include
```
DragonMinded: Ooh, good idea, I like this!

n64.ld
anacierdem: This was already present in one of linker scripts indeed.

Makefile
anacierdem: I had some version headaches when testing these all out, so here you go :)

PR #60, opened 2019-06-10
fixed 16 bit accesses to 8 bit buffer in __validate_header

jago85:
The mputest-example did't work for me. I've found this to be the reason. 
The base code is accessing the sector pointer as an uint8_t and is then comparing the data to the uint16_t checksum. Don't know how this could have been working.

This could be done nicer but I did a quick fix.

Greetings,
jago85

### DIFF
--- a/src/mempak.c
+++ b/src/mempak.c
@@ -159,16 +159,16 @@ static int __validate_header( uint8_t *sector )
     /* Check 4 checksums of copied header data */
     current_block = 0x20;
     checksum = __get_header_checksum((uint16_t *)&sector[current_block]);
-    if( (checksum != *(uint16_t *)(&sector[current_block + 0x1C])) && (checksum != 0xFFF2 - *(uint16_t *)(&sector[current_block + 0x1E])) ) { return -1; }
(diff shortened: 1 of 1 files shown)


jago85:
Maybe the && in
`if( (checksum != (uint16_t)(sector[current_block + 0x1C])) && (checksum != 0xFFF2 - (uint16_t)(sector[current_block + 0x1E])) ) { return -1; }`
should also be || instead depending on what exactly you want to check (check both or only one).

DragonMinded:
You're possibly right on the || versus &&. Does it pass with || as well?

jago85:
Yes, just tried 3 paks and they are working with the ||. Do you want to use this version? Then I would push the changes and make a new pull request.

I wonder if there is a more elegant way to check the checksum (without having to compare two values), but I have not found one yet.

DragonMinded:
That sounds good. You should be able to add to this pull request?

jago85:
Done. The new commit has been added here automatically. Cool, did not know that because I have not worked with pull requests yet.

DragonMinded:
Yeah, it rules actually, super nice flow.

PR #72, opened 2019-10-27
fix examples following the mirror changes

vieux:
Follow up of #70 

Sorry about this @anacierdem @DragonMinded


DragonMinded:
I wondered about those actually... ;P

PR #74, opened 2019-10-29
New release 1.3.15

anacierdem:
Bump version
Add changelog
NPM improvements.
`libdragon install` now skips CI shecks.
Fix non-volatile asm - see https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Extended-Asm.html

> An asm instruction without any output operands will be treated identically to a volatile asm instruction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,6 @@
+# Change Log
+
+## [1.3.9] - 2019-27-10
+### Added
+
+- Add texture mirroring. https://github.com/DragonMinded/libdragon/commit/00a6cc8e6d136cf2578a50320f6ff0814dfb6657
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,17 @@
...
+### Changed
+
+- Reduce response time for NPM commands.
+- Remove unnecessary console statements and double logs.
+
+## [1.3.11] - 2019-29-10
+### Fixed
+
+- Fix problem with wait ticks. They do not lock the system now.
+
```
vieux: @anacierdem what went into 1.3.10 ?

anacierdem: I skipped that one due to a mistake :) I'm a little impatient when it comes to releases, trying to improve myself on that regard :D

PR #82, opened 2019-11-18
Simpler ucode linking

anacierdem:


anacierdem:
This will break memory location naming on the other hand...

PR #93, opened 2020-04-07
Added WIP Textured Triangle rendering

twixn:
A work in progress expansion on the filled triangle calls to add textured triangles.
Affine texturing works, however perspective is still WIP.

To Test:
        rdp_sync( SYNC_PIPE );
        rdp_set_tri_prim_color( 0xffffffff );
        rdp_texture_cycle(_1CYCLE,ATOMIC_PRIM | SAMPLE_TYPE | IMAGE_READ_EN);
        rdp_load_texture( 0, 0, MIRROR_DISABLED, mario );
        rdp_draw_textured_triangle(0, x1, y1, w1, s1, t1, 
                                    x2, y2, w2, s2, t2, 
                                    x3, y3, w3, s3, t3 );
        rdp_sync( SYNC_PIPE );


twixn:
Sure thing. Here is a picture of it running.
image

Very PSX-like atm. But it's a start :)

I'll be cleaning up and committing the rest of my 3D pipeline code as my next steps.
The ultimate goal is to have some thing like 'void graphics_draw_model( display_context_t disp, matrix4 transform , model_t *model );'

twixn:
Just added an example project for it. A funky spinning cube.
Just no depth buffer, or perspective correct texturing.

twixn:
> Maybe we can also checkout https://github.com/datenwolf/linmath.h what do you think?
> Or this; https://github.com/felselva/mathc

Sure, I can give either a try. Do you all have a preference on how external libraries are pulled in? I might have to checkout and just add the source files to the project, as I'll need to adjust their perspective projection matrix to suit.

twixn:
Hrmmmm, I'm starting to have doubts. I've given linmaths.h a try and it looks unstable on the n64.

One example is the matrix multiplication method: 'void mat4x4_mul(mat4x4 M, mat4x4 a, mat4x4 b)'. Which crashes when it is called, even if I comment out all of it's contents.

If I was going to guess, having 3*16float objects in the stack being passed as parameters is too much for the n64?

anacierdem:
Those were just libraries I was keeping around in case they become handy. It is not a showstopper at this time, we can just replace them later on if needed. On libdragon docker Ive implemented some kind of a dependency management but it requires a makefile present. I have no idea why they are not working out of the box and dont think its a stack problem. Let’s just not bother if it is complicating things... Keep up the good work 👍🏼

PR #95, opened 2020-04-24
Fix exception handler

### LINKED ISSUE #94: Exception in exception handler :)
`e->info = __get_exception_name((uint32_t)e->regs->gpr[30]);` in the `__fetch_regs` method is causing another exception, resulting in an exception loop.

This only happens when `(uint32_t)e->regs->gpr[30]` is used, so propably there is a problem with addresses.

anacierdem:
Fix #94

### DIFF
--- a/include/exception.h
+++ b/include/exception.h
@@ -37,6 +37,8 @@ typedef volatile struct
 	volatile uint64_t gpr[32];
     /** @brief SR */
 	volatile uint32_t sr;
+    /** @brief CR */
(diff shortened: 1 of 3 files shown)


rasky:
This still looks OK to me. For this kind of PRs, the only way is to really test it on real hardware, so I guess you did and it works :)

anacierdem:
C and assembly is not something I’m very comfortable with, especially if it is a public code and on a critical part of it 😊 I will merge this soon though.

PR #100, opened 2020-06-10
Fix makedfs tool when built for Windows

networkfusion:
makedfs doesnt work when built on Windows due to file operations adding a carriage return when on windows. The fix is to open the files in binary mode (e.g. rb instead of r and wb instead of w). There also issues with endianness. and use of dirent.h for which the user will need to add something like `curl -L https://raw.githubusercontent.com/tronkko/dirent/master/include/dirent.h --output ./include/dirent.h` to their build script (unless it can be added here)

This PR fixes DragonMinded#30

### DIFF
--- a/tools/mkdfs/mkdfs.c
+++ b/tools/mkdfs/mkdfs.c
@@ -1,5 +1,5 @@
 #include <stdio.h>
-#include <malloc.h>
+#include <stdlib.h>
 #include <sys/stat.h>
 #include <string.h>
 #include <errno.h>
--- a/tools/mkdfs/mkdfs.c
+++ b/tools/mkdfs/mkdfs.c
@@ -1,17 +1,27 @@
 #include <stdio.h>
-#include <stdlib.h>
-#include <dirent.h>
+#include <malloc.h>
 #include <sys/stat.h>
 #include <string.h>
 #include <errno.h>
 #include <stdint.h>
+
 #include <sys/types.h>
+#ifndef _MSC_VER
+#include <dirent.h>
 #include <sys/param.h>
+#else
+#include "dirent.h"
+#endif // !_MSC_VER
+
 #include "dragonfs.h"
 #include "dfsinternal.h"
 
+#ifndef _MSC_VER
 #if BYTE_ORDER == BIG_ENDIAN
 #define SWAPLONG(i) (i)
+#else
+#define SWAPLONG(i) (((uint32_t)(i & 0xFF000000) >> 24) | ((uint32_t)(i & 0x00FF0000) >>  8) | ((uint32_t)(i & 0x0000FF00) <<  8) | ((uint32_t)(i & 0x000000FF) << 24))
+#endif
 #else
 #define SWAPLONG(i) (((uint32_t)(i & 0xFF000000) >> 24) | ((uint32_t)(i & 0x00FF0000) >>  8) | ((uint32_t)(i & 0x0000FF00) <<  8) | ((uint32_t)(i & 0x000000FF) << 24))
 #endif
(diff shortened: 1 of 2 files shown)


tools/mkdfs/mkdfs.c
```diff
@@ -1,20 +1,30 @@
 #include <stdio.h>
-#include <stdlib.h>
-#include <dirent.h>
+#include <malloc.h>
```
anacierdem: This is reverting https://github.com/DragonMinded/libdragon/pull/36 maybe we can also do this conditionally?

tools/mkdfs/mkdfs.c
```diff
@@ -1,20 +1,30 @@
...
 #include <string.h>
 #include <errno.h>
 #include <stdint.h>
+
 #include <sys/types.h>
+#ifndef _MSC_VER
+#include <dirent.h>
 #include <sys/param.h>
+#else
+#include "dirent.h"
```
anacierdem, line 13: How is this resolved on vs? If you have a local file, maybe we should also include that.

tools/mkdfs/mkdfs.c
```diff
@@ -1,20 +1,30 @@
...
 #include "dragonfs.h"
 #include "dfsinternal.h"
 
+#ifndef _MSC_VER
 #if BYTE_ORDER == BIG_ENDIAN
 #define SWAPLONG(i) (i)
 #else
 #define SWAPLONG(i) (((uint32_t)(i & 0xFF000000) >> 24) | ((uint32_t)(i & 0x00FF0000) >>  8) | ((uint32_t)(i & 0x0000FF00) <<  8) | ((uint32_t)(i & 0x000000FF) << 24))
 #endif
+#else
```
anacierdem, line 25: Can't see the difference here, am I blind :)

tools/mkdfs/mkdfs.c
networkfusion: Windows might work with stdlib, but this change was made before #36 and I seemed to have missed it. I will have to test.

tools/mkdfs/mkdfs.c
networkfusion: As per the PR description...

tools/mkdfs/mkdfs.c
networkfusion: MSVC does not recognise the macro BYTE_ORDER or BIG_ENDIAN and there is no simple way to detect otherwise. Since all Windows systems are little endian, this was the easiest option.

networkfusion: for reference: https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html

https://stackoverflow.com/questions/8978935/detecting-endianness

image

tools/mkdfs/mkdfs.c
anacierdem: Ah now seen it sorry about that. Still we may need to include that info somewhere visible

tools/mkdfs/mkdfs.c
anacierdem: This makes sense now :)

anacierdem:
Maybe we can also add -Wall so that the CI fails when someone introduce a warning?

tools/mkdfs/mkdfs.c
networkfusion: I agree, would a readme in the folder be acceptable, including the compiler command for MSVC i.e.       
call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
      cl .\tools\mkdfs\mkdfs.c -I .\include\

anacierdem: I think that’ll be fine

networkfusion: Readme added. I guess I should check that malloc is still needed, so please hold off on committing until I am able to test...

networkfusion:
Builds as expected: 
image

networkfusion:
With MSVC  using /Wall:

image

PR #101, opened 2020-06-10
Fix tool compiler warnings

networkfusion:


tools/n64tool.c
```diff
@@ -23,7 +23,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define WRITE_SIZE	(1024*1024)
+#define WRITE_SIZE	1024
```
anacierdem, line 26: Is there a reason to reduce this?

tools/n64tool.c
```diff
@@ -160,21 +159,28 @@ int copy_file(FILE *dest, char *file, int byte_swap)
...
+	
 	int i;
-	while (amount > 0) {
-		int sz = amount;
-		if (sz > sizeof(zero))
-			sz = sizeof(zero);
-		fwrite(zero, 1, sz, dest);
-		amount -= sz;
+	
+	for(i = 0; i < (amount >> 2); i++)
```
anacierdem, line 176: This is reverting @rasky 's changes at https://github.com/DragonMinded/libdragon/commit/3c45c51277fab510daeacc4f336a07bb00713b10

tools/n64tool.c
```diff
@@ -71,7 +70,7 @@ uint32_t get_file_size(FILE *fp)
 
 int swap_bytes(uint8_t *buffer, int size)
 {
-	if((size & 1) == 0)
+	if(size & 1 == 0)
```
anacierdem, line 73: paranthesis would be nice to keep if there is not a specific reason?

tools/mksprite/convtool.c
```diff
@@ -13,25 +13,31 @@ int main( int argc, char *argv[] )
...
 
+    #ifndef _MSC_VER
+    int ret __attribute__ ((unused));
+    #else
+    int ret;
+    #endif
+
     /* Read in old width */
-    fread( &val, sizeof( val ), 1, stdin );
+    ret = fread( &val, sizeof( val ), 1, stdin );
```
anacierdem, line 23: If we are assigning ret, would be nice to use it instead of doing it only to make compiler happy (if that's the case here)

tools/n64tool.c
rasky: This is also reverting my change, and thus re-introducing a bug (the parentheses change the operator evaluation order).

tools/n64tool.c
rasky: This is also revering my change, and thus introducing a performance regression.

rasky:
I think this PR has been wrongly rebased as it ends up reverting my recent commits.

anacierdem:
Seems like it..

networkfusion:
Yes, I am sorry about that. Done this in a rush on a new fork (as my old one isn't and based off of conker64 version). I will go through the changes thoroughly when I get a chance.

BTW, libdragon and tools should be build with -Wall anyway (so that these can be spotted before they are an issue...

networkfusion:
Will reopen after changes made.

PR #109, opened 2020-10-06
n64tool: fix bug in detection of unaligned image sizes

jago85:
The alignment check in swap_bytes() was still incorrect. This should fix it.


DragonMinded:
LGTM, thanks for the fix, but I can't merge due to conflicts. Can you rebase on top of trunk and force push?

jago85:
Ok, is it better now?

DragonMinded:
Yup!

PR #115, opened 2021-01-12
dfs: fix performance of dfs_seek to be constant-time.

### LINKED ISSUE #85: Support files > 16M (or at least abort when creating the filesystem)
DFS can encode only file sizes up to 2^24 because it uses a `uint32_t flags` that uses only 24 bits to encode file size, and 8 bit for flags:

https://github.com/DragonMinded/libdragon/blob/122a61546bac17f36537b1b73d2ea8758e188f3f/tools/mkdfs/mkdfs.c#L216

I think size and flags could safely be splitted into two different variables, considering that the same structure contains a pathname with `MAX_PATH_LEN` size, so a single additional byte for flags shouldn't matter much. The rest of the code (`dfs_read`, etc.) seems not to have problems with files larger than 16M.

Alternatively, if this is not wanted, at least `mkdfs` should abort when a file's size is bigger than 16M, rather than silently truncating it.

rasky:
Currently, within DFS, files are laid out as a linked list of sectors.
At runtime, dfs_seek / dfs_read must go through this list in linear
time to access data at a specific offset (see walk_sectors for the
seeking loop). This is very inefficient especially for huge files,
where seeking can take literally seconds.

This PR changes the DFS layout to fix this.

Files are now stored as "raw blobs" within the filesystem. They are
still padded out to be multiple of sectors so that the rest of the code
(directory handling) does not need to change.

Once files are opened, they can be seeked by simply moving the location
pointer, without needing to go through a linked list. This allows for
a true random access, as expected on modern filesystems.

Since I was at it, I also slightly changed the directory layout to
separate the flags and size words, allowing files larger than 2**24.
I also needed this for my experiments with full motion videos.

This is a small benchmark run on a real N64. The benchmark opens
a 16M file, seeks to 6M and read 16K of data:

 * libdragon trunk: 144025 Kcycles
 * this PR, unaligned buffer: 470 Kcycles
 * this PR, aligned buffer: 345 Kcycles

(A "Kcycle" is 1024 cycles as measured by COP0 count register).

Fixes #85

### DIFF
--- a/include/dfsinternal.h
+++ b/include/dfsinternal.h
@@ -12,9 +12,11 @@
  */
 
 /** @brief The special ID value in #directory_entry::flags defining the root sector */
-#define FLAGS_ID        0xFFFFFFFF
+#define ROOT_FLAGS      0xFFFFFFFF
 /** @brief The special ID value in #directory_entry::next_entry defining the root sector */
-#define NEXTENTRY_ID    0xDEADBEEF
+#define ROOT_NEXT_ENTRY 0xDEADBEEF
+/** @brief Special path value in #directory_entry::path defining the root sector */
+#define ROOT_PATH       "DragonFS 2.0"
 
 /** @brief The size of a sector */
 #define SECTOR_SIZE     256
@@ -24,54 +26,41 @@
 /** @brief Representation of a directory entry */
 struct directory_entry
 {
-    /** @brief File size and flags.  See #FLAGS_FILE, #FLAGS_DIR and #FLAGS_EOF */
-    uint32_t flags;
     /** @brief Offset to next directory entry */
     uint32_t next_entry;
+    /** @brief File size and flags.  See #FLAGS_FILE, #FLAGS_DIR and #FLAGS_EOF */
+    uint32_t flags;
     /** @brief The file or directory name */
     char path[MAX_FILENAME_LEN+1];
     /** @brief Offset to start sector of the file */
     uint32_t file_pointer;
 } __attribute__((__packed__));
 
+_Static_assert(sizeof(struct directory_entry) == SECTOR_SIZE, "invalid directory_entry size");
+
 /** @brief Type definition */
 typedef struct directory_entry directory_entry_t;
-
-/** @brief Representation of a file sector */
-struct file_entry
-{
-    /** @brief Offset of next sector of the file */
-    uint32_t next_sector;
-    /** @brief File data */
-    uint8_t data[SECTOR_PAYLOAD];
-} __attribute__((__packed__));
-
-/** @brief Type definition */
-typedef struct file_entry file_entry_t;
 
 /** @brief Open file handle structure */
 typedef struct open_file
 {
-    /** @brief Cached copy of the current sector */
-    file_entry_t cur_sector;
-    /** @brief Pointer to the first sector */
-    file_entry_t *start_sector;
+    /** @brief Cached data to speed up small reads.
+     *
+     *  We want this buffer to be 8-byte aligned for DMA, but also
+     *  16-byte aligned so that it doesn't share cachelines with other
+     *  members of the structure, so it's easier to handle coherency.
+     * */
+    uint8_t cached_data[512] __attribute__((aligned(16)));
+    /** @brief location of the cached data */
+    uint32_t cached_loc;
     /** @brief The unique file handle to refer to this file by */
     uint32_t handle;
     /** @brief The size in bytes of this file */
     uint32_t size;
     /** @brief The offset of the current location in the file */
     uint32_t loc;
-    /** @brief The sector number of the current sector */
-    uint32_t sector_number;
-    /** 
-     * @brief Padding
-     * 
-     * If this isn't here, the second file opened will fail due to cur_sector
-     * not being on a 8 byte aligned boundary, so I just aligned it to 512
-     * bytes. 
-     */
-    uint8_t padding[236];
+    /** @brief The offset within the filesystem where the file is stored */
+    uint32_t cart_start_loc;
 } open_file_t;
 
 /** @} */ /* dfs */
--- a/src/dragonfs.c
+++ b/src/dragonfs.c
@@ -206,7 +206,7 @@ static inline int data_left_in_sector(uint32_t loc)
  */
 static inline uint32_t get_flags(directory_entry_t *dirent)
 {
-    return (dirent->flags >> 24) & 0x000000FF;
+    return (dirent->flags >> 28) & 0x0000000F;
 }
 
 /**
@@ -219,7 +219,7 @@ static inline uint32_t get_flags(directory_entry_t *dirent)
  */
 static inline uint32_t get_size(directory_entry_t *dirent)
 {
-    return dirent->flags & 0x00FFFFFF;
+    return dirent->flags & 0x0FFFFFFF;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


include/dfsinternal.h
```diff
@@ -24,10 +24,12 @@
...
 struct directory_entry
 {
-    /** @brief File size and flags.  See #FLAGS_FILE, #FLAGS_DIR and #FLAGS_EOF */
-    uint32_t flags;
+    /** @brief File size */
+    uint32_t size;
     /** @brief Offset to next directory entry */
     uint32_t next_entry;
+    /** @brief File flags.  See #FLAGS_FILE, #FLAGS_DIR and #FLAGS_EOF */
+    uint8_t flags;
```
DragonMinded: This changes the struct size, and alignment as well, does it not? I'm not sure that path will be the next byte unless we add a packed pragma to this structure as well.

src/dragonfs.c
```diff
@@ -989,50 +952,60 @@ int dfs_read(void * const buf, int size, int count, uint32_t handle)
...
+    if (!to_read)
+        return 0;
+
+    /* If the destination pointer, ROM location and the amount of data to read
+       are all 8-bytes aligned, we can DMA directly into the destination buffer.
+       This is much faster than using cached data, so bypass also our local
+       cache as well. */
+    if ((((uint32_t)buf | (file->cart_start_loc + file->loc) | to_read) & 7) == 0)
+    {
+        /* FIXME: cache ops currently have a bug where data is not properly
```
DragonMinded: This is fixed now, right?

include/dfsinternal.h
```diff
@@ -24,10 +24,12 @@
 /** @brief Representation of a directory entry */
 struct directory_entry
 {
-    /** @brief File size and flags.  See #FLAGS_FILE, #FLAGS_DIR and #FLAGS_EOF */
-    uint32_t flags;
+    /** @brief File size */
+    uint32_t size;
```
DragonMinded: This breaks size out into its own variable to get more than 2^24 bytes in a file, but its still entirely wasteful for directories. Given there's only 3 flag values (IIRC from looking at the code), couldn't we instead get 30 bits for the file size and 2 bits for flags. That's a file size of 1GB instead of 16MB and well beyond anything we could use for files on N64.

DragonMinded (commented):
Couple of comments.

include/dfsinternal.h
rasky: How does it matter that we changed size and / or alignment of fields here? My assumption is that if you're upgrading libdragon you need to regenerate the DFS anyway (files have changed serialization anyhow), so there's no backward compatibility concern.

src/dragonfs.c
rasky: Yes, I'll test it later and submit an update without this FIXME.

include/dfsinternal.h
rasky: (BTW, the structure is indeed already packed, this is also why I decreased MAX_FILENAME_LEN by 1 and moved the flags field just before the path).

include/dfsinternal.h
rasky: That's true, but each directory_entry is save in each own sector, so it's still padded out to 512 bytes no matter what we decide to change here.

But I'm OK with packing flags in upper bits anyhow. If we do, we can restore MAX_FILENAME_LEN to the previous value, and maybe at this point we can also decrease the sector size to 256 bytes rather than 512, so that we actually benefit from the packing we're performing.

Does that sound OK?

rasky: Oh sorry, forget about that, the sector *is* 256 bytes already! I'm not sure why I was thinking 512.

OK I can pack the flags back with the size and maybe add an assert that directory_entry always fits a sector.

jago85:
I also have a suggestion for a possible consistency check. In one of my projects I use the last 8 bytes of the root sector for storing the payload sector count (@ SECTOR_SIZE - 8) and the CRC32 of the payload data (@ SECTOR_SIZE - 4).
Maybe you like to add this or at least don't use the positions for something else.

rasky:
> Could we increase the version number, please?
> The old code cannot read new images and the new code cannot read old images right? So this is worth a new major version number => 2.0, I think.

I can do this, but notice that __dsf_init does not check the version number. The PR still achieves the expected behavior (error message rather than random failures or crashes) because FLAGS_ID (the magic flag value for the root sector) changes from the previous version. If we want to have a real version number handling, we could use NEXTENTRY_ID for that.

> I also have a suggestion for a possible consistency check.

I'll leave this for a followup PR, maybe you could submit what you already have for review?

jago85:
> If we want to have a real version number handling, we could use NEXTENTRY_ID for that.

NEXTENTRY_ID would be possible. But why do you find this better than just use the version string? Is this to not have to compare strings and only use a switch-case around root.next_entry?

> maybe you could submit what you already have for review?

I don't have this on the N64 side yet. I'm using this on a PC software written in C#. But I was thinking about implementing this to check the DFS if this was a useful feature.
Here is a description of what I'm doing just for a better understanding:
I have a flash cart ("Brutzelkarte"). It isn't based on SD Card and SDRAM like ED64 or 64drive, but uses a bootloader which is based on libdragon. The main flash/save ram can be shared between multiple games. A DFS is used to store an INI file that describes the contents of the cart for the bootloader. Maybe I add game covers in the DFS later. The point were I use the sector count and CRC is the tool which creates the DFS. It is written in C# and I have a class that can read/write files from/to that DFS. The connection to the cart is not that fast, so I don't want to read all the cart's memory. So I've added the sector count. Now my tool reads the root sector first and then knows the total amount of memory for the DFS. After reading everything it checks the data using the CRC.

My problem is now, if someone compiles the bootloader with the new libdragon/DFS, the tool needs to be upgraded to. I know this still is the case with only the version number changed. I just wanted to prepare a clean solution for this problem. I'm not 100% sure how I'll manage this. Maybe implementing backwards compatibility. But this may be a lot of work and may not be worth it if no one else needs this. So, I could move forward and upgrade everything as soon as the changes are merged and mark a certain version of libdragon which is to be used.

DragonMinded:
DFS should check the version number, that sounds like a bug in the original implementation.

@jago85 your situation is interesting. DFS was always meant to be read-only since it was a ROM filesystem. Hence why it has no CRC on anything and why it contains no ability to write from within libdragon. Its not even laid out in a way that would allow expansion or updating. The question of graceful upgrades without rebuilding a ROM was never in my mind when I wrote it and I'm not sure it should even handle things like that aside from erroring when you attempt to initialize it with a recognizable and unique error that tells the ROM author "rebuild the FS".

include/dfsinternal.h
DragonMinded: It matters because of wasted space otherwise, as well as possible GCC differences on the host and R4300. AFAIK, struct packing without any directives is undefined in C.

include/dfsinternal.h
DragonMinded: Assert is a good idea! Thanks for the change!

rasky:
OK I've updated the PR:

 * Removed FIXME about cache flush. Also improved this part a bit for speed purposes by selecting the correct cache op depending on the alignment.
 * Restored 32-bit flags and size, using 28 bits for size.
 * Added static assert for directory_entry_t size
 * Bumped version number and added check for it at mount time
 * Fixed dumpdfs to make it work (it looks like it's always been broken on PC, even before my PR?)

jago85:
> DFS was always meant to be read-only

@DragonMinded Yes, I understand this. My tool generates and exchanges the complete DFS everytime something has changed. The readback feature is only used to update the content view when connecting to the cart. My only problem was, that there is no size info for the DFS. If ROMs are added or removed a new DFS is created. The bootloader doesn't use the additional information yet. It uses the normal libdragon routines. The PC tool has access to the flash over USB. No code on the N64 side is running for this.

I find this is a neat feature but not a must-have as the DFS is not particularly dynamic. As long as the last 8 bytes of the root sector are not used for anything else, nothing collides and I can use this as an extension in my project.

tools/dumpdfs/dumpdfs.c
```diff
@@ -5,6 +5,12 @@
 #include "dragonfs.h"
 #include "dfsinternal.h"
 
+#if BYTE_ORDER == BIG_ENDIAN
```
DragonMinded, line 8: I know why this didn't work. I started developing this toolchain on sparc so of course it wouldn't work on intel.

DragonMinded (approved):
Thanks for the updates! There should probably be additional tests around mkdfs and dfsdump, at least to verify that files can be packed and unpacked. No reason to hold up this PR however, and the test suite on-target for DFS will be useful going foward.

PR #117, opened 2021-01-14
Initial libdragon testsuite

rasky:
This commit adds an initial implementation of libdragon testsuite.

The testsuite creates a ROM that runs a series of tests, that verify
libdragon functionality and performance. Hopefully, new tests will be
added as people contribute new functionalities or bugfixes.

The Makefile creates two ROM: one meant for emulators, and another
for a real N64. The latter also benchmarks execution of all tests
and fail when the duration of a test changes unexpectedly. This can
be used to do regression testing on performance improvements contributed
to the codebase.

This is a screenshot of the test ROM:
Schermata 2021-01-14 alle 18 33 05

This is an example of a failure:
Schermata 2021-01-14 alle 18 57 42


tests/test_cache.c
```diff
@@ -0,0 +1,38 @@
...
+			// Write the whole buffer through cache,
+			// so it's all populated in D-Cache.
+			memset(buf, 0xAA, sizeof(buf));
+
+			// Writeback+Invalidate buf[i..i+j]. Now only
+			// those lines should be invalidated.
+			data_cache_hit_writeback_invalidate(buf+i, j);
+			
+			// Read from ROM (header of DFS)
+			dma_read(UncachedAddr(buf), DFS_DEFAULT_LOCATION, 32);
```
anacierdem, line 19: Does using an UncachedAddr make a difference here?

anacierdem:
This looks really good. Just waiting for dragonminded to see it. o/w good to go.

anacierdem:
I think it is not final yet though. Are we going to generate counter.bin or commit it?

tests/test_cache.c
DragonMinded: Does it make sense to ever read from a cached address with DMA? I don't think so, maybe dma_read should auto-promote addresses to uncached for correctness? Not worth fixing in this PR however.

tests/testrom.c
```diff
@@ -0,0 +1,246 @@
...
+// Testsuite definition
+#define TEST_FUNC(fn, dur)   { #fn, fn, dur }
+static const struct Testsuite
+{
+	const char *name;
+	TestFunc fn;
+	uint32_t duration;
+} tests[] = {
+	TEST_FUNC(test_dfs_read, 1104),
+	TEST_FUNC(test_cache_invalidate, 1337),
```
DragonMinded, line 169: I see what you did there.

DragonMinded (commented):
This rules.

Looks like we need the filesystem directory or the DFS tests will fail?

tests/testrom.c
rasky: I won't try to make you believe that it was random... :)

tests/test_cache.c
rasky: yes it might be worth doing it within dma_read. I'll do it at some point I guess

rasky:
I haven't realized that counter.bin matched an exclusion in gitignore, which is why it was left out from the commit. I've renamed it to counter.dat and pushed it.

PR #118, opened 2021-01-24
Fix cpptest example

### LINKED ISSUE #92: cpptest does not work
It compiles OK but linking with `g++` does not work.
See the Makefile -> `LD = $(GCCN64PREFIX)g++`

meeq:
I have verified that the cpptest example now works correctly in MAME.

Resolves https://github.com/DragonMinded/libdragon/issues/92

* Fix global constructors for C++
* Remove unused global destructors from `ld` script
* Minor improvements to cpptest example

According to https://gcc.gnu.org/onlinedocs/gccint/Initialization.html
>Each list always begins with an ignored function pointer (which may hold 0, -1, or a count of the function pointers after it, depending on the environment).

>This is followed by a series of zero or more function pointers to constructors (or destructors), followed by a function pointer containing zero. 

>Constructors are called in reverse order of the list

Global destructors weren't exactly hurting anything by being included, but they were never being called and there is no logical reason _to_ call them since the system effectively halts after `main` returns (which it should pretty much never do).

### DIFF
--- a/examples/cpptest/cpptest.cpp
+++ b/examples/cpptest/cpptest.cpp
@@ -14,7 +14,7 @@ class TestClass
 
     public:
         TestClass() {
-            d = 2;
+            d = 100;
         }
         int f1()
         {
--- a/examples/cpptest/cpptest.cpp
+++ b/examples/cpptest/cpptest.cpp
@@ -10,7 +10,7 @@ static bitdepth_t bit = DEPTH_32_BPP;
 class TestClass
 {
     private:
-        int d;
+        uint32_t d;
 
     public:
         TestClass() {
@@ -18,16 +18,17 @@ class TestClass
         }
         int f1()
         {
+            d = d + 1;
             return d;
         }
 };
 
 // Test global constructor
-TestClass o1;
+TestClass globalClass;
 
 int main(void)
 {
-    int* a = new int(2);
+    TestClass* localClass = new TestClass();
 
     init_interrupts();
 
@@ -41,9 +42,10 @@ int main(void)
     {
         console_clear();
         printf("Libdragon v%d.%d.%d \n", libdragon_version.major, libdragon_version.minor, libdragon_version.revision);
-        printf("Test: %d\n", o1.f1());
+        printf("Global class method: %d\n", globalClass.f1());
+        printf("Local class method: %d\n", localClass->f1());
         console_render();
     }
 
-    delete a;
+    delete localClass;
 }
--- a/n64.ld
+++ b/n64.ld
@@ -66,8 +66,6 @@ SECTIONS {
     . = ALIGN(8);
 
     .ctors : {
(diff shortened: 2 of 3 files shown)


bryanperris:
I wondered if 0xFFFFFFFF was the result of a bug subtracting 1 from a 32-bit unsigned 0.

meeq:
[According to my research](https://reverseengineering.stackexchange.com/a/6286), `0xFFFFFFFF` is supposed to be the sentinel value at the _end_ of the `.ctors` list, but for whatever reason it is always appearing at the _beginning_ of the list.

```sh
$N64_INST/mips64-elf/bin/objdump -Dr -j .ctors cpptest.elf
```
```
cpptest.elf:     file format elf32-bigmips


Disassembly of section .ctors:

80028930 <__CTOR_LIST_SIZE__>:
80028930:	00000003

80028934 <__CTOR_LIST__>:
80028934:	ffffffff
80028938:	800215c0
8002893c:	80022958

80028940 <__CTOR_END__>:
	...
```

meeq:
[According to some other research](https://www.exploit-db.com/papers/13234), the expected structure of `.ctors` is:
```
ffffffff <ANOTHER ADDRESS OF FUNCTION> <ADDRESS OF FUNCTION> 00000000
```

...which is consistent with what we're seeing in the linked ELF.

rasky:
Patch looks great to me, thanks!

anacierdem:
Let me try it on hardware and merge. Thank you! This is a really important step.

meeq:
I'm actually noticing that the global constructor for `TestClass globalClass;` in the cpptest example isn't actually being called... 😞

At least it's not completely failing anymore. 😅 

I'll see if I can figure out what's going on here.

meeq:
@anacierdem I have fixed the issue with the global constructors not being called: pointers to pointers are confusing. 😅 

Luckily, MAME's debugger is super awesome and let me step through to find the problem. 🎉

anacierdem:
Those external labels make things even more complicated. I did the same many times myself too 😁 I have a minor comment. I will try to test and merge this when I have the time

anacierdem:
Unfortunately this does not work on the real console. There is something wrong with how we arelinking it IMO b/c removing everything and only leaving a `printf` does not work either with that  makefile. This is the barebones code I'm testing;
    
    #include <stdio.h>
    #include <libdragon.h>
    
    static resolution_t res = RESOLUTION_320x240;
    static bitdepth_t bit = DEPTH_32_BPP;
    
    int main(void)
    {
        /* enable interrupts (on the CPU) */
        init_interrupts();
    
        /* Initialize peripherals */
        console_init();
        controller_init();
    
        console_set_render_mode(RENDER_MANUAL);
        while(1) 
        {
            console_clear();
            printf( "TEST" );
            console_render();
        }
    }
Still we can merge it though as it seem to fix the ctors according to the spec.

PR #120, opened 2021-01-30
Fix delay functions

anacierdem:
`wait_ticks` and `wait_ms` functions were not working correctly because of the limited range of the underlying hardware register size (32bits). This gets consumed in about 91 seconds and it is not possible to do a simple comparison to wait for some time.

It is fixed by limiting the wait range to `0x7FFFFFFF` ticks which allows detecting the direction of the current count register with respect to the reference time.

### DIFF
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -183,8 +183,7 @@ static inline void wait_ticks( unsigned long wait )
  */
 static inline void wait_ms( unsigned long wait_ms )
 {
-    unsigned int wait = wait_ms * (TICKS_PER_SECOND / 1000);
-    wait_ticks(wait);
+    wait_ticks(TICKS_FROM_MS(wait_ms));
 }
 
 void data_cache_hit_invalidate(volatile void *, unsigned long);
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -127,7 +127,7 @@
 /**
  * @brief Returns equivalent count ticks for the given millis.
  */
-#define TICKS_FROM_MS(val) ((uint32_t)(val * (TICKS_PER_SECOND / 1000)))
+#define TICKS_FROM_MS(val) ((uint32_t)((val) * (TICKS_PER_SECOND / 1000)))
 
 #ifdef __cplusplus
 extern "C" {
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -83,11 +83,26 @@
 #define MEMORY_BARRIER() asm volatile ("" : : : "memory")
 
 /**
+ * @brief Returns the COP0 register $9 (count).
+ *
+ * The coprocessor 0 (system control coprocessor - COP0) register $9 is
+ * incremented at "half maximum instruction issue rate" which is the processor
+ * clock speed (93.75MHz) divided by two. (also see TICKS_PER_SECOND) It will
+ * always produce a 32-bit unsigned value which overflows back to zero in
+ * 91.625 seconds. The counter will increment irrespective of instructions
+ * actually being executed. This macro is for reading that value.
+ * Do not use for comparison without special handling.
+ */
+#define TICKS_READ() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$9":"=r"(x)); \
+    x; \
+})
+
+/**
  * @brief Number of updates to the count register per second
  *
- * The count register updates at "half maximum instruction issue rate".
- * This appears to be half the CPU clock.  Every second, this many counts
- * will have passed in the count register
+ * Every second, this many counts will have passed in the count register
  */
 #define TICKS_PER_SECOND (93750000/2)
 
@@ -104,22 +119,15 @@
  *
  * This is similar to t1 < t2, but it correctly handles timer overflows
  * which are very frequent. Notice that the N64 counter overflows every
- * 90 seconds, so it's not possible to compare times that are more than
- * 45 seconds apart.
+ * ~91 seconds, so it's not possible to compare times that are more than
+ * ~45 seconds apart.
  */
 #define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
 
 /**
- * @brief Returns the COP0 register $9 (count).
- *
- * Returns the COP0 register $9 (count).
- * It will always produce a 32 bit unsigned value.
+ * @brief Returns equivalent count ticks for the given millis.
  */
-#define TICKS_READ() ({ \
-    uint32_t x; \
-    asm volatile("mfc0 %0,$9":"=r"(x)); \
-    x; \
-})
+#define TICKS_FROM_MS(val) ((uint32_t)(val * (TICKS_PER_SECOND / 1000)))
 
 #ifdef __cplusplus
 extern "C" {
@@ -130,9 +138,7 @@ void sys_set_boot_cic(int bc);
 /**
  * @brief Read the number of ticks since system startup
  *
- * @note This is the clock rate divided by two.
- * It will wrap back to 0 after about 91.6 seconds
- * Do not use for comparison without special handling
+ * @note Also see the TICKS_READ macro
  *
  * @return The number of ticks since system startup
  */
@@ -144,8 +150,9 @@ static inline volatile unsigned long get_ticks(void)
 /**
  * @brief Read the number of millisecounds since system startup
  *
- * It will wrap back to 0 after about 91.6 seconds
- * Do not use for comparison without special handling
+ * @note It will wrap back to 0 after about 91.6 seconds.
+ * Also see the TICKS_READ macro. Do not use for comparison
+ * without special handling.
  *
  * @return The number of millisecounds since system startup
  */
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -89,36 +89,37 @@
  * This appears to be half the CPU clock.  Every second, this many counts
  * will have passed in the count register
  */
-#define COUNTS_PER_SECOND (93750000/2)
+#define TICKS_PER_SECOND (93750000/2)
 
-// Returns the signed difference of time between "from" and "to". If "from"
-// is before "to", the distance in time is positive, otherwise it is
-// negative .
+/**
+ * @brief The signed difference of time between "from" and "to".
+ *
+ * If "from" is before "to", the distance in time is positive,
+ * otherwise it is negative.
+ */
 #define TICKS_DISTANCE(from, to) ((int32_t)((uint32_t)(to) - (uint32_t)(from)))
 
-// Returns true if "t1" is before "t2". This is similar to t1 < t2,
-// but it correctly handles timer overflows which are very frequent.
-//
-// Notice that the N64 counter overflows every 90 seconds, so it's
-// not possible to compare times that are more than 45 seconds apart.
+/**
+ * @brief Returns true if "t1" is before "t2".
+ *
+ * This is similar to t1 < t2, but it correctly handles timer overflows
+ * which are very frequent. Notice that the N64 counter overflows every
+ * 90 seconds, so it's not possible to compare times that are more than
+ * 45 seconds apart.
+ */
 #define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
 
-// Reads the COP0 register $9 (count) into the provided variable.
-// It will always produce a 32 bit unsigned value.
-#define READ_TICKS(variable) do { \
-    _Static_assert(sizeof(variable) >= 4, "target variable is narrower than 32 bits"); \
-    asm volatile("mfc0 %0,$9":"=r"(variable)); \
-} while(0)
-
-#define WAIT_TICKS(wait) do { \
-    unsigned int initial_tick; \
-    unsigned int current_tick; \
-    READ_TICKS(initial_tick); \
-    READ_TICKS(current_tick); \
-    while( current_tick - initial_tick < wait ) { \
-        READ_TICKS(current_tick); \
-    }; \
-} while(0)
+/**
+ * @brief Returns the COP0 register $9 (count).
+ *
+ * Returns the COP0 register $9 (count).
+ * It will always produce a 32 bit unsigned value.
+ */
+#define TICKS_READ() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$9":"=r"(x)); \
+    x; \
+})
 
 #ifdef __cplusplus
 extern "C" {
@@ -137,9 +138,7 @@ void sys_set_boot_cic(int bc);
  */
 static inline volatile unsigned long get_ticks(void)
 {
-    unsigned long count;
-    READ_TICKS(count);
-    return count;
+    return TICKS_READ();
 }
 
 /**
@@ -152,9 +151,7 @@ static inline volatile unsigned long get_ticks(void)
  */
 static inline volatile unsigned long get_ticks_ms(void)
 {
-    unsigned long count;
-    READ_TICKS(count);
-    return count / (COUNTS_PER_SECOND / 1000);
+    return TICKS_READ() / (TICKS_PER_SECOND / 1000);
 }
 
 /**
@@ -166,7 +163,8 @@ static inline volatile unsigned long get_ticks_ms(void)
  */
 static inline void wait_ticks( unsigned long wait )
 {
-    WAIT_TICKS(wait);
+    unsigned int initial_tick = TICKS_READ();
+    while( TICKS_READ() - initial_tick < wait );
 }
 
 /**
@@ -178,8 +176,8 @@ static inline void wait_ticks( unsigned long wait )
  */
 static inline void wait_ms( unsigned long wait_ms )
 {
-    unsigned int wait = wait_ms * (COUNTS_PER_SECOND / 1000);
-    WAIT_TICKS(wait);
+    unsigned int wait = wait_ms * (TICKS_PER_SECOND / 1000);
+    wait_ticks(wait);
 }
 
 void data_cache_hit_invalidate(volatile void *, unsigned long);
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -82,16 +82,105 @@
  */
 #define MEMORY_BARRIER() asm volatile ("" : : : "memory")
 
+/**
+ * @brief Number of updates to the count register per second
+ *
+ * The count register updates at "half maximum instruction issue rate".
+ * This appears to be half the CPU clock.  Every second, this many counts
+ * will have passed in the count register
+ */
+#define COUNTS_PER_SECOND (93750000/2)
+
+// Returns the signed difference of time between "from" and "to". If "from"
+// is before "to", the distance in time is positive, otherwise it is
+// negative .
+#define TICKS_DISTANCE(from, to) ((int32_t)((uint32_t)(to) - (uint32_t)(from)))
+
+// Returns true if "t1" is before "t2". This is similar to t1 < t2,
+// but it correctly handles timer overflows which are very frequent.
+//
+// Notice that the N64 counter overflows every 90 seconds, so it's
+// not possible to compare times that are more than 45 seconds apart.
+#define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
+
+// Reads the COP0 register $9 (count) into the provided variable.
+// It will always produce a 32 bit unsigned value.
+#define READ_TICKS(variable) do { \
+    _Static_assert(sizeof(variable) >= 4, "target variable is narrower than 32 bits"); \
+    asm volatile("mfc0 %0,$9":"=r"(variable)); \
+} while(0)
+
+#define WAIT_TICKS(wait) do { \
+    unsigned int initial_tick; \
+    unsigned int current_tick; \
+    READ_TICKS(initial_tick); \
+    READ_TICKS(current_tick); \
+    while( current_tick - initial_tick < wait ) { \
+        READ_TICKS(current_tick); \
+    }; \
+} while(0)
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 int sys_get_boot_cic();
 void sys_set_boot_cic(int bc);
-volatile unsigned long get_ticks( void );
-volatile unsigned long get_ticks_ms( void );
-void wait_ticks( unsigned long wait );
-void wait_ms( unsigned long wait );
+/**
+ * @brief Read the number of ticks since system startup
+ *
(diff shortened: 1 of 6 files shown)


jago85:
I'd recommend to solve it like that:

- save current tick value at entry of wait_ticks
- than in the while loop calculate the difference and wait until it reaches the desired time
- this will also work if there is an overflow because of the unsigned logic
- I'd call wait_ticks in wait_ms
 
something like that:
```
void wait_ticks( unsigned int wait )
{
    unsigned int tick_value = get_ticks();

    while( get_ticks() - tick_value < wait );
}
```

anacierdem:
Changed my perspective there. You are right, this is far better.

rasky:
Unfortunately this won't also work because get_ticks returns a 31-bit counter. This causes endless troubles.

I've got pending patches to fix this, where the basic trick is that get_ticks() should shift left its counter by 1. Then you need to adjust conversions to ms of course, but at least you get a full 32-bit counter on which you can perform differences with safe overflow semantics.

anacierdem:
Hmm, I didn't get why it is 31 bits? I assume N64 is not based on R4300 variant then?

rasky:
OK scratch that, I just retested it and it's false. The counter is 32-bit.

```C
    uint32_t wcount = 0xFFFFA000;
    asm volatile("mtc0 %0,$9\n\t nop \n\t" : : "r" (wcount)); \
    for (int i=0;i<10;i++)
        printf("%08lx\n", get_ticks());
```
```
ffffa01b
ffffc764
ffffd65e
ffffe547
fffff455
00000d32
00001c91
00002b97
00003aa6
000049c3
```

and there's no overflow at 31 bit:


```C
    uint32_t wcount = 0x7FFFA000;
    asm volatile("mtc0 %0,$9\n\t nop \n\t" : : "r" (wcount)); \
    for (int i=0;i<10;i++)
        printf("%08lx\n", get_ticks());
```
```
7fffa01c
7fffc698
7fffd561
7fffe3fb
7ffff295
8000014c
80000fef
80001e89
80002d23
80003be2
```

I wonder why I got this idea, probably a different unrelated bug.

bryanperris:
Instead of polling the count, why not poll for the MIPS timer interrupt pending bit?  It should still get set even if interrupts are disabled. You will have to set the compare register to count + x.

anacierdem:
There is already a timer subsystem that is based on the compare register, which @rasky will probably re-implement at some point. So we will probably spare the compare register for that. Just reading the register will not interfere with that mechanism and if we fix the timers too (it currently writes to the count - which it shouldn't) this can work together with it.
Maybe we can incorporate the delay mechanism into the timer system but I don't think it is necessary. This is only there for a simple delay mechanism. With the IRQ and stuff it will probably also get less accurate.

bryanperris:
For a simple delay loop, what about incrementing a GPR till you hit the target value?  

I don't think the delay function is safe when count can wrap over especially if the routine gets interrupted.  At least with a GPR, the exception makes sure it gets restored after returning. Also it seems the count increment rate is different between NTSC and PAL based on official docs.

anacierdem:
Doing with GPRs is also an option but much more tedious to make accurate. Even if it gets interrupted, it is safe as the calculation itself is wrap-safe (unless the interrupt handlers take longer than the time required to wrap to the same `tick_value`, which is highly unlikely).
Also the CPU clock rate seem to be the same for different regions. Which doc are you basing that on?

rasky:
@anacierdem can you add tests for this? I'd test at least the overflow condition.

bryanperris:
@anacierdem The official N64 doc where it explains about the osGetTime function.

rasky:
The patch looks good to me. There are the following closely related modifications that I think should be added, but can be done in a separate PR if you want:

 * Adding ticks manipulation functions. I'm using these, but they should be probably renamed for libdragon (change `HWTIMER` into `TICKS`). If you add these to n64sys.h, you can use them within your PR:

```c
// Returns the signed difference of time between "from" and "to". If "from"
// is before "to", the distance in time is positive, otherwise it is
// negative .
#define HWTIMER_DISTANCE(from, to) ((int32_t)((uint32_t)(to) - (uint32_t)(from)))

// Returns true if "t1" is before "t2". This is similar to t1 < t2,
// but it correctly handles timer overflows which are very frequent.
//
// Notice that the N64 counter overflows every 90 seconds, so it's
// not possible to compare times that are more than 45 seconds apart.
#define HWTIMER_BEFORE(t1, t2) ({ HWTIMER_DISTANCE(t1, t2) > 0; })
```

* Move `get_ticks` to the header file as an inline function. I need the lowest-level primitive to read the tick counter to be as fast as possible (especially in the context of a profiling library I will contribute at some point), so the overhead of a call is not acceptable.
* Add tests at least for the overflow condition.

anacierdem:
- Added them as macros and header inline functions.
- Added tests.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
 #define MEMORY_BARRIER() asm volatile ("" : : : "memory")
 
+/**
+ * @brief Number of updates to the count register per second
+ *
+ * The count register updates at "half maximum instruction issue rate".
+ * This appears to be half the CPU clock.  Every second, this many counts
+ * will have passed in the count register
+ */
+#define COUNTS_PER_SECOND (93750000/2)
```
rasky: I think we should convert this to `TICKS_PER_SECOND`, given that we're using the term "ticks" in other macros.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+
+// Returns the signed difference of time between "from" and "to". If "from"
```
rasky: Maybe you should use the Doxygen style comments here as well, and for other macros.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+// Notice that the N64 counter overflows every 90 seconds, so it's
+// not possible to compare times that are more than 45 seconds apart.
+#define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
+
+// Reads the COP0 register $9 (count) into the provided variable.
+// It will always produce a 32 bit unsigned value.
+#define READ_TICKS(variable) do { \
+    _Static_assert(sizeof(variable) >= 4, "target variable is narrower than 32 bits"); \
+    asm volatile("mfc0 %0,$9":"=r"(variable)); \
+} while(0)
```
rasky: Can we use a statement expression here, so that the macro returns the value, rather than filling a variable passed in? It's far easier to use `var = READ_TICKS()`, especially because you can do stuff `if (TICKS_BEFORE(READ_TICKS(), deadline))`

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+// but it correctly handles timer overflows which are very frequent.
+//
+// Notice that the N64 counter overflows every 90 seconds, so it's
+// not possible to compare times that are more than 45 seconds apart.
+#define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
+
+// Reads the COP0 register $9 (count) into the provided variable.
+// It will always produce a 32 bit unsigned value.
+#define READ_TICKS(variable) do { \
+    _Static_assert(sizeof(variable) >= 4, "target variable is narrower than 32 bits"); \
```
rasky: Maybe you needed this because of the way the macro works now, but in general I wouldn't do this. Anywhere there is a function returning a variable, it gets silently truncated if it's assigned to another one of smaller type. This is just the way C works, I don't think there's a good reason for this specific macro to forbid this behavior.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+// Returns true if "t1" is before "t2". This is similar to t1 < t2,
+// but it correctly handles timer overflows which are very frequent.
+//
+// Notice that the N64 counter overflows every 90 seconds, so it's
+// not possible to compare times that are more than 45 seconds apart.
+#define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
+
+// Reads the COP0 register $9 (count) into the provided variable.
+// It will always produce a 32 bit unsigned value.
+#define READ_TICKS(variable) do { \
```
rasky: Maybe `TICKS_READ()` so that all these macros have the same prefix `TICKS_`? In general, in C I'm a fan of using prefixes in place of namespaces that don't exist.

anacierdem:
Hmm ARM_INTERRUPT is not used anymore, I'll remove it

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+
+#define WAIT_TICKS(wait) do { \
+    unsigned int initial_tick; \
+    unsigned int current_tick; \
+    READ_TICKS(initial_tick); \
+    READ_TICKS(current_tick); \
+    while( current_tick - initial_tick < wait ) { \
+        READ_TICKS(current_tick); \
+    }; \
+} while(0)
```
rasky: Personally, I don't see a value in making this a macro. This function is slow by definition -- it's waiting. Inlining too much code just increase the code size which is particularly bad on N64 because of the slow RDRAM. If somebody needs to wait, they can't be concerned for the small overhead of a function call.

On the contrary, a function call to read the timer can kill profiling on high-speed code, which is why I was favoring a macro to read the register.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+    _Static_assert(sizeof(variable) >= 4, "target variable is narrower than 32 bits"); \
+    asm volatile("mfc0 %0,$9":"=r"(variable)); \
+} while(0)
+
+#define WAIT_TICKS(wait) do { \
+    unsigned int initial_tick; \
+    unsigned int current_tick; \
+    READ_TICKS(initial_tick); \
+    READ_TICKS(current_tick); \
+    while( current_tick - initial_tick < wait ) { \
```
rasky: you should probably use `TICKS_DISTANCE` here.

include/n64sys.h
```diff
@@ -82,16 +82,105 @@
...
+ *
+ * @param[in] wait
+ *            Number of millisecounds to wait
+ *            Maximum accepted value is 91625 ms
+ */
+static inline void wait_ms( unsigned long wait_ms )
+{
+    unsigned int wait = wait_ms * (COUNTS_PER_SECOND / 1000);
+    WAIT_TICKS(wait);
+}
```
rasky, line 187: The same comment as above applies to these wait functions.

tests/test_ticks.c
```diff
@@ -0,0 +1,149 @@
...
+	uint32_t ticks_0;
+	uint32_t ticks_1;
+
+	uint32_t continue_ticks;
+	READ_TICKS(continue_ticks);
+
+	disable_interrupts();
+
+	// Put the instructions on the same cacheline
+	asm volatile(".align 4");
```
rasky: cachelines are 64 bytes, so `.align 4` sounds too little. Are you trying to avoid a RDRAM stall while executing these instructions? If you do, I would keep it simpler: just run it in a loop and discard the first occurrence. The second time you loop, the loop body will be in the cache.

tests/test_ticks.c
```diff
@@ -0,0 +1,149 @@
...
+	WRITE_TICKS(0x0);
+	READ_TICKS(ticks_0);
+	WRITE_TICKS(0xFFFFFFFF);
+	READ_TICKS(ticks_1);
+
+	enable_interrupts();
+
+	ASSERT(ticks_0 == 0x0 && ticks_1 == 0xFFFFFFFF, "not reading correct register. Received %#010lX and %#010lX", ticks_0, ticks_1);
```
rasky: What is the purpose of this test? Assuming that the hardware works, which functionality is this test asserting which is not covered by the other tests already? This test is really strict on timing, so hopefully it's not hard to maintain.

tests/testrom.c
```diff
@@ -167,80 +167,85 @@ static const struct Testsuite
...
+
+			if (ctx.log != logbuf) {			
+				printf("%s\n\n", logbuf);
+			}
+		}
+	#if BENCHMARK_TESTS
+		// If there's more than a 10% drift on the running time (/1024) compared to
+		// the expected one, make the test fail. Something happened and we
+		// need to double check this.
+		else if (tests[i].duration > 0 && ((float)test_diff / (float)test_duration > 0.1))
```
rasky: 10% sounds like a lot to keep as a performance test. Can you elaborate on why this is needed?

tests/test_ticks.c
```diff
@@ -0,0 +1,149 @@
...
+// Timeout: If the wait loop fails to exit upon setting the mock value, the VI
+
+// interrupt will move the state into this phase, which is asserted in the test.
+// So we can test them each of them in three frames, without waiting the full
+// delay. Tested wait should be an order of magnitude longer than the VI interval
+// to prevent the loop from exiting before we are able to update the count reg.
+// If wait calculation is unable to wrap properly, the loop will run for a maximum
+// of 90 secs and endup in Timeout state, failing the tes.
+// Execution times are negligible with this setup.
+// 0x2CB4178 is 1 second
```
rasky, line 36: Can you elaborate why this complex system with VI is required for this test? What I would do is simply to test the functions with small intervals, and then verify after the wait that the current ticks fall within a range of the expected value.

Maybe there's something I'm missing.

rasky (changes requested):
Thanks for the updates! I've left a few comments.

include/n64sys.h
anacierdem: Hmm, I explicitly want this to be unsigned o/w the comparison will be limited to 45 secs

tests/test_ticks.c
anacierdem: That was how I implemented it first. Maybe I can go back to that

tests/testrom.c
anacierdem: Because on my tests on an everdrive they tend to be off by 7-9%. Couldnt find a reason for that. I think having more slack for IO benchmarks is the only way here.

tests/test_ticks.c
anacierdem: I wanted to make sure to test it with intervals much larger than the execution time of any code to make sure we are actually crossing the wrapping boundary i.e it should not cross it without us knowing. Might be a little overengineered you are probably right. Should be fine with 5ms around the boundary. Let me try to do it that way and see

tests/test_ticks.c
anacierdem: It also makes sure they are properly inlined without any overhead, might be unnecessary

include/n64sys.h
anacierdem: How can we do it as a statement? Inline assembly is expecting an lvalue

anacierdem: Fpund [this](https://stackoverflow.com/questions/59754748/inline-assembly-in-c-function-with-return-statement) but that would remove the inlining guarantee. We can use get_ticks for that purpose then

tests/test_ticks.c
anacierdem: Also this enables us to test a very long wait, which would be impractical o/w. Lets keep it for now if thats ok?

include/n64sys.h
rasky: If you use GCC [statement expressions](https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html), you can do that like this:

```C
#define TICKS_READ() ({ \
	uint32_t x; \
	asm volatile("mfc0 %0,$9\n\t nop \n\t" : "=r" (x) : ); \
	x; \
})
```

tests/test_ticks.c
rasky: I've never seen dropping asm align statements in the middle of a function to align specific opcodes. I wouldn't swear it always works and has no side effect.

tests/testrom.c
rasky: I see, it is probably the case that I/O speed with different SDKs is really unpredictable (eg: the SD card might be slower or faster). I would suggest to add a flag to mark those tests as "I/O TESTS" in the tests structure (this flag set can also be useful in the future), and use that flag to raise the allowed drift to 10% for now.

I now we have a too little testsuite for now, but my hope is that for many tests we can keep some kind of semi-perfect timing. Maybe I'm optimistic and we will have to change all of them to 10%, but I'd try to keep this assumption for now.

tests/test_ticks.c
rasky: OK. Out of curiosity, have you verified the test actually fails without the rest of this PR?

tests/test_ticks.c
anacierdem: Ah this is the macro version and should be inlined already, I think that's what you mean. In case someone changes that macro to a fnction call maybe?

tests/test_ticks.c
anacierdem: Yes it does fail with the old implementation. Also, inspired by your proposal, I have changed it a little bit to actually wait for 5ms, to verify it is doing some waiting indeed :)

include/n64sys.h
anacierdem: I does not change the behaviour (does unsigned compare) unless we change the r.h.s of the comparison to a signed integer, so it is also viable. On the other hand, I think this is the intended calculation and we should keep it this way, what do you think?

anacierdem:
Thank you for your time reviewing this!
I tried to address as many of them as possible. I have also improved the tests to force the wait functions to do some waiting (i.e make sure they are not exiting as soon as we set the mock count). They fail with the old implementation as expected. I agree this might be a little overengineered but it allows testing very long waits in a short time and it also constitutes a nice benchmark for emulator timer/VI implementations.

One more note: I haven't tried this on an emulator and I don't think it will pass. Should we also wrap those with a `#if BENCHMARK_TESTS`?

include/n64sys.h
```diff
@@ -89,36 +89,37 @@
...
-        READ_TICKS(current_tick); \
-    }; \
-} while(0)
+/**
+ * @brief Returns the COP0 register $9 (count).
+ *
+ * Returns the COP0 register $9 (count).
+ * It will always produce a 32 bit unsigned value.
+ */
+#define TICKS_READ() ({ \
```
rasky: Re-reading the whole file like I was new to it, I found out confusing this specific documentation as it's too technical. I think this function should maybe be moved as first macro (even before TICKS_PER_SECOND) and it should explain what it is this counter, the expected frequency (also referring to TICKS_PER_SECOND, etc). Then TICKS_PER_SECOND could have a simpler documentation. 

See also the documentation of get_ticks(). I think that, in general, they should be very similar since they're the same function.

tests/test_ticks.c
```diff
@@ -4,23 +4,18 @@
...
-	WRITE_TICKS(0x0); \
-	WRITE_COMPARE(compare); \
-} while(0)
-
 #define TEST(name, tests) do { \
 	const int NUM_CASES = sizeof(tests) / sizeof(tests[0]); \
 	for (int i=0; i < NUM_CASES; i++) { \
-		mock_tick = tests[i][1]; \
+		/* move ticks just before 5ms of the target to make sure wait is actually waiting */ \
+		mock_tick = tests[i][1] - 0x39387; \
```
rasky: I wonder if we should have a `TICKS_FROM_MS()` macro as well. I know we have `wait_ms()` but this code proves us that we're now giving out all the helpers that we require to write the code without magic numbers.

This line would become `mock_tick = tests[i][1] - TICKS_FROM_MS(5)`

tests/test_ticks.c
```diff
@@ -64,53 +58,52 @@ static uint32_t test_cases[][3] = {
...
 
 	disable_interrupts();
 
 	// Put the instructions on the same cacheline
-	asm volatile(".align 4");
-	WRITE_TICKS(0x0);
-	READ_TICKS(ticks_0);
-	WRITE_TICKS(0xFFFFFFFF);
-	READ_TICKS(ticks_1);
+	for (int i = 0; i < 2; i++) {
```
rasky, line 90: The comment needs updating here. You may want to explain that you're looping twice to make sure everything is in icache.

tests/test_ticks.c
rasky: yeah, anyway that's fine, it's not a big deal.

rasky (approved):
Thanks, I've left a few other comments but most of it looks wonderful to me!
I haven't gotten around testing it yet, I'll test it later today on real hardware.

> One more note: I haven't tried this on an emulator and I don't think it will pass.

The testsuite contains an "emulator mode". I'll check that as well.

rasky:
OK tested it on real hardware and it's fine.

The whole test won't pass under an emulator so I'd just comment it away altogether (remove the entry from the testsuite with #ifdef).

anacierdem:
I think we are almost there.

include/n64sys.h
```diff
@@ -82,16 +82,110 @@
...
+ * which are very frequent. Notice that the N64 counter overflows every
+ * ~91 seconds, so it's not possible to compare times that are more than
+ * ~45 seconds apart.
+ */
+#define TICKS_BEFORE(t1, t2) ({ TICKS_DISTANCE(t1, t2) > 0; })
+
+/**
+ * @brief Returns equivalent count ticks for the given millis.
+ */
+#define TICKS_FROM_MS(val) ((uint32_t)(val * (TICKS_PER_SECOND / 1000)))
```
rasky: This required parentheses around `val`

rasky (commented):
The new documentation is better in my opinion, thanks!

include/n64sys.h
anacierdem: Im not sure I get this one?

anacierdem:
Ill be merging this soon. Thanks!

tests/test_ticks.c
```diff
@@ -0,0 +1,139 @@
...
+		ticks_0 = TICKS_READ(); \
+		ASSERT(state != Timeout, "Case: %u Test timed out. " #name " didn't finish on time (Ticks: %#10lX)", i, ticks_0); \
+		ASSERT(state == Test && ticks_0 >= tests[i][1], "Case: %u " #name " finished too early (Ticks: %#10lX)", i, ticks_0); \
+		ASSERT(state <= Timeout, "Case: %u Unexpected state for " #name " (Ticks: %#10lX)", i, ticks_0); \
+		/* re-sync to frame */ \
+		while (state < Timeout); \
+		/* reset state */ \
+		state = Start; \
+	} \
+} while(0)
```
rasky: As a final read, why this must be a macro and can't simply be a function?

anacierdem: 😀 I have no idea. Let me change it

anacierdem: Ah I remember now, I used to test the macro versions as well and probably wanted to be able to test them both without much change. It is a function now.

rasky: Great! For me this PR is done, love it!

PR #121, opened 2021-02-06
Makefile: add -ffunction-sections and -fdata-sections to libdragon

rasky:
I found beneficial to always compile and link with these options that
allow to reduce code/data footprint of libdragon because the linker
is able to remove single functions (and relative global variables) that
are unused by the game. Given the small amount of RAM available, every
bit counts.


anacierdem:
Do we also need the `--gc-sections` for the linker?

rasky:
Yes but that's a linker flags that games must specify while linking to libdragon. Anyway, I've updated all examples to use it.

anacierdem:
We may also want it for the test rom, what do you think?

rasky:
Yes, done.

PR #122, opened 2021-02-07
Fix icache ops and support GCC nested functions

rasky:
GCC has an extension that allows to create nested functions that can
refer to variable in the outer scope (aka lexical closures). This is
more useful than it sounds even in the context of N64 programming,
to define stuff like interrupt callbacks.

Since this feature requires generating a trampoline at runtime, GCC
generates a call to a function named "_flush_cache" that must be
implemented by the operating system runtime, to flush the instruction
cache. This PR adds such a function, whose implementation is trivial.

While implementing this, I noticed that cache ops are wrong for
instruction cache because the cachelines are 32 bytes (instead of
16 bytes like the data cache).

Also, while writing the code, an existing test started to fail and I realized
it was not correct (a buffer that was supposedly in cache was only
written and never read). I fixed this as well.

### DIFF
--- a/tests/test_cache.c
+++ b/tests/test_cache.c
@@ -7,9 +7,10 @@ void test_cache_invalidate(TestContext *ctx) {
 	for (int i=0;i<32;i++) {
 		for (int j=0;j<32;j++) {
 
-			// Write the whole buffer through cache,
+			// Read/write the whole buffer through cache,
 			// so it's all populated in D-Cache.
-			memset(buf, 0xAA, sizeof(buf));
+			memset(buf, 0xA0, sizeof(buf));
+			for (int i=0;i<sizeof(buf);i++) buf[i] += 0xA;
 
 			// Writeback+Invalidate buf[i..i+j]. Now only
 			// those lines should be invalidated.
--- a/tests/test_irq.c
+++ b/tests/test_irq.c
@@ -1,0 +1,45 @@
+
+void test_irq_reentrancy(TestContext *ctx) {
+	// Make sure that enable_interrupts() does not cause an interrupt
+	// to trigger if called under interrupt.
+	volatile bool cb1_called = false;
+	volatile bool cb2_called = false;
+	bool cb1_running = false;
+	bool fail_reentrant = false;
+	bool fail_order = false;
+
+	void cb1(int ovlf) {
+		cb1_called = true;
+		cb1_running = true;
+		disable_interrupts();
+
+		// Wait for the other timer interrupt to trigger
+		wait_ms(30);
+
+		// This enable_interrupts should not trigger the other
+		// interrupt, otherwise the test will fail.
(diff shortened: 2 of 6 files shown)


tests/test_cache.c
```diff
@@ -7,9 +7,10 @@ void test_cache_invalidate(TestContext *ctx) {
 	for (int i=0;i<32;i++) {
 		for (int j=0;j<32;j++) {
 
-			// Write the whole buffer through cache,
+			// Read/write the whole buffer through cache,
```
anacierdem, line 10: Why do we need to also read it?

tests/test_irq.c
```diff
@@ -0,0 +1,45 @@
...
+		cb1_called = true;
+		cb1_running = true;
+		disable_interrupts();
+
+		// Wait for the other timer interrupt to trigger
+		wait_ms(30);
+
+		// This enable_interrupts should not trigger the other
+		// interrupt, otherwise the test will fail.
+		enable_interrupts();
```
anacierdem, line 21: I am not able to verify why this does not immediately trigger the awating exception. I have tried checking the cause register but the interrupt pending bit for timer is not set just before we re-enable interrupts. What is preventing the secondary interrupt here?

anacierdem:
I have a few questions on my end to understand the behavior better.
Unfortunately the timer subsystem is messing with the time calculations (test finishes in 31:01, which is not true), we may need to fix that too, though that can wait for your next PR IMO.

tests/test_irq.c
```diff
@@ -0,0 +1,45 @@
...
+void test_irq_reentrancy(TestContext *ctx) {
+	// Make sure that enable_interrupts() does not cause an interrupt
+	// to trigger if called under interrupt.
+	volatile bool cb1_called = false;
+	volatile bool cb2_called = false;
+	bool cb1_running = false;
+	bool fail_reentrant = false;
+	bool fail_order = false;
+
+	void cb1(int ovlf) {
```
anacierdem, line 11: This is really cool!

tests/test_cache.c
rasky: That's a good question. I'm not 100% sure, I haven't studied internal cache workings. What I am sure of is that at some point I needed this to make the test pass. Obviously this is really unstable -- it's using a buffer on the stack and the stack might be already in the cache for previous functions -- but anyway I'm sure it was required.

Now that I think of it, I'll try to move the buffer to global data, so that it'll use separate cache lines and maybe I can reproduce this better

tests/test_irq.c
rasky: it's just that we are already under interrupt (status register, bit 1, EXL).

anacierdem: Ah I see, non-maskable interrupts are not subject to that 👍

PR #123, opened 2021-02-13
Rewrite timer.c to avoid messing up the COP0 hardware counter

rasky:
The previous implementation of timer.c assumed that it could change
the COP0 counter as it pleased. While this makes sense if timer.c is
the only user of the hardware counter, this behavior makes it impossible
for applications or other modules to use the hardware counter for their
own bookkeeping.

Obviously, one could object that timer_ticks() should be used instead
but timer_ticks() used to be unpredictably slow when interrupts were
disabled (it would invoke timer callbacks that might have expired)
and in general far slower than just reading the hardware timers, which
makes it a bad fit for high-performance scenarios like profiling of
critical sections.

This PR rewrites timer.c so that it keeps the COP0 counter flowing
naturally, and adjusts the COP0 compare register as required to
trigger the timer callbacks. This in turn makes it harder to provide
a true 64-bit timer; the PR implements it by registering an internal
timer synchronized with the 32-bit overflow, and then doing some careful
very racy manipulations.

Since I were at it, I also improved timer.c a little bit around corner
cases; for instance, if a timer expires while another timer's callback
was being executed, depending on the timer list order, the former
timer could have been ignored. Now this bug is fixed.

The code comes with a comprehensive testsuite that tests many different
corner cases; it's been hard to get this code right, which I think
make sense and should prove that the code is relatively safe.
Especially the 64-bit counter is stress-tested with fuzzying trying
to uncover latent race conditions.

In one of the commits, I also fixed a long-standing bug where the timer
interrupt was never enabled under Libdragon (!), and timer.c worked
"by chance" because typically VI interrupts were happening often and
at that point also pending timer interrupts were being serviced (but with
a random added latency).

### DIFF
--- a/include/cop0.h
+++ b/include/cop0.h
@@ -1,0 +1,69 @@
+/**
+ * @file cop0.h
+ * @brief N64 COP0 Interface
+ * @ingroup n64sys
+ */
+
+/**
+ * @addtogroup n64sys
+ * @{
+ */
+
+#ifndef __LIBDRAGON_COP0_H
+#define __LIBDRAGON_COP0_H
+
+/** @brief Read the COP0 Count register (see also TICKS_READ). */
+#define C0_COUNT() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$9":"=r"(x)); \
+    x; \
+})
+
+/** @brief Write the COP0 Count register. */
+#define C0_WRITE_COUNT(x) ({ \
+    asm volatile("mtc0 %0,$9"::"r"(x)); \
+})
+
+
+/** @brief Read the COP0 Compare register. */
+#define C0_COMPARE() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$11":"=r"(x)); \
+    x; \
+})
+
+/** @brief Write the COP0 Compare register. */
+#define C0_WRITE_COMPARE(x) ({ \
+    asm volatile("mtc0 %0,$11"::"r"(x)); \
+})
+
+
+/** @brief Read the COP0 Status register. */
+#define C0_STATUS() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$12":"=r"(x)); \
+    x; \
+})
+
+/** @brief Write the COP0 Status register. */
+#define C0_WRITE_STATUS(x) ({ \
+    asm volatile("mtc0 %0,$12"::"r"(x)); \
+})
+
+/* COP0 Status bits definition. Please refer to MIPS R4300 manual. */
+#define C0_STATUS_IE        0x00000001
+#define C0_STATUS_EXL       0x00000002
+#define C0_STATUS_ERL       0x00000004
+
+#define C0_STATUS_IM0       0x00000100
+#define C0_STATUS_IM1       0x00000200
+#define C0_STATUS_IM2       0x00000400
+#define C0_STATUS_IM3       0x00000800
+#define C0_STATUS_IM4       0x00001000
+#define C0_STATUS_IM5       0x00002000
+#define C0_STATUS_IM6       0x00004000
+#define C0_STATUS_IM7       0x00008000
+
+/** @} */
+
+#endif
--- a/src/inthandler.S
+++ b/src/inthandler.S
@@ -137,7 +137,11 @@ notprenmi:
 	and $30,k1,0x8000
 	beqz $30,notcount
 	nop
-	mtc0 $0,$11
+	/* Writing C0_COMPARE acknowledges the timer interrupt (clear the interrupt
+	   bit in C0_CAUSE, otherwise the interrupt would retrigger). We write
+	   the current value so that we don't destroy it in case it's needed. */
+	mfc0 k0,C0_COMPARE
+	mtc0 k0,C0_COMPARE
 
 	/* handle timer exception */
 	jal __TI_handler
--- a/src/timer.c
+++ b/src/timer.c
@@ -210,11 +210,13 @@ void timer_init(void)
 	}
 
 	/* Reset the count and compare registers. Avoid to accidentally trigger
-	   an interrupt by setting count to 1 and compare to 0. */
+	   an interrupt by setting count to 1 and compare to 0. Also enable
+	   timer interrupts in COP0. */
 	disable_interrupts();
 	ticks64_high = 0;
 	C0_WRITE_COUNT(1);
 	C0_WRITE_COMPARE(0);
+	C0_WRITE_STATUS(C0_STATUS() | C0_STATUS_IM7);
 	enable_interrupts();
 
 	register_TI_handler(timer_callback);
@@ -349,6 +351,9 @@ void timer_close(void)
 {
 	disable_interrupts();
     
+    /* Disable generation of timer interrupt. */
+	C0_WRITE_STATUS(C0_STATUS() & ~C0_STATUS_IM7);
+
     unregister_TI_handler(timer_callback);
 
 	timer_link_t *head = TI_timers;
--- a/src/timer.c
+++ b/src/timer.c
@@ -222,9 +222,8 @@ void timer_init(void)
 	C0_WRITE_COUNT(1);
 	C0_WRITE_COMPARE(0);
 	C0_WRITE_STATUS(C0_STATUS() | C0_STATUS_IM7);
-	enable_interrupts();
-
 	register_TI_handler(timer_callback);
+	enable_interrupts();
 }
 
 /**
--- a/src/timer.c
+++ b/src/timer.c
@@ -33,26 +33,39 @@
 
 /** @brief Internal linked list of timers */
 static timer_link_t *TI_timers = 0;
-/** @brief Total ticks elapsed since timer subsystem initialization */
-static long long total_ticks;
 
-/**
- * @brief Write the count to the count register
- *
- * @param[in] x
- *            Value to write into the count register
- */
-#define write_count(x) asm volatile("mtc0 %0,$9\n\t nop \n\t" :  : "r" (x) )
-/**
- * @brief Set the compare register
- *
- * This sets up the compare register so that when the count register equals the compare
- * register, an interrupt will be generated.
- *
- * @param[in] x
- *            Value to write into the compare register
- */
-#define write_compare(x) asm volatile("mtc0 %0,$11\n\t nop \n\t" :  : "r" (x) )
+/** @brief Higher-part of 64-bit tick counter */
+volatile uint32_t ticks64_high;
+
+/** @brief Time at which interrupts were disabled */
+extern volatile uint32_t interrupt_disabled_tick;
+
+/* @brief Timer is the special overflow timer. */
+#define TF_OVERFLOW    0x40
+
+/* @brief Timer has been called once in this interrupt. */
+#define TF_CALLED      0x80
+
+/* @brief Update the compare register to match the first expiring timer. */
+static void timer_update_compare(timer_link_t *head) {
+	uint32_t now = TICKS_READ();
+	uint32_t smallest = 0xFFFFFFFF;
+
+	while (head)
+	{
+		/* See how much time is left before the timer expires. Notice that
+		   the subtraction is also safe with overflows. */
+		uint32_t left = head->left - now;
+		if (left < smallest)
+			smallest = left;
+
+		/* Go to next */
+		head = head->next;
+	}
+
+	/* set compare to shortest time left */
+	C0_WRITE_COMPARE(now + smallest);
+}
 
 /**
  * @brief Process linked list of timers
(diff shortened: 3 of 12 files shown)


src/inthandler.S
```diff
@@ -137,7 +142,10 @@ notprenmi:
 	and $30,k1,0x8000
 	beqz $30,notcount
 	nop
-	mtc0 $0,$11
+	/* Acknowledge the timer interrupt by writing C0_COMPARE. We write
+	   the current value so that we don't destroy it in case it's needed. */
+	mfc0 k0,C0_COMPARE
```
anacierdem, line 143: I believe this is here to clear the interrupt, we may consider commenting that too. It is not apparent for someone who is not aware of that fact

src/timer.c
```diff
@@ -53,6 +57,41 @@ static long long total_ticks;
  *            Value to write into the compare register
  */
 #define write_compare(x) asm volatile("mtc0 %0,$11\n\t nop \n\t" :  : "r" (x) )
+/**
+ * @brief Read the compare register.
+ *
+ * @return the compare value
+ */
+#define COMPARE_READ() ({ \
```
anacierdem: I suggest moving this to n64sys.h where we keep system specific stuff.  And following your earlier suggestion, we may prefix it with something. TICKS_ or TIMER_ maybe?

src/timer.c
```diff
@@ -53,6 +57,41 @@ static long long total_ticks;
...
+
+/* @brief Update the compare register to match the first expiring timer. */
+static void timer_update_compare(timer_link_t *head) {
+    uint32_t now = TICKS_READ();
+    uint32_t smallest = 0xFFFFFFFF;
+
+    while (head)
+    {
+		uint32_t left = TICKS_DISTANCE(now, head->left);
+		if (left < smallest)
```
anacierdem, line 64: Isnt it better if we use a signed integer for the comparison? Is it possible to have negative distances here? Then positives will look smaller than the negative. E.g the initial smaller is -1 if it was signed and a distance of positive one will be smaller than FFFFFFFF

src/timer.c
```diff
@@ -140,31 +190,56 @@ static int __proc_timers(timer_link_t * head)
...
+		timer->callback = timer_callback_overflow;
+		timer->next = NULL;
+
+		TI_timers = timer;
+	}
+
+	ticks64_high = 0;
 	write_count(0);
-	write_compare(0x7FFFFFFF);
+	write_compare(0);
```
anacierdem: Does this cause the first interrupt right away? If that is the intention, we should make sure count and compare writes are on the same i cacheline and interrupts are disabled. Or if thats not the intention, we may consider handling the overflow at FFFFFFFF instead

src/timer.c
```diff
@@ -335,16 +397,50 @@ void timer_close(void)
...
 {
-	disable_interrupts();
-	timer_callback();					// force processing the timers
-	enable_interrupts();
-	return total_ticks;
+	uint32_t low, high;
+
+	/* Read status register */
+    uint32_t sr;
+    asm volatile("mfc0 %0,$12":"=r"(sr));
```
anacierdem: We may also have a macro for this on n64sys

src/timer.c
```diff
@@ -335,16 +397,50 @@ void timer_close(void)
...
+	} else {
+		/* Interrupts are currently disabled. If they've been disabled for more
+		 * than 2**32 ticks, it's game over because we can't know how many times
+		 * the counter has overflown.
+		 * So assuming they were disabled for not too long, check whether there
+		 * was a counter overflow between now and the last time an
+		 * interrupt triggered. If there was, increment high. */
+	    low = TICKS_READ();
+	    high = ticks64_high;
+		if (last_interrupt_tick > low)
```
anacierdem: This requires a really long time without interrupts, maybe we shouldnt even bother and just return high. If ints are disabled that long it is highly unlikely that something will try to read the timer just before we lost track of time. I mean waiting a few more usecs wll already invalidate the result

tests/test_timer.c
```diff
@@ -0,0 +1,248 @@
...
+	DEFER(delete_timer(tt1));
+
+	wait_ms(2);
+	ASSERT_EQUAL_SIGNED(cb1_called, 1, "timer 1 not called");
+	wait_ms(3);
+	ASSERT_EQUAL_SIGNED(cb1_called, 1, "timer 1 called again?");
+
+	// Restart the timer. This time, wait with interrupts disabled, so that
+	// we check the the timer triggers as soon as we re-enable them.
+	stop_timer(tt1);
```
anacierdem: We may also assert it is not called again when stopped as an additional sanity check

tests/test_timer.c
```diff
@@ -0,0 +1,248 @@
...
+	start_timer(tt1, TICKS_FROM_MS(3), TF_ONE_SHOT, cb1);
+	wait_ms(2);
+	ASSERT_EQUAL_SIGNED(cb1_called, 1, "timer 1 called again?");
```
anacierdem, line 28: It should not be called here. The assertion msg seems off

tests/test_timer.c
```diff
@@ -0,0 +1,248 @@
...
+		"invalid timer_ticks");
+}
+
+// Change the hardware count register.
+static uint32_t write_count(uint32_t x) {
+	extern uint32_t last_interrupt_tick;
+
+	uint32_t old = TICKS_READ();
+
+	asm volatile("mtc0 %0,$9\n\t nop \n\t" :  : "r" (x) );
```
anacierdem: The nop should not be required

tests/test_timer.c
```diff
@@ -0,0 +1,248 @@
...
+void test_timer_ticks(TestContext *ctx) {
+	timer_init();
+	DEFER(timer_close());
+
+	// We want to fuzz different conditions around the overflow, that is
+	// when the hardware counter goes to zero. That's where all problems lie
+	// in the implementation. Try first to be exhaustive. Also, check with
+	// both enabled and disabled interrupts because the codepaths are different.
+	for (int j=0; j<2; j++) {
+		for (int i=0; i<512; i++) {
```
anacierdem, line 175: Is 512 choosen b/c it is longer than the exec. time here?

tests/test_timer.c
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void test_timer_ticks(TestContext *ctx) {
+	timer_init();
+	DEFER(timer_close());
+
+	// We want to fuzz different conditions around the overflow, that is
+	// when the hardware counter goes to zero. That's where all problems lie
+	// in the implementation. Try first to be exhaustive. Also, check with
+	// both enabled and disabled interrupts because the codepaths are different.
```
anacierdem, line 173: I think we are not testing inside of the last interrupt tick condition if Im  not mistaken. We can start a timer, then simulate a very long interruptless period writing over the count while ints are disabled. That condition might be a very edge case though, so might be better to just remove it 😄

anacierdem:
This looks great! I have a few comments. If we can go over them I think we can improve things a little. Feel free to prove me wrong where Im mistaken

src/inthandler.S
rasky: "Acknowledge the timer interrupt" means exactly that. If you prefer I can change "Acknowledge" into "Clear"

src/timer.c
rasky: Do we want to expose this as public API? If we do, maybe we want to have a `cop0.h` with all the macros with `COP0_` prefix, and then `TICKS_READ` can simply call `COP0_COUNTER_READ`. What do you think?

src/timer.c
rasky: No we cannot use a signed integer here. We need to calculate how many ticks are left before a timer triggers. If a timer triggers at 0 and now is 0x100, we need to calculate 0xFFFFFE00 which is a very large (unsigned) number, which is correct because there are 4294966784 cycles left before the counter becomes 0 again.

src/timer.c
rasky: No there's no need to trigger an interrupt because we're just resetting. The sequence does not trigger an interrupt because first count is written, and then compare, and by the time compare is set to 0, count is at least 1. Now that you mention, it's probably a little "edgy" and can also confuse emulators. Maybe we should just set count to 1 and keep compare to 0.

I don't get your comment about handling the overflow at FFFFFFFF instead. Can you elaborate on why you think it's necessary?

src/timer.c
rasky: No actually it doesn't require such a long time. Consider this sequence:

 * FFFFFE00: VI interrupt triggers (last interrupt)
 * FFFFFE88: user disables interrupts
 * 00000010: user calles `timer_ticks()`
 
With the code as in the PR, `timer_ticks()` will return `(high+1) << 32 + 0x10` which is indeed correct because an overflow occurred. The code realizes that an overflow occurred because it knows that the timer was recently FFFFFE00, so a timer of 10 indicates an overflow. On the contrary:

 * 00000100: VI interrupt triggers (last interrupt)
 * 00000188: user disables interrupts
 * 000001AA: user calles `timer_ticks()`

In this case, `high` would not be incremented because the timer was recently 0x100, so 0x1AA has not overflown yet.

tests/test_timer.c
rasky: yes, it's longer the execution time, so since I'm testing all starting points from 0 backward to -512, theoretically I'm generating overflows in all possibile situations.

tests/test_timer.c
rasky: I hope the explanation above clarifies. I for sure am testing that condition because it happens any time there's an overflow while interrupts are disabled, and the test is executed with and without interrupts for this very reason.

src/inthandler.S
anacierdem: I needed to go check the document to understand it but that might be b/c I'm rather new to the terminology. Might be clearer for the ones like me :) We can even explicitly write that behavior (writing to the compare registers clears the interrupt bit)

src/timer.c
anacierdem: Hmm, we may expose only the read? I'm not sure if that's useful indeed maybe we don't need to expose it you're right. If you have a valid use case for that in mind w may consider `COP0_` prefix.

src/timer.c
anacierdem: Then using a plain difference (`head->left - now`) should also result in the same behavior and might be better for the reader considering the documentation (and resulting type) for `TICKS_DISTANCE`.

rasky:
@DragonMinded I'd love it if you could find time to review this since you wrote the original module

DragonMinded:
I don't remember anything about this module so your reviews are going to be better than anything I could offer.

src/timer.c
anacierdem: I didn't get why count is 1 when we write the compare. It is being updated at half maximum instruction rate, so I expect it to not being incremented if we are on the other half. This was what I'm thinking;
```
<-- icrement was on this cycle
write_count(0); # <-- we reset to 0
write_compare(0); # <-- we have written compare but it is now 1 and we're safe
```
The above case is safe. But what if;
```
write_count(0); # <-- icrement was on this cycle, but we reset to 0
write_compare(0); # <-- we have written compare while count was 0, will this trigger an interrupt?
```

If the interrupt only fires when count is incremented, this should not be an issue, but if it is fired when they are equal, this maybe a problem. I couldn't find more info on the exact behavior.

I meant using FFFFFFFF as the initial compare value so that it wont cause an issue for sure till we setup things. And I thought this was also related to overflow timers trigger point for a moment, and we might use it as the trigger value for the overflow timer

src/timer.c
anacierdem: Ah, I thought we were updating `last_interrupt_tick` only if it is a timer interrupt. Why not only update it for a timer interrupt? Then you're still right there may be many timers registered that may trigger at any time.

src/inthandler.S
anacierdem: Also we need 2 nops between the two b/c cop0 hazards if Im not mistaken

src/timer.c
anacierdem: Im not even sure on the first case now as I dont know the exact operation (is count incremented before we write compare?). If you are sure it wont trigger this is ok.

anacierdem: Also what is the initial value for compare? What if we -by chance- are setting count to it? Maybe we should disable ints, set count to 1 compare to 0 and enable. This will make sure we dont trigger it. Even if initial compare is 1, ints are disabled and we write compare next which will clear the interrupt bit before we re-enable.

anacierdem: One last thing: we can only enable timer interrupt on timer init instead of init interrupts so that we dont cause unnecessary interrupts.

src/inthandler.S
rasky: I changed the comment. I don't think we need an explicit nop here: hazards are because of pipeline delays within COP0; in this case, the R4300 knows to insert the pipeline stall because k0 won't be available for a couple of cycles.

src/timer.c
rasky: Thinking about this, one of the thing that I want to fix in libdragon is that I'd like it to expose hardware access more, rather than only wrap with high-level libraries. The latter are obviously fine and a good goal, but the fact that no hardware access is often allowed / defined is too constraining.

So this is one of those cases where I think it makes sense to start exposing hardware details through a standard interface. I've added a cop0.h (included by n64sys.h, since it's a system interface) with standard macros for COP0 access. It can grow over time, but we can start using ourselves. I've updated interrupt.c and timer.c to use it rather than directly using inline assembly.

src/timer.c
rasky: Done

src/timer.c
rasky: Done.

src/timer.c
rasky: Done.

src/timer.c
rasky: `last_interrupt_tick` is just a recent timer snapshot, so it doesn't really matter how often we update it. 

Thinking more of it, what we really want is to know when interrupts were disabled. I've changed `last_interrupt_tick` into `interrupts_disabled_tick`, modified only by `disable_interrupts()`. This fails if the user disables interrupts without going through `disable_interrupts()` but I think that's already not fully supported anyway, so we might ignore it.

I hope it's more clear now. It also allows to remove the `last_interrupt_tick` modification from the test when the timer is jumping around, as we just now record the correct value when disabling the interrupts anyway.

tests/test_timer.c
rasky: Done.

tests/test_timer.c
rasky: Yes, I'm using the "called again?" assertion whenever it was called when it was not supposed to.

rasky (commented):
Done, I think I've implemented all requested changes.

tests/test_timer.c
anacierdem: The "again" part was confusing to me but not a big deal

anacierdem:
Thank you for the changes. Ill go over it one last time and also test it on my environment.

src/inthandler.S
anacierdem: You are right, it is only required for the reverse sequence (mtc,mfc)

include/cop0.h
```diff
@@ -0,0 +1,70 @@
...
+
+/** @brief Read the COP0 Compare register. */
+#define C0_COMPARE() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$11":"=r"(x)); \
+    x; \
+})
+
+/** @brief Write the COP0 Compare register. */
+#define C0_WRITE_COMPARE(x) ({ \
```
anacierdem, line 36: I think we should wrap these with `do { ... } while (0)`, What do you think?

include/cop0.h
```diff
@@ -0,0 +1,70 @@
...
+#define C0_STATUS() ({ \
+    uint32_t x; \
+    asm volatile("mfc0 %0,$12":"=r"(x)); \
+    x; \
+})
+
+/** @brief Write the COP0 Status register. */
+#define C0_WRITE_STATUS(x) ({ \
+    asm volatile("mtc0 %0,$12"::"r"(x)); \
+    asm volatile("nop"); \
```
anacierdem: We don't need this. If this is for cop0 hazards, it is dependent on the next instruction being executed, so we probably shouldn't have it here on this helper.

anacierdem (commented):
Final notes;

- We may also want to write somewhere (we may add it to Timer Subsystem description) that maximum allowed timer length is ~90 secs.
- Let's decide on a tab/space convention for the future
- I have tested the changes and they are working great.

include/cop0.h
rasky: The statement expression `({ ... })` is a GNU extension that has the same effect and also allows to return a value. I'd settle for using it everywhere in macros because it's more powerful.

include/cop0.h
rasky: Yes, it's for hazards. Those bugs are really hard to debug, and one additional nop in status registers manipulation seems innocuous. I think it's better to have it as a default. If there are situations where it must absolutely be absent (not sure why), I think those are the exceptions rather than the rule, and the user will know to simply call `mtc0` themselves without the macro.

anacierdem: Maybe it is safer to have two then? That will protect from most of hazards (except TLB, which dragon currently does not use)

PR #124, opened 2021-02-19
Fix 16k EEPROM detection

meeq:
This will not break existing callers of `eeprom_present` since 4k EEPROM still maps to `1`

`eeprom_read` and `eeprom_write` appear to work correctly for 16k EEPROM

### DIFF
--- a/include/controller.h
+++ b/include/controller.h
@@ -45,6 +45,19 @@
 /** @} */
 
 /**
+ * @name EEPROM Probe Values
+ * @see #eeprom_present
+ * @{
+ */
+/** @brief No EEPROM present */
+#define EEPROM_NONE          0
+/** @brief 4 kilobit (64-block) EEPROM present */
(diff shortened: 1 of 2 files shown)


include/controller.h
```diff
@@ -44,6 +44,19 @@
...
+ * @see #eeprom_present
+ * @{
+ */
+/** @brief No EEPROM present */
+#define EEPROM_NONE          0
+/** @brief 4 kilobit (64-block) EEPROM present */
+#define EEPROM_4k            1
+/** @brief 16 kilobit (256-block) EEPROM present */
+#define EEPROM_16k           2
+/** @} */
```
rasky, line 58: Minor nit: we should keep the macros uppercase, so this should be `EEPROM_4K` and `EERPOM_16K`.

rasky (approved):
Fine by me!

PR #125, opened 2021-02-24
Implement EEPROM Filesystem and test ROM

meeq:
EEPROM allows up to either 512 bytes or 2048 bytes to be stored, retrieved, and persisted across resets and power-offs.

EEPROM FS allows homebrew to list out distinct "files" that map to one or more blocks of EEPROM and read or write to these files.

The impetus and proof-of-concept for this was porting a PC game that used stdio calls. With the EEPROM filesystem, all I had to do was configure the game’s save path to `eeprom://Profile.dat` and the game was able to read and write save data without making any changes to the original file handling code.

This is intended to be a companion to DragonFS that allows games to easily save data such as structs and arrays using a friendly file-like interface. ~Just like DragonFS, EEPROM FS can be used either with POSIX functions through a path prefix (e.g. `eeprom:/`) or using the `eepfs_*` functions.~ 

Originally, this PR included stdio hooks to access EEPROM files, but this turned out to be a leaky abstraction that had some rough edge-cases. To simplify the implementation, EEPROM FS currently only supports reading/writing/erasing entire files using the `eepfs_*` functions.

Due to the limitations of the storage medium, there are a few considerations to keep in mind:

* All file names and file sizes must be declared before initialization
  * It is not possible to add or remove files or change their size
  * All files will always exist at their declared size
  * When a file is "erased", all blocks are zeroed out
* Files must start on a block boundary
  * In practice, this limits the number of files to the EEPROM's block capacity minus 1 (the first block of EEPROM is reserved by EEPROM FS)
  * Files that are not aligned to the EEPROM block size (8-bytes) will cause the unaligned bytes to be wasted as padding.
* Changes made by you (the homebrew developer) between builds/versions of your homebrew may require EEPROM data to be wiped or migrated.
  * EEPROM FS reserves the first block of EEPROM for a "signature" which is used to check if the configured filesystem is roughly compatible with the saved data. This means that if the number of files or the total file size of your configured filesystem changes then the data in EEPROM may no longer line up with the new filesystem and needs to be erased (easy) or migrated (non-trivial).
  * It is possible (but not recommended) to skip this signature check, but the data read through EEPROM FS may be nonsensical.

See `examples/eepromfstest` for some contrived examples of storing and loading data on both 4k (64-block) and 16k (256-block) EEPROM.

The example detects which EEPROM (if any) is available, offers to wipe the existing data. and then attempts to run through reading, then writing, then reading back again with three different data structures:

* A list of "high scores" (an array of structs)
* A simple "global settings" struct
* A "saved game state" struct

For the 16k EEPROM, the saved game state struct is used to demonstrate how multiple save "slots" could be accomplished.

Note that [cen64 is currently hard-coded to report a 4k EEPROM](
https://github.com/n64dev/cen64/blob/89e47d2968af2227017e52008d6c893e58b300c5/si/controller.c#L124-L132) even if it is not specified on the command-line, but will fail the read/write tests unless you explicitly set the save type with `-eep4k` or `-eep16k` and a path the save the EEPROM data.


meeq:
Pre-built test ROM attached: [eepromfstest.zip](https://github.com/DragonMinded/libdragon/files/6055063/eepromfstest.zip)

meeq:
Updated test ROM based on https://github.com/DragonMinded/libdragon/pull/125/commits/abfbe0ab4c7b697b188ad2eded2dba0deacac7ed: [eepromfstest.zip](https://github.com/DragonMinded/libdragon/files/6056214/eepromfstest.zip)

examples/eepromfstest/eepromfstest.c
```diff
@@ -0,0 +1,495 @@
...
+        { "high_scores.dat", sizeof(game_high_scores)  },
+        { "settings.dat",    sizeof(game_settings_t)   },
+        { "saves/slot1.sav", sizeof(game_save_state_t) },
+        { "saves/slot2.sav", sizeof(game_save_state_t) },
+        { "saves/slot3.sav", sizeof(game_save_state_t) },
+        { "saves/slot4.sav", sizeof(game_save_state_t) }
+    };
+    eepfs_config_t eeprom_16k_config = {
+        .files = eeprom_16k_files,
+        .num_files = EEPFS_NUM_FILES(eeprom_16k_files),
```
rasky: Rather than storing a separate counter of elements, another option is to have the files array to be terminated with a NULL entry. This might be easier to use for user code.

Otherwise, I'm not sure I like this macro. This macro is typically called `countof()` or `ARRAY_SIZE`, and unfortunately it's not part of standard C yet. Anyway, I'd let users figure it out by themselves, rather than providing a `EEPFS_NUM_FILES` macro which seems that contains something specific of eepfs. I think the above code is more clear if you just put a straight `6` here. The macro made me wonder what was hidden behind it (turns out, nothing).

examples/eepromfstest/eepromfstest.c
```diff
@@ -0,0 +1,495 @@
...
+        .stdio_prefix = "eeprom:/",
```
rasky: The problem with having this configuration here is that the whole stdio filesystem code is always linked to the binary, irrespective of game code actually using it or not.

I suggest you move this functionality into a separate `eepfs_mount()` function. This way, if the game code only calls `eepfs_read/write`, the stdio functionality will not be linked in the final binary.

include/eepromfs.h
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * @brief Whether to validate EEPROM data during initialization.
+     * 
+     * If this is false and the signature does not match,
+     * all EEPROM data will be erased!
+     * 
+     * @see #eepfs_check_signature
+     */
+    const bool skip_signature_check;
+} eepfs_config_t;
```
rasky: All these functions declare their members as `const`. This makes the structures more complex to handle for user code (they need to go through the hoops to initialize them) and I'm not sure what kind of guarantee or safety is providing. I think it's enough to document that the config structure should not be modified after eepfs_init() is called (which, if you ask me, is already clear when you first read this header file).

Also, if you agree on moving out `stdio_prefix` and removing `num_files` (using the NULL member as sentinel), this structure just contains `files` and `skip_signature_check`. At which point,  I would just pass those two arguments to `eepfs_init()`. I would expect most user code to declare their `eepfs_entry_t` array as a global const variable and then use it. I think this simplifies the overall design.

include/eepromfs.h
```diff
@@ -0,0 +1,139 @@
...
+
+int eepfs_init(const eepfs_config_t config);
+int eepfs_deinit();
```
rasky: This should be called `eepfs_close` to match the naming of all other libraries.

Also notice that in C, declaring a function as `()` means that it takes a variable number of arguments (it's basically equivalent to `(...)`. If you want the function to accept zero arguments, you must explicitly declare it as `(void)`. I know this is a common bug in libdragon, and I'll eventually get around fixing it.

include/eepromfs.h
```diff
@@ -0,0 +1,139 @@
...
+
+int eepfs_read(const char * const path, void * const dest);
+int eepfs_write(const char * const path, const void * const src);
+int eepfs_erase(const char * path);
```
rasky, line 82: `const char* path` is fine, `const char * const path` has an additional constraint which is not very useful and in fact is not used everywhere else in libdragon.

src/eepromfs.c
```diff
@@ -0,0 +1,888 @@
...
+    const uint8_t num_files = eepfs_num_files - 1;
+
+    /* Sum up the total bytes used by all files.
+       At most a 16k EEPROM can hold 2048 bytes. 
+       Skip the first file (the signature block). */
+    uint16_t total_bytes = 0;
+    for ( size_t i = 1; i < eepfs_num_files; ++i )
+    {
+        total_bytes += eepfs_files[i].max_cursor;
+    }
```
rasky, line 131: A non-table-based CRC16 is marginally more complex but it's more solid; eg, it detects if you keep the same files but rearrange the order, which I would say it can be a common "mistake" made by game code.

For a simple implementation, see https://stackoverflow.com/a/23726131/118015

include/eepromfs.h
```diff
@@ -0,0 +1,139 @@
...
+
+void eepfs_wipe();
+int eepfs_check_signature();
```
rasky: I'd use `bool` here as return value (remember to include `stdbool.h`).

include/eepromfs.h
```diff
@@ -0,0 +1,139 @@
...
+ * @file eepromfs.h
+ * @brief EEPROM Filesystem
+ * @ingroup eepfs
+ */
+#ifndef __LIBDRAGON_EEPROMFS_H
+#define __LIBDRAGON_EEPROMFS_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include "controller.h"
```
rasky: Does this header file really need `controller.h`? I don't see dependencies.

src/eepromfs.c
```diff
@@ -0,0 +1,888 @@
...
+ * @param[in]  path
+ *             Path of file in EEPROM filesystem to read from
+ * @param[out] dest
+ *             Buffer to read into
+ * @param[in]  len
+ *             Number of bytes to read
+ *
+ * @return The actual number of bytes read or a negative value on failure.
+ */
+static int eepfs_cursor_read(eepfs_file_t * const file, void * const dest, const size_t len)
```
rasky: I think in general you may want to keep a cached 8-byte block in `eepfs_file_t` and only read/write it when it's complete (and / or the user seeks away).

This is not a real problem if you use the `eepfs_read` / `eepfs_write` functions where everything is read/written in a single shot, but it may become critical for people using the filesystem API. The EEPROM is extremely slow to access (~15ms per block), and can become worn. The stdio API is normally buffered, so people will assume that it's OK to, eg., read 1 byte at a time. Instead, without an internal buffering of a block, this will be extremely slow (and writing 1 byte at a time would reduce the EEPROM life).

I think that if we provide a filesystem API, the onus is on us to provide a buffering layer that enforces correct practices.

src/eepromfs.c
```diff
@@ -0,0 +1,888 @@
...
+        /* Fill the current block with the data to write */
+        while ( bytes_to_write > 0 && block_byte_offset < EEPROM_BLOCK_SIZE )
+        {
+            eeprom_buf[block_byte_offset] = *(ptr++);
+            bytes_written++;
+            bytes_to_write--;
+            block_byte_offset++;
+        }
+        /* Write the block and move on to the next one */
+        eeprom_write(current_block, eeprom_buf);
```
rasky: Given the above comment on caching the current block, I would defer actual writing to when the user has issued a write commond on a different block, so that each block is normally written only once.

src/eepromfs.c
```diff
@@ -0,0 +1,888 @@
...
+    uint8_t * const sig_bytes = (uint8_t *)&signature;
+    sig_bytes[0] = 'e';
+    sig_bytes[1] = 'e';
+    sig_bytes[2] = 'p';
+    sig_bytes[3] = 'f';
+    sig_bytes[4] = 's';
+    sig_bytes[5] = num_files;
+    sig_bytes[6] = total_bytes >> 8;
+    sig_bytes[7] = total_bytes & 0xFF;
+    return signature;
```
rasky, line 144: This signature is checking for consistent filesystem structure, and I totally see its value.

Another common issue with EEPROM access (not only in N64 world) is that it's basically required to have some sort of consistency checking on actual data being read/written because those things wear out and consistency is never absolutely guaranteed. I think Nintendo was also advising developers to absolutely keep internal checksums in EEPROM data.

This can be done in a followup PR if you want, as it's an additional feature at the end of the day. I'm not sure how it would affect the API though. Some considerations:

 * Reading a 8-byte block of the EEPROM takes ~15ms. This is very slow. Reading a full 128K EEPROM would require almost 4 seconds, so it doesn't sound like validating the content is something that can be done at init time.
 * If each file had a 16bit CRC associated with it (either appended at the start/end of actual data, or in a separate table to be kept in the EEPROM), it would be easy to read/update it when `eepfs_read` / `eepfs_write` are called. That is a good point to do it because the code is already going through all the content bytes.
 * For the stdio filesystem, it's less clear how this would work. Given the stdio API, users might read / write a file only partially, seek through it, etc. so there's not good point where the CRC can be calculated over the whole content. I thought of something quite convoluted/"smart" that might work, but not sure if it's worth the effort:
    * Each file has a 16bit "checksum" associated with it. This can be stored in the header of the filesystem or next to each file, it doesn't matter.
    * This "checksum" is calculated block by block: calculate the CRC16 of each 8 byte block, then xor all the resulting CRC16 together. This means that you can calculate this magic checksum in any order, as long as all blocks of the files are processed.
    * When reading, the cursor API keeps a bitmask of processed blocks. As soon as a new block is fetched, if it's the first time, the CRC16 is calculated and a XOR checksum is updated. Also the block is flagged as "done" in the bitmask. As soon as the bitmask is full (all blocks are processed), all I/O functions start returning error if the checksum doesn't match. As a last ditch, fclose() would read all missing blocks and return error if the checksum doesn't match (though nobody checks fclose... so is it worth it? Maybe yes).
    * When writing, something similar happens. At fclose time, all blocks that were not written are read once, so that their CRC can be calculated and the final checksum can be updated.
    * Maybe checksum checks can be disabled altogether at either mount time, or open time (ab)using a POSIX flag. One option would be to reuse `O_EXCL` (which is mapped to `x` in `fopen` format strings) to signal "I don't want checksum calculation while reading / writing". 
       * if the checkum is "appended" at the end of each file, one option would be for `O_EXCL` to switch to "raw mode" where the checksum is returned as part of the actual file content, and performing zero calculations on it. This would also to document to always use `O_EXCL` if the game code wants to perform its own checksum on data and completely bypass the filesystem support.

rasky (changes requested):
Thanks, I've done a first round of general architecture review (following our Discord discussion last night). I haven't actually reviewed all the function implementations yet, I'm still focusing on the API design and overall structure. 

I would say the other main issue I see with this API (which is not your fault at all) is that it provides synchronous blocking access to EEPROM. Reading/writing the EEPROM is very slow (~15ms per block), and doing it while blocking the CPU means that the game would not be allowed to perform audio / video operations in background as the EEPROM is accessed. This is actually quite common on some consoles (eg: Playstation 1), but nonetheless is a problem.

libdragon doesn't support threading and most APIs are blocking, so I can see that we're not providing any foundation to solve this problem. I'm actually going to submit soon a very simple async / future / completion "library", which is a small set of helper functions to provide a unified API for asynchronous operations. This is how it would work for this library:

```
    async_t op = async_new(my_callback, NULL);
    eepfs_write("scores.dat", &scores, &op);
```

or if you want to block later:

```
    async_t op = async_new(NULL, NULL);
    eepfs_write("scores.dat", &scores, &op);
    [...]
    while (!async_is_finished(&op)) {}
```

or even easier:
```
    eepfs_write("scores.dat", &scores, ASYNC_BLOCK);
```

To support this, internally `eepfs` (and `controller.c` in the first place) would have to refactored to use chained interrupts to perform operations.

I'm just mentioning this as a general comment, I'm sure this PR can go in the way it is right now, with the only gotcha that we might have to change the API down the road.

examples/eepromfstest/eepromfstest.c
anacierdem: The game code will highly likely use the dfs already and link against stdio. On the other hand, the devs might want to develop a custom filesystem for their needs and may also want to simply use the eeprom and they will have to ship it with stdio even though they don't need dfs. So I agree keeping the binding out of `eepfs_init` is a good idea.

rasky: I was referring to the stdio-related code in eepromfs. Might be small but everything counts.

include/eepromfs.h
anacierdem: Do you think it is not useful b/c it will prevent the implementation from using that pointer? it should be enough if we don't manipulate the contents by mistake right?

include/eepromfs.h
anacierdem: Is that for semantics only?

src/eepromfs.c
rasky: Now that I think of it: maybe `fopen` in newlib already provides a buffering layer? I'm not sure how that interacts with eepfs and I'm sure it's not designed around the concept of underlying fixed size blocks, but maybe it's something to take into account.

include/eepromfs.h
rasky: Yes. We have bool in C now, so I think it's more clear if we use it whenever the return value is actually a boolean. It's one less think to look up in documentation.

src/eepromfs.c
anacierdem: Then we may also need a flush function to gracefully write pending writes in order not to loose information IMO.

src/eepromfs.c
rasky: I checked newlib sources and it looks like newlib by default adds a 1Kb buffer to all stdio files. I'm not sure on how that interacts with seeking for instance, but for simple cases it looks like this is useful for us because it means that we would probably get a single read/write call for the whole file as soon as the buffer is flushed (at fclose time, or in case of explicit fflush -- and then I don't know if seeking also causes a flush). This is something I'd love to be investigated and documented though.

anacierdem:
@rasky thanks for the review, I'm rather slow at reviewing C code (and I'm pretty sure it will not be as good as yours) so it is really appreciated. @meeq Thanks for the implementation! 🥇  Very few are opening PRs and we don't have the time to collect them from the forks unfortunately. We are trying hard to improve libdragon as much as possible so please be patient with the reviews, it is for the greater good.

include/eepromfs.h
meeq: `dragonfs` uses `const char * const path`, which was a strong influence on the `eepromfs` API:

https://github.com/DragonMinded/libdragon/blob/87bcff52696abe21a1a322a895677d26ce764e46/include/dragonfs.h#L89

include/eepromfs.h
meeq: Yes, `controller.h` is where `eeprom_read`, `eeprom_write`, `eeprom_total_blocks`, and `EEPROM_BLOCK_SIZE` are declared

include/eepromfs.h
meeq: ### Regarding `const` everywhere:

No objection to removing `const` from the public struct declarations. My heuristic is to `const` everything unless there's a good reason not to, but you're right that it's unnecessary and potentially annoying here.

### Regarding `stdio_prefix`:

No objection to removing this argument and moving it to its own function.

### Regarding `num_files`:

I think that a `NULL` member as a sentinel is potentially more confusing and error-prone than requiring an explicit length.

I cannot think of a good way of enforcing the presence of the sentinel, and its absence would lead to very unexpected results.

Typically C APIs that accept an array expect callers to also provide a length; the big exception is strings, which have a long and sordid history of issues related to the accidental or intentional omission of `\0`.

examples/eepromfstest/eepromfstest.c
meeq: `EEPFS_NUM_FILES` macro has been removed.

I considered the NULL sentinel value idea, but could not come up with a simple and practical way to ensure the array is properly terminated.

examples/eepromfstest/eepromfstest.c
meeq: Attaching the filesystem to stdio is now done through a separate `eepfs_attach` function

include/eepromfs.h
meeq: `eepfs_config_t` has been removed.

The option to disable the signature check has been removed to simplify the API. Upon further reflection, there is no good reason to skip the signature verification. If the signature does not match, it's likely that the filesystem will return unexpected garbage data.

include/eepromfs.h
meeq: Fixed for all eepromfs functions that do not accept arguments.

src/eepromfs.c
meeq: Implemented and included in the filesystem signature. Thanks!

include/eepromfs.h
meeq: Fixed

src/eepromfs.c
meeq: This is an interesting problem that I'll definitely give some more thought and research into.

I considered doing some sort of checksum for each file (or each block) but ultimately punted on the idea due to the new realities of EEPROM save reliability in the age of emulators and flash carts. The vast majority of people running games/software built by libdragon are likely not actually using a cartridge with real EEPROM.

There are some programmable cartridges out there (the [N64 Blaster by RetroStage](https://retrostage.net/index.php/product/n64-blaster-2-0/), and the [N64 Project by RetroCircuits](https://www.mortoffgames.com/index.php?route=product/product&path=294&product_id=877)) which support EEPROM saves, but it's unclear whether they actually use EEPROM (further investigation needed).

I think it's a worthwhile follow-up to explore how best to effectively implement checksums with minimal overhead. Hopefully it will be possible to drop-in without much (or any) of a noticeable change on the API side.

include/eepromfs.h
meeq: As part of integrating this PR back into the PC game port that originally spawned the idea for EEPROM FS, I have encountered an objection to removing the `const` qualifier on `eepfs_entry_t`'s `path` member: 

`ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]`

In order to make `eepfs_config_t` compatible with C++, `path` needs to be `const char *`

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+ */
+#include <stdio.h>
+#include <string.h>
+#include <stdint.h>
+#include <malloc.h>
+#include <sys/stat.h>
+#include "libdragon.h"
+#include "system.h"
+
+#define divide_ceil( x, y ) ((x / y) + (x % y != 0))
```
anacierdem: `#define divide_ceil( x, y ) (((x)/(y)) + ((x) % (y) != 0))` for safer precedence

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+    }
+
+    size_t cursor = file->cursor;
+    size_t bytes_to_read = len;
+
+    /* Bounds check to make sure it doesn't read past the end */
+    if ( cursor + bytes_to_read >= file->max_cursor )
+    {
+        /* It will, let's shorten it */
+        bytes_to_read = file->max_cursor - cursor;
```
anacierdem: Maybe we should return an error here as well. The buffer will not be filled by len bytes and that may cause confusion.

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+    uint8_t eeprom_buf[EEPROM_BLOCK_SIZE];
+    uint8_t * ptr = dest;
+
+    do {
+        eeprom_read(current_block, eeprom_buf);
+        /* Fill the buffer with the data from the current block */
+        while ( bytes_to_read > 0 && block_byte_offset < EEPROM_BLOCK_SIZE )
+        {
+            *(ptr++) = eeprom_buf[block_byte_offset];
+            bytes_read++;
```
anacierdem: As `eepfs_file_t`'s contents are writable now, we may just mutate it here to track position. We already do that at the end so that might be easier to read.

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+        return 0;
+    }
+
+    /* Calculate the starting point to read out of EEPROM from */
+    size_t current_block = file->start_block + (cursor / EEPROM_BLOCK_SIZE);
+    size_t block_byte_offset = cursor % EEPROM_BLOCK_SIZE;
+    int bytes_read = 0;
+
+    uint8_t eeprom_buf[EEPROM_BLOCK_SIZE];
+    uint8_t * ptr = dest;
```
anacierdem: non-const pointers may help with these. I think @rasky's suggestion was to simplify these

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+    }
+
+    size_t cursor = file->cursor;
+    size_t bytes_to_write = len;
+
+    /* Bounds check to make sure it doesn't write past the end */
+    if ( cursor + bytes_to_write >= file->max_cursor )
+    {
+        /* It will, let's shorten it */
+        bytes_to_write = file->max_cursor - cursor;
```
anacierdem: Again we may consider an error here

src/eepromfs.c
```diff
@@ -0,0 +1,933 @@
...
+    if ( new_cursor > max_cursor )
+    {
+        new_cursor = max_cursor;
+    }
+    if ( new_cursor < 0 )
+    {
+        new_cursor = 0;
+    }
+
+    file->cursor = new_cursor;
```
anacierdem: We can set the cursor in place as it is not a const anymore.

anacierdem (commented):
I think this is ok to merge as is. I have a few comments to clear some things up but I will not insist on any of them. The API is ok to begin with and is extensible with callbacks or async handles later on. const pointers are also easy to update if we want so.

src/eepromfs.c
meeq: I borrowed this pattern from `dragonfs.c`: https://github.com/DragonMinded/libdragon/blob/87bcff52696abe21a1a322a895677d26ce764e46/src/dragonfs.c#L949-L957

src/eepromfs.c
meeq: I borrowed this pattern from `dragonfs.c`:
https://github.com/DragonMinded/libdragon/blob/87bcff52696abe21a1a322a895677d26ce764e46/src/dragonfs.c#L930
https://github.com/DragonMinded/libdragon/blob/87bcff52696abe21a1a322a895677d26ce764e46/src/dragonfs.c#L981-L982

anacierdem: I'd love to hear @rasky's thoughts on this.

src/eepromfs.c
anacierdem: It is a really old code and we don't have to duplicate that, but for the sake of consistency I think this is ok for now.

anacierdem:
Another note: I don't think backwards compatiblity is much of an issue for an open source library on an archaic system, but still the first 3 bytes of the signature can act as a version to convert between different versions if someone really has to keep saves between versions. So maybe we;
- should not wipe the system on init if it is found to be incompatible
- should add a mechanism to read the first sector, e.g with a custom path and special handling maybe? current `eepfs_get_file` does not allow reading it at all.

src/eepromfs.c
rasky: I saw that in `dragonfs.c` and I think it's wrong there as well. I see zero reasons why the pointer `buf` should be const; in fact, I see zero reasons why a function argument should ever declared const, unless you're coming from a different language where you're used to program the type system more and you're trying to borrow the same concepts to C. In general, I think we can agree that in idiomatic C function arguments are rarely/never defined as const, so I think `dragonfs` should be changed as well.

meeq:
* Removed signature verification from `eepfs_init`; it is now the responsibility of the homebrew developer to call `eepfs_verify_signature` and `eepfs_wipe`. Example has been updated accordingly.

* divide_ceil macro has been made safe(r)

* Added documentation about slow EEPROM writes. From my research, reading EEPROM does not incur the 15ms wait, just writing.

I believe this is in a good-enough state to ship. I know there was some discussion around buffering reads/writes, but I don't think that is necessary to get this merged. It shouldn't impact the API, so it can be addressed later.

src/eepromfs.c
meeq: It is my opinion that `const` is a useful tool for signaling intent, and there is merit in communicating that an API will not be modifying certain data (specifically pointers).

However, I agree that `void * const` for the `src`/`dest` parameters is worthless, communicates nothing, and necessitates more code just to work around it.

This has been fixed in https://github.com/DragonMinded/libdragon/pull/125/commits/8a71b66c8626ab99272b1a4a3e679bd9db54ceee

rasky:
To clarify, I was referring to const variables or const pointers, not
pointers to const data. The latter is obviously useful.

rasky (approved):
Good to go for me!

meeq:
I want to add proper tests to this before merging. I encountered some issues I haven't fully debugged while integrating this branch back into a larger project that heavily uses fread/fwrite

meeq:
@rasky @anacierdem I have removed the stdio hooks and added some tests.

Please let me know if you want to see any other changes. Thanks!

tests/test_eepromfs.c
```diff
@@ -0,0 +1,93 @@
...
+        SKIP("EEPROM not found; skipping eepfs tests");
+    }
+    if (IN_EMULATOR) {
+        // Gracefully handle broken default EEPROM in cen64
+        uint64_t eeprom_block = 0;
+        uint8_t * eeprom_bytes = (uint8_t *)&eeprom_block;
+        eeprom_write(0, eeprom_bytes);
+        eeprom_read(0, eeprom_bytes);
+        if (eeprom_block != 0) {
+            SKIP("cen64 defaults to broken 4K EEPROM implementation");
```
meeq: https://github.com/n64dev/cen64/blob/a56fa4ba41da2c15cb054c063082988fb66780c0/si/controller.c#L124-L132

sp1187: A better option that could be investigated would be to make a PR for CEN64 that fixes this issue there.

meeq: I completely agree that cen64's behavior is incorrect, but the very existence of the hack means that someone found it to be necessary under some circumstance (the details of which are unfortunately missing).

I implemented this workaround in the spirit of [Chesterton's Fence](https://www.chesterton.org/taking-a-fence-down/):
>In the matter of reforming things, as distinct from deforming them, there is one plain and simple principle; a principle which will probably be called a paradox. There exists in such a case a certain institution or law; let us say, for the sake of simplicity, a fence or gate erected across a road. The more modern type of reformer goes gaily up to it and says, “I don’t see the use of this; let us clear it away.” To which the more intelligent type of reformer will do well to answer: “If you don’t see the use of it, I certainly won’t let you clear it away. Go away and think. Then, when you can come back and tell me that you do see the use of it, I may allow you to destroy it.

Since I do not know precisely why this hack was implemented in cen64, I was afraid to touch it in case my attempt to "fix it" actually ended up breaking something else in unexpected ways.

rasky: I do not maintain cen64, but in general for emulators that focus on accuracy its better to fix bugs even if that means that temporarily games get broken -- it would prompt people to find and do the correct fix.

meeq: It looks like mupen64plus has the same hack; it defaults to a 4K EEPROM in the Joybus Info command response, even if the cart doesn't actually have an EEPROM.

https://github.com/mupen64plus/mupen64plus-core/blob/db56853707ca6f4223adf5e7f0b7725f90d3a0a3/src/main/main.c#L1703-L1724

https://github.com/mupen64plus/mupen64plus-core/blob/db56853707ca6f4223adf5e7f0b7725f90d3a0a3/src/device/cart/cart.c#L45-L52

I tried to find documentation on how this is "supposed" to work, but I can't find a definitive answer.

https://n64brew.dev/wiki/Joybus_Protocol#0x00_-_Info does not indicate that there is an identifier for "No EEPROM present".

Libdragon assumes that no EEPROM is present if the identifier is not 4K EEPROM (`0x0080`) and is not 16K EEPROM (`0x00C0`), which seems reasonable.

I will build a test ROM that logs out the Joybus response on real hardware to get to the bottom of this.

meeq: A real console responds with:

Save Type | Rx Bytes
--- | ---
EEPROM 4K | `0x008000`
EEPROM 16K | `0x00C000`
All others | `0xFFFFFF`

I'll write up some PRs for Mupen64Plus and cen64 to respond accurately to EEPROM status inquiries.

Thanks for your feedback @rasky  and @sp1187

meeq: https://github.com/n64dev/cen64/pull/203

meeq: https://github.com/mupen64plus/mupen64plus-core/pull/878

meeq: Cen64's behavior when no EEPROM is specified is straight-up broken. Mupen64Plus just always has a 4K EEPROM unless 16K EEPROM is specified, which is not accurate but is not exactly harmful either.

src/controller.c
```diff
@@ -133,17 +133,36 @@ static void __controller_exec_PIF( const void *inblock, void *outblock )
...
+
+    assert(in_size <= sizeof(input));
+    assert(out_size <= sizeof(output));
+
+    memcpy( input, data_in, in_size );
+
+    __controller_exec_PIF( input, output );
+
+    memcpy( data_out, output, out_size );
+}
```
rasky: Why do you need an internal `input` / `output` here? If it's just because in_size could be less than 64 bytes, that is already taken care of by `__controller_exec_PIF`, isn't it?

IOW how's this function different from `__controller_exec_PIF`?

src/eepromfs.c
```diff
@@ -0,0 +1,523 @@
...
+ * @brief Reads an entire file from the EEPROM filesystem.
+ *
+ * @param[in]  path
+ *             Path of file in EEPROM filesystem to read from
+ * @param[out] dest
+ *             Buffer to read into
+ *
+ * @return EEPFS_ESUCCESS on success or a negative error otherwise
+ */
+int eepfs_read(const char * path, void * dest)
```
rasky: I was wondering if `eepfs_read` (and `eepfs_write`) should receive a size argument, and the function should return error if the size is different from that expected by the filesystem.

It's obviously a redundant information but it's easy to provide by the caller (once you pass the buffer, you must have it allocated somehow of the correct size, so the size must be known at the point of the call) and can prevent stupid bugs and memory overruns which are always hard to debug on an embedded system.

tests/test_eeprom.c
```diff
@@ -0,0 +1,86 @@
...
+    } else if (pif_recv_buf[1] == 0x80) {
+        LOG( "4K EEPROM detected.\n" );
+        ASSERT_EQUAL_HEX(pif_out[1], (uint64_t)0x008000fe00000000, "pif response mismatch");
+        ASSERT(eeprom_type == EEPROM_4K, "eeprom type changed?");
+    } else {
+        LOG( "EEPROM not detected.\n" );
+        ASSERT_EQUAL_HEX(pif_out[1], (uint64_t)0x000000fe00000000, "pif response mismatch");
+        ASSERT(eeprom_type == EEPROM_NONE, "eeprom type changed?");
+    }
+}
```
rasky: I'm not sure how I feel about this test. This sounds more like a test to help emulator authors. Libdragon tests should be tests of libdragon functionality to make sure it keeps working over time. For instance, I would expect here a test of `eeprom_read_bytes` / `eeprom_write_bytes`. This wouldn't required to export `pif_execute` BTW.

tests/test_eepromfs.c
```diff
@@ -0,0 +1,83 @@
...
+    ASSERT_EQUAL_SIGNED(result, 0, "eepfs erase/read mismatch");
+
+    // Test erasing file2
+    result = eepfs_erase("file2");
+    ASSERT_EQUAL_SIGNED(result, EEPFS_ESUCCESS, "eepfs erase failed");
+    memset(file2_src, 0, sizeof(file2_src));
+    result = eepfs_read("file2", file2_dst);
+    ASSERT_EQUAL_SIGNED(result, EEPFS_ESUCCESS, "eepfs read failed");
+    result = memcmp(file2_src, file2_dst, sizeof(file2_src));
+    ASSERT_EQUAL_SIGNED(result, 0, "eepfs erase/read mismatch");
```
rasky, line 92: Maybe at the end you can initialize a different filesystem and check that the signature fails? So that we cover also that codepath.

rasky (commented):
Done another round, I think we're very close now!

src/controller.c
meeq: `pif_execute` exists for a few reasons:
1. `__controller_exec_PIF` is static and not exposed outside the translation unit.
2. `pif_execute` allows input less than 64-bytes and sets the last bit to `1` (which is required)
3. `execute_raw_command` is targeted specifically at controllers and does not support commands on channel 4

`pif_execute` really only services the EEPROM test, which (as you point out elsewhere) is more of a test for emulators than it is for libdragon.

I'll remove `pif_execute` and the EEPROM test from this PR. They were super helpful for confirming real hardware behavior and testing various emulators, but you're right that they're not really appropriate to be included in libdragon's tests.

tests/test_eeprom.c
meeq: See https://github.com/DragonMinded/libdragon/pull/125#discussion_r658879871

src/eepromfs.c
meeq: This is a good idea. I'll incorporate this.

tests/test_eepromfs.c
meeq: https://github.com/DragonMinded/libdragon/pull/125/commits/fd04a6359d93cd5f1d34906a97b287301ab69b55

tests/test_eeprom.c
meeq: `eeprom_read_bytes` and `eeprom_write_bytes` get exercised by the EEPROM filesystem tests, so I didn't see a need to include them here

src/eepromfs.c
meeq: https://github.com/DragonMinded/libdragon/pull/125/commits/fd04a6359d93cd5f1d34906a97b287301ab69b55

src/controller.c
meeq: https://github.com/DragonMinded/libdragon/pull/125/commits/fd04a6359d93cd5f1d34906a97b287301ab69b55

rasky:
Approved. Just missing doc update to eeprom_read_bytes and eeprom_write_bytes.

PR #126, opened 2021-02-28
Default exception handler

anacierdem:
~~Fix console new line bug~~

~~- Previously it was adding an extra space when the current line is exactly CONSOLE_WIDTH~~

Interrupt handler improvements

    - Remove some unnecessary nops
    - No need to disable interrupts while in the interrupt handler
    - No need to save/restore k0 & k1
    - Do not restore fcr31 register, it should be always cleared by the ex handler.
    - Store FCR31 as a 32bit value

Add cop0 & cop1 macros and start using them

Remove inline wait functions from the header

Implement and register a default exception handler

I'll be implementing a test for the interrupt handler next.

### DIFF
--- a/include/console.h
+++ b/include/console.h
@@ -40,10 +40,18 @@
  * @{
  */
 
-/** @brief The console width in characters */
-#define CONSOLE_WIDTH       74
-/** @brief The console height in characters */
-#define CONSOLE_HEIGHT      30
+/**
+ * @brief The console width in characters
+ * @note Right padding is not enforced by code, adjust this to fit into
+ * [horizontal resolution] - 2 * HORIZONTAL_PADDING
+ */
+#define CONSOLE_WIDTH       64
+/**
+ * @brief The console height in characters
+ * @note Bottom padding is not enforced by code, adjust this to fit into
+ * [vertical resolution] - 2 * VERTICAL_PADDING
+ */
+#define CONSOLE_HEIGHT      28
 /** @} */
 
 /** 
@@ -51,7 +59,19 @@
  *
  * @note This needs to divide evenly into #CONSOLE_WIDTH 
  */
-#define TAB_WIDTH           5
+#define TAB_WIDTH           4
+
+/**
+ * @brief Padding from the left and right ends of the screen in pixels
+ *
+ * @note This should be ~10% of the horizontal resolution to be safely visible
+ */
+#define HORIZONTAL_PADDING  64
+
+/**
+ * @brief Padding from the top and bottom ends of the screen in pixels
+ */
+#define VERTICAL_PADDING    8
 
 #ifdef __cplusplus
 extern "C" {
--- a/include/console.h
+++ b/include/console.h
@@ -41,9 +41,9 @@
  */
 
 /** @brief The console width in characters */
-#define CONSOLE_WIDTH       35
+#define CONSOLE_WIDTH       74
 /** @brief The console height in characters */
-#define CONSOLE_HEIGHT      25
+#define CONSOLE_HEIGHT      30
 /** @} */
 
 /** 
--- a/include/cop0.h
+++ b/include/cop0.h
@@ -89,12 +89,12 @@
 /* COP0 interrupt bits definition. These are compatible bothwith mask and pending bits. */
 #define C0_INTERRUPT_0      0x00000100
 #define C0_INTERRUPT_1      0x00000200
-#define C0_INTERRUPT_2      0x00000400 // RCP
+#define C0_INTERRUPT_RCP    0x00000400 // RCP
 #define C0_INTERRUPT_3      0x00000800
 #define C0_INTERRUPT_4      0x00001000
 #define C0_INTERRUPT_5      0x00002000
 #define C0_INTERRUPT_6      0x00004000
-#define C0_INTERRUPT_7      0x00008000 // Timer
+#define C0_INTERRUPT_TIMER  0x00008000 // Timer
 
 /**
  * @brief Get the CE value from the COP0 status register
--- a/include/cop0.h
+++ b/include/cop0.h
@@ -51,6 +51,30 @@
 })
 
 /**
+ * @brief Returns the COP0 register $13 (Cause Register)
+ *
+ * The coprocessor 0 (system control coprocessor - COP0) register $13 is a read
+ * write register keeping pending interrupts, exception code, coprocessor unit
+ * number referenced for a coprocessor unusable exception.
+ *
+ * @see #C0_GET_CAUSE_EXC_CODE, #C0_GET_CAUSE_CE and #C0_CAUSE_BD
+ */
+#define C0_READ_CR() ({ \
+	uint32_t x; \
+	asm volatile("mfc0 %0,$13" : "=r" (x) : ); \
+	x; \
+})
+
+/**
+ * @brief Write the COP0 register $13 (Cause register)
+ * 
+ * Use this to update it for a custom exception handler.
+ * */
+#define C0_WRITE_CR(x) ({ \
+    asm volatile("mtc0 %0,$13"::"r"(x)); \
+})
+
+/**
  * @brief Returns the COP0 register $8 (BadVAddr)
  *
  * The coprocessor 0 (system control coprocessor - COP0) register $8 is a read
--- a/include/cop0.h
+++ b/include/cop0.h
@@ -73,8 +73,16 @@
  * needs correction in the exception handler. This macro is for writing its
  * value so that execution continues there.
  */
-#define C0_WRITE_EPC(dest) ({ asm volatile("mtc0 %0,$14"::"r"(dest)); })
+#define C0_WRITE_EPC(value) ({ asm volatile("mtc0 %0,$14"::"r"(value)); })
 
+/**
+ * @brief Read the COP0 register $14 (EPC)
+ */
+#define C0_READ_EPC() ({ \
+	uint32_t x; \
+	asm volatile("mfc0 %0,$14" : "=r" (x) : ); \
+	x; \
+})
 
 /* COP0 Status bits definition. Please refer to MIPS R4300 manual. */
 #define C0_STATUS_IE        0x00000001
--- a/include/cop0.h
+++ b/include/cop0.h
@@ -50,19 +50,49 @@
     asm volatile("mtc0 %0,$12"::"r"(x)); \
 })
 
+/**
+ * @brief Returns the COP0 register $8 (BadVAddr)
+ *
+ * The coprocessor 0 (system control coprocessor - COP0) register $8 is a read
+ * only register holding the last virtual address to be translated which became
+ * invalid, or a virtual address for which an addressing error occurred.
+ */
+#define C0_READ_BADVADDR() ({ \
+	uint32_t x; \
+	asm volatile("mfc0 %0,$8" : "=r" (x) : ); \
+	x; \
+})
+
+/**
+ * @brief Writes to the COP0 register $14 (EPC)
+ *
+ * The coprocessor 0 (system control coprocessor - COP0) register $14 is the
+ * return from exception program counter. For asynchronous exceptions it points
+ * to the place to continue execution whereas for synchronous (caused by code)
+ * exceptions, point to the instruction causing the fault condition, which
+ * needs correction in the exception handler. This macro is for writing its
+ * value so that execution continues there.
+ */
+#define C0_WRITE_EPC(dest) ({ asm volatile("mtc0 %0,$14"::"r"(dest)); })
+
+
 /* COP0 Status bits definition. Please refer to MIPS R4300 manual. */
 #define C0_STATUS_IE        0x00000001
 #define C0_STATUS_EXL       0x00000002
 #define C0_STATUS_ERL       0x00000004
 
-#define C0_STATUS_IM0       0x00000100
-#define C0_STATUS_IM1       0x00000200
-#define C0_STATUS_IM2       0x00000400
-#define C0_STATUS_IM3       0x00000800
-#define C0_STATUS_IM4       0x00001000
-#define C0_STATUS_IM5       0x00002000
-#define C0_STATUS_IM6       0x00004000
-#define C0_STATUS_IM7       0x00008000
+/* COP0 Cause bits definition. Please refer to MIPS R4300 manual. */
+#define C0_CAUSE_BD         0x80000000
+
+/* COP0 interrupt bits definition. These are compatible bothwith mask and pending bits. */
+#define C0_INTERRUPT_0      0x00000100
+#define C0_INTERRUPT_1      0x00000200
+#define C0_INTERRUPT_2      0x00000400 // RCP
+#define C0_INTERRUPT_3      0x00000800
+#define C0_INTERRUPT_4      0x00001000
+#define C0_INTERRUPT_5      0x00002000
+#define C0_INTERRUPT_6      0x00004000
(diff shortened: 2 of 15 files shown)


include/cop0.h
```diff
@@ -50,19 +50,49 @@
...
+
+/* COP0 interrupt bits definition. These are compatible bothwith mask and pending bits. */
+#define C0_INTERRUPT_0      0x00000100
+#define C0_INTERRUPT_1      0x00000200
+#define C0_INTERRUPT_2      0x00000400 // RCP
+#define C0_INTERRUPT_3      0x00000800
+#define C0_INTERRUPT_4      0x00001000
+#define C0_INTERRUPT_5      0x00002000
+#define C0_INTERRUPT_6      0x00004000
+#define C0_INTERRUPT_7      0x00008000 // Timer
```
rasky: Maybe we should call them `C0_INTERRUPT_RCP` and `C0_INTERRUPT_TIMER`.

src/exception.c
```diff
@@ -36,6 +40,145 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+		console_render();
+	}
+}
+
+void exception_default_handler(exception_t* ex) {
+	// We will naturally return from the ex handler, which
+	// will gracefully restore interrupt state so that we can
+	// continue using interrupt dependent functionalities
+	// such as console updates.
+	C0_WRITE_EPC(&exception_halt);
```
rasky: Did you try using `&abort` here? So that we don't need a specific function. But if we need to actually call `console_render()`, then a function is required.

src/inthandler.S
```diff
@@ -17,17 +17,13 @@ inthandler:
...
 
 	la k1,save01
 	sd $1,(k1)
 
 	mfc0 k1,C0_SR
 	la $1,saveSR
 	sw k1,($1)
-	la $1, ~1
-	and k1,$1
-	mtc0 k1,C0_SR
```
rasky: I think it made sense to turn off the interrupt bit here. Doesn't it?

src/inthandler.S
```diff
@@ -98,8 +92,7 @@ inthandler:
 	sdc1 $f23,saveFR23
 	sdc1 $f24,saveFR24
 	sdc1 $f25,saveFR25
-	sdc1 $f26,saveFR26
-	sdc1 $f27,saveFR27
```
rasky: `$f26` and `$f27` I guess we need to save?

src/inthandler.S
```diff
@@ -229,8 +221,6 @@ endint:
 	ld $30,saveFC31
 	nop
 
-	ctc1 $30,$f31
-
```
rasky, line 233: Uhm not sure about this. I think that, in general, we need restore `$fc31` in case the interrupt code modifies it. For instance, it contains floating point flags. I see that for the specific case of FPU exception, you need to mask some flags but maybe that should be handled specially, while still allowing the other bits to be restored.

src/exception.c
```diff
@@ -36,6 +40,145 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+			if (fcr31 & C1_CAUSE_INEXACT_OP) fprintf(stdout, "Inexact Op.\n");
+			if (fcr31 & C1_CAUSE_UNDERFLOW) fprintf(stdout, "Underflow\n");
+			if (fcr31 & C1_CAUSE_OVERFLOW) fprintf(stdout, "Overflow\n");
+			if (fcr31 & C1_CAUSE_DIV_BY_0) fprintf(stdout, "Div-by-0\n");
+			if (fcr31 & C1_CAUSE_INVALID_OP) fprintf(stdout, "Invalid Op.\n");
+			if (fcr31 & C1_CAUSE_NOT_IMPLEMENTED) fprintf(stdout, "Not impl.\n");
+
+			fprintf(stdout, "Floating-point registers\n");
+			for (int i = 0; i<32; i++) {
+				fprintf(stdout, "%02u:%08llX ", i, ex->regs->fpr[i]);
```
anacierdem: We need 32 bit values here if we dont switch to a higher res.

src/exception.c
```diff
@@ -36,6 +40,145 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+		sr & C0_INTERRUPT_5 ? 1 : 0,
+		sr & C0_INTERRUPT_6 ? 1 : 0,
+		sr & C0_INTERRUPT_7 ? 1 : 0);
+
+	switch(ex_code) {
+		case EXCEPTION_CODE_STORE_ADDRESS_ERROR:
+			fprintf(stdout, "BadVAddr: %08lX\n", C0_READ_BADVADDR());
+		break;
+		case EXCEPTION_CODE_COPROCESSOR_UNUSABLE:
+			fprintf(stdout, "COP:%1lu\n", (cr >> 28) & 3);
```
anacierdem: Use macros here as well

src/exception.c
```diff
@@ -36,6 +40,145 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+		sr & C0_INTERRUPT_1 ? 1 : 0,
+		sr & C0_INTERRUPT_2 ? 1 : 0,
+		sr & C0_INTERRUPT_3 ? 1 : 0,
+		sr & C0_INTERRUPT_4 ? 1 : 0,
+		sr & C0_INTERRUPT_5 ? 1 : 0,
+		sr & C0_INTERRUPT_6 ? 1 : 0,
+		sr & C0_INTERRUPT_7 ? 1 : 0);
+
+	switch(ex_code) {
+		case EXCEPTION_CODE_STORE_ADDRESS_ERROR:
```
anacierdem: Also combine with other addressing errors

anacierdem (commented):
Self review

src/exception.c
anacierdem: Which wont make any sense for doubles.

src/inthandler.S
anacierdem: Oops..

src/inthandler.S
anacierdem: Maybe we should set bits except the ones that needs to be reset by the handler. That would be better youre right.

src/exception.c
anacierdem: Ill try that but I suspect it wont work

src/inthandler.S
anacierdem: Why do you think so? EXL is already set at this point and I dont think an explicit disable is necessary.

src/exception.c
anacierdem: Unfortunately it does not work. I tried to comment on why it does not work without the interrupts. It may not be 100% accurate and I think it is not worth pursuing further just to remove the extra function. We may have a look again when we update he display system someday :)

include/console.h
```diff
@@ -41,9 +41,9 @@
  */
 
 /** @brief The console width in characters */
-#define CONSOLE_WIDTH       35
+#define CONSOLE_WIDTH       74
 /** @brief The console height in characters */
-#define CONSOLE_HEIGHT      25
+#define CONSOLE_HEIGHT      30
```
rasky: Self note: I want to check this on my CRT. I think borders in general help because of the typical overscan of CRTs. I'm not sure we should remove them

src/console.c
```diff
@@ -268,7 +268,7 @@ static void __console_render()
 
             /* Draw to the screen using the forecolor and backcolor set in the graphics
              * subsystem */
-            graphics_draw_character( dc, 20 + 8 * x, 16 + 8 * y, t_buf );
+            graphics_draw_character( dc, 24 + 8 * x, 8 * y, t_buf );
```
rasky: Maybe expose the margins as defines as well, so that we can reason upon them better?

src/exception.c
```diff
@@ -36,6 +42,159 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+			C1_WRITE_FCR31(
+				C1_FCR31() & ~(
+					C1_CAUSE_INEXACT_OP |
+					C1_CAUSE_UNDERFLOW |
+					C1_CAUSE_OVERFLOW |
+					C1_CAUSE_DIV_BY_0 |
+					C1_CAUSE_INVALID_OP |
+					C1_CAUSE_NOT_IMPLEMENTED
+				)
+			);
```
rasky, line 103: So to clarify: this is only necessary because we need the hack of returning from the exception to call console_render, right? In principle if we fix that, we could remove these lines, right? If so, I'll add a comment for later when we get around and fix the console.

src/exception.c
```diff
@@ -36,6 +42,159 @@ void register_exception_handler( void (*cb)(exception_t*))
...
+	console_set_render_mode(RENDER_MANUAL);
+	console_clear();
+
+	fprintf(stdout, "**************************************************************************");
+	fprintf(stdout, "%s exception at PC:%08lX\n", ex->info, ex->regs->epc);
+
+	fprintf(stdout, "cr:%08lX (COP:%1lu BD:%lu)\n", cr,  C0_GET_CAUSE_CE(cr), cr & C0_CAUSE_BD);
+	fprintf(stdout, "sr:%08lX FCR31:%08X BVAdr:%08lX \n", sr, (unsigned int)fcr31, C0_READ_BADVADDR());
+	fprintf(stdout, "--------------------------------------------------------------------------");
+	fprintf(stdout, "FPU     IOP UND OVE DV0 INV  NI  | INTs    sw0 sw1 ex0 ex1 ex2 ex3 ex4  tm");
```
rasky: Maybe the last string can be "tmr"

src/interrupt.c
```diff
@@ -594,9 +594,9 @@ void init_interrupts()
         __interrupt_depth = 0;
 
         /* Enable interrupts systemwide. We set the global interrupt enable,
-           and then specifically enable RCP interrupts (IM2). */
+           and then specifically enable RCP interrupts (INTERRUPT_2). */
```
rasky: INTERRUPT_2 in the command doesn't refer anymore to a macro, right? Maybe just remove it as the macro now has RCP in the name

src/inthandler.S
rasky: Ok I think you're right, it's probably not required.

src/inthandler.S
```diff
@@ -69,8 +64,7 @@ inthandler:
 	mfhi $30
 	sd $30,saveHI
 	cfc1 $30,$f31
-	nop
-	sd $30,saveFC31
+	sw $30,saveFC31
```
rasky, line 70: I assume you have tested this specific NOP removal on real hardware right?

src/inthandler.S
```diff
@@ -226,9 +219,18 @@ endint:
...
+	# Override the cause bits (12-17) of stored value from cop1.
+	# If they were cleared by the ex. handler, they will remain cleared.
+	# We will restore the other values to their state before the exception
+	# has occured. If we restore all the bits, the exception handler will
+	# be unable to clear them if needed. This is important as they will
+	# re-trigger the exception upon handler exit if remains set.
+	cfc1 k0, $f31
+	and k0, 0x3F000
+	or k0, ~0x3F000
+	and $30, k0
```
rasky: OK this makes sense if we keep the current approach where the exception handler registered by the application has the task to also clear the exception (contrary to interrupts, where the interrupt is acknowledge by libdragon itself before calling the handler).

So I would maybe document at the top `exception.c` and in the comment of `register_exception_handler` that the handler must clear the exception in case it's required to avoid re-triggering.

rasky (approved):
Great work! I left some minor comment. 

The only thing that really concerns me is using the full resolution on a CRT for the console by default. I would be less concerned if you disabled the margin just for the exception handler, but I would expect the console by default to be more usable with some margin.

include/console.h
anacierdem: Vertical padding does not seem to be an issue on the highest res mode. We even increased the horizontal padding as it was not enough for 640x240. Please test on your screen and let's see. We are already on the limits of available screen space so I hope it looks good :)

src/exception.c
anacierdem: Exactly, we can remove them from the default handler if we fix that.

src/interrupt.c
anacierdem: HMm I missed this one apparently

src/inthandler.S
anacierdem: True, it is still working on real hardware

anacierdem:
Unfortunately current console subsytem is not easy to extend. Will need a re-write to make it more customizable. I'm not a fan of increasing the default resolution. Maybe we should do it with explicit text render calls.

src/inthandler.S
anacierdem: I think it is better to leave this to the consumer indeed. Floating point exceptions are disabled by default and user may never decide to enable them.

I initially had an alternative implementation where I removed the const specifier from the exception struct so that the ex handler can change the contents to be loaded back. I started to believe that is more useful now.

If we do that, we can remove this. We will only give write access to epc and fcr31 (and maybe sr and cr) so that the custom handler won't have to use inline assembly and at the same time it means we completely inverted the control to the consumer.

What do you think @rasky?

anacierdem:
I have also strting adding tests. Lets keep them in a single PR. It will take some time but will be worth it IMO.

src/inthandler.S
rasky: I think that in general exception handlers might need to modify registers. It’s a pretty advanced use case, but why not. So maybe I’d simply drop const everywhere and document that it can be done.

anacierdem:
Added tests & improved console paddings. The only remaining issue is the increased timer benchmark time, for which I couldnt find any lead yet.

src/exception.c
```diff
@@ -21,13 +27,26 @@
...
 extern const void* __baseRegAddr;
 
 /**
  * @brief Register an exception handler to handle exceptions
  *
+ * The registered handle is responsible for clearing any bits that may cause
+ * a re-trigger of the same exception and updating the EPC. To manipulate the
+ * registers, update the values in the exception_t struct. They will be
+ * restored to appropriate locations when returning from the handler. Setting
+ * them directly will not work as expected as thhey will get overwritten.
```
rasky:
```suggestion
 * them directly will not work as expected as they will get overwritten.
```

src/exception.c
```diff
@@ -21,13 +27,26 @@
...
+ * An important example is the cause bits (12-17) of FCR31 from cop1. To
+ * prevent re-triggering the exception they should be cleared by the handler.
+ * k0 ($26), k1 ($27), s0-s7 ($16-$23), and gp ($28) are not saved/restored
+ * and will not be available in the handler. These are removed to gain some
+ * performance as they are already saved by GCC when necessary. Theoretically
```
rasky: I think that for *interrupts* it makes sense not to save those registers as they are callee-saved (BTW, the same I think it applies to f15-f31 but should be verified). Instead for exceptions, one might want to have a look at the whole state of all regs. So maybe the interrupt handlers should save/restore those only after it has verified that it's an exception.

Anyway, this can be improved in the future, no need to do it now.

rasky (approved):
Great work!

src/exception.c
anacierdem: https://github.com/DragonMinded/libdragon/issues/129

anacierdem:
That benchmark works fine with this state. I didn't find the cause or applied a solution though. So we better keep an eye on that. I'm mostly happy with this PR. Can merge when @rasky approves :)

rasky:
Fine by me, I tested it everywhere and it works. I previously told you that I had an unexpected FPU exception but it turned out that was really a bug in my code so this PR actually helped me debugging a real issue.

PR #130, opened 2021-03-16
New debugging library

rasky:
This PR contributes a new debugging library, to ease development. The new library provides:

 * Logging support. In general, on N64, logging is hooked up to `stderr`, which can be done with the provided `debugf` macro. There are several different output that can be activated even simultaneously:
   * USB logging with development cartridges, and compatible PC-side tools (like UNFLoader). On N64, this is hooked up to stderr.
   * Logging on a file in a SD card inserted in a development cartridge
   * Logging in supported emulators
 * Read/write filesystems. This can be useful to access external data and write dump files.
   * Currently, this is supported on SD cards inserted in development cartridges.

Moreover, even without explicitly activating the new debug library, `assert()` is now hooked up to the console to actually show the assertion to the user, rather than silently aborting. If instead the debugging library is enabled, the assertion is also logged to stderr. A new function `assertf()` is also added to simplify assertion with additional context (why oh why this is not available in standard C...).

To hook up `assert()` to the console, we need to recompile newlib with `-DHAVE_ASSERT_FUNC` so that newlib stops declaring `__assert_func` and let us declare it (in our code, it's in `debug.c`). That's what the build-system part of this PR does, in addition to upgrading all the toolchain to the latest version since we're at it. This part of this PR was prepared by @anacierdem.

NOTE: the PR is rather large because it also imports two third-party libraries (FatFS for FAT access, and a USB library from UNFLoader). It might be easier to review the PR commit by commit.


rasky:
Finally, this is ready for review! @DragonMinded @anacierdem

src/fatfs/ffconf.h
```diff
@@ -117,7 +117,7 @@
 /  ff_memfree() exemplified in ffsystem.c, need to be added to the project. */
 
 
-#define FF_LFN_UNICODE	0
+#define FF_LFN_UNICODE	2
```
anacierdem, line 120: Do we still need to set `FF_CODE_PAGE` when using UTF-8? Does it then still apply to short filenames?

src/debug.c
```diff
@@ -0,0 +1,603 @@
...
+			data_cache_hit_writeback_invalidate(buff, 512);
+			dma_read(buff, D64_CIBASE_ADDRESS + D64_BUFFER, 512);
+		}
+		else
+		{
+			typedef uint32_t u_uint32_t __attribute__((aligned(1)));
+
+			uint32_t* src = (uint32_t*)(D64_CIBASE_ADDRESS + D64_BUFFER);
+			u_uint32_t* dst = (u_uint32_t*)buff;
+			for (int i = 0; i < 512/16; i++)
```
anacierdem, line 237: Would a partially aligned dma be always slower than this? What is this 16 byte optimization for? Is it to reduce the jumps? What about completely unrolling this in place then?
Also why not do a memcopy?

src/debug.c
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	return RES_OK;
+}
+
+static DRESULT fat_disk_ioctl_64drive(BYTE cmd, void* buff)
+{
+	switch (cmd)
+	{
+	case CTRL_SYNC: return RES_OK;
```
anacierdem, line 287: Hmm doing `usb_64drive_wait()` here might be more smeantically correct although it would not change the actual behavior.
Probably not needed as documented;
> Nothing to do for this command if each write operation to the media is completed within the disk_write function.

src/debug.c
```diff
@@ -0,0 +1,603 @@
...
+	for (int i=0; i<sizeof(debug_writer) / sizeof(debug_writer[0]); i++)
+		if (debug_writer[i])
+			debug_writer[i]((uint8_t*)buf, len);
+
+	// Pretend stderr is written correctly even if it isn't. 
+	// There's really no benefit in bubbling up I/O errors
+	// for the log spew. Also there could be multiple channels
+	// and we wouldn't know how to report errors only on
+	// some of them.
+	return len;
```
anacierdem, line 477: Maybe we can fail if _none_ is written to get rid of some possible confusion. We may just assert here saying an err channel is present but we failed to write. Not worth the effort if the underlying functions are unable to report written chars though.
Also this maybe a pain point if some of the streams occasionally fail. On the other hand, it may cause some err logs to be missing and confuse the dev. further if we do not signal

src/console.c
```diff
@@ -188,6 +179,7 @@ void console_init()
     console_set_render_mode(RENDER_AUTOMATIC);
 
     /* Register ourselves with newlib */
+    stdio_t console_calls = { 0, __console_write, 0 };
     hook_stdio_calls( &console_calls );
```
anacierdem, line 191: Would be really nice if the hook call could manage the array of callbacks for us instead of `__stderr_write`. That would allow other consumers to update hooks without changing `__stderr_write`.
This is potentailly extendible later on by changing hook/unhook to add/remove the provided func to a linked list of pointers although it will have a little weird usage;
```
stdio_t console_calls1 = { 0, __console_write1, 0 };
hook_stdio_calls( &console_calls1 );

stdio_t console_calls2 = { 0, __console_write2, 0 };
hook_stdio_calls( &console_calls2 );
// We now have them both

unhook_stdio_calls( &console_calls1 ); // This will remove __console_write1 from the list
```

src/system.c
```diff
@@ -1328,4 +1334,20 @@ void _flush_cache(uint8_t* addr, unsigned long bytes) {
     inst_cache_hit_invalidate(addr, bytes);
 }
 
+/**
+ * @brief Implement underlying function for assert()
+ *
+ * Implementation of the function called when an assert fails. By default,
+ * we just abort execution, but this will be overriden at startup with
```
anacierdem, line 1341: If we are always overriding this at startup, why do we need to keep a pointer? Is it for in case we need to override it in the future?

tests/testrom.c
```diff
@@ -185,6 +194,7 @@ static const struct Testsuite
 	TEST_FUNC(test_irq_reentrancy,       230, TEST_FLAGS_RESET_COUNT),
 	TEST_FUNC(test_dfs_read,            1104, TEST_FLAGS_IO),
 	TEST_FUNC(test_cache_invalidate,    1763, TEST_FLAGS_NONE),
+	TEST_FUNC(test_debug_sdfs,             0, TEST_FLAGS_NO_BENCHMARK),
```
anacierdem, line 198: Why were we stressing it in the test if we are not measuring?

tests/test_debug.c
```diff
@@ -0,0 +1,74 @@
...
+	ASSERT(randf, "cannot open file: rom:/%s", ROM_FILE);
+	int sz = fread(random, 1, sizeof(random), randf);
+	ASSERT_EQUAL_UNSIGNED(sz, sizeof(random), "cannot read enough data");
+	fclose(randf); randf = NULL;
+
+	// Write file to SD.
+	randf = fopen(SD_FILE, "wb");
+	ASSERT(randf, "cannot create file: sd:/%s", SD_FILE);
+
+	// Write random-sized chunks of 1-25 bytes to stress the underlying
```
anacierdem, line 32: I'm not sure what this exactly testing, Are we trying to make sure fatfs is returning the correct value? Or maybe we are trying to make sure we have correctly registered it with newlib? This might not be that valuable of a test but, will not hurt.

src/usb.c
```diff
@@ -604,7 +604,7 @@ s8 usb_64drive_wait()
         #endif
         
         // Took too long, abort
-        if((timeout++) > 1000000)
+        if((timeout++) > 10000)
```
anacierdem, line 607: We probably need to PR these to the original source if we haven't already. Eve though this is a small lib, we may get out of sync very quickly

include/console.h
```diff
@@ -79,6 +80,7 @@ extern "C" {
 
 void console_init();
 void console_close();
+void console_debug(bool debug);
```
anacierdem: nit: console_set_debug maybe?

anacierdem (commented):
This is a really comprehensive PR. Loved it when reading it! Thank you. I have generally minor comments here and there. The only thing bothering me right now is keeping `usb.c` in sync with upstream. It is probably managable but considering the limited time we have working on libdragon, we may need a more systematic approach. fatfs is not an issue as it is (and will be) mostly used as is and does not seem to get updated often.

src/fatfs/ffconf.h
rasky: Yes it does apply to the way short filenames are written. Now, short filenames are mostly useless nowadays unless you're using a very old or very limited FAT implementation, but anyway it won't hurt to configure it one way or another. 

Another advantage of settings `FF_CODE_PAGE=437` is that the CP437 -> UTF8 conversion table is very small (see `uc437` in `ffunicode.c`), so it uses very little ROM space. The default was `FF_CODE_PAGE=932` which is Japanese Shift-JIS and that requires a far lager table.

src/debug.c
rasky: The ROM address space on N64 can only be accessed with aligned 32-bit reads/writes (or via DMA). So when DMA cannot be used because the target buffer is not aligned, then we need to do a manual copy. We cannot use `memcpy` because `memcpy` does not guarantee that all writes will be 32 bits, it expects to be able to access the memory with reads/writes of any width. Notice that the loop is also using an unaligned uin32_t type for the buffer in RDRAM so that the compiler will generate `LWL`/`SWL` for that as expected.

This whole loop is a fallback in case `fat_disk_read_64drive` is called with an unaligned buffer, otherwise DMA is used (which is much faster). I don't know if/when that happens, it is part of fatfs internals. Most of the times I have seen calls to read the whole `win` buffer in the `FATFS` structure, so it looks like fatfs tries to read full sectors as an internal optimization. `win` should be aligned -- now that I read this code again, I don't see an `attribute((aligned(8)))` added to it, maybe I've lost that as part of the PR. I will readd it.

src/debug.c
rasky: I think as a followup optimization, we can try to remove the very last `usb_64drive_wait()` from `fat_disk_write_64drive` and add it to `CTRL_SYNC`. Notice that this would parallelize only the last sector write in a group of write performed by `fat_disk_write_64drive`, so maybe it's not really worth it. To make something more elaborated, we would need a background thread to perform writes, but that sounds complicated, and also I'm not sure you can do that so easily -- the contents of `buff` might not be available anymore after `fat_disk_write_64drive` returns. 

At the end of the day, it's probably better to leave it synchronous like this, and then ask the user to move sdfs accesses in a different thread if they wish so. This would still require that `usb_64drive_wait` yields, which is complicated by the fact that we don't get an interrupt from 64drive or ED64 when the command is finished. Sorry, not many good news here :)

src/debug.c
rasky: I'm not opposed to this in principle, but 1) we don't have any backend that returns error codes, and 2) I don't foresee devs checking error codes from `debugf()` calls. 

So given that we don't have concrete cases to reason about (especially wrt how callers would handle errors), and given that it's a log spew, I think it won't matter much anyway.

src/console.c
rasky: I was concerned that this is a change in behavior to `hook_stdio_calls` (public API) compared to before. Do you think we can ignore this?

src/system.c
rasky: libdragon is compiled in two libraries `libdragonsys.a` and `libdragon.a`. `libdragonsys` contains the "OS backend" of newlib (`system.c` mostly), while `libdragon.a` contains everything else.

This is because historically C linkers have not been able to solve circular dependencies of symbols among libraries and required command line invocations to list libraries in strict dependency order. So the order we expect is `game` -> `libdragon` -> `newlib` -> `libdragonsys.a` (or even `game` -> `newlib` -> `libdragonsys.a`). This is why all examples have the following library list: `-ldragon -lc -lm -ldragonsys` in that order.

So we need to implement `__assert_func` which is called by newlib, so it must stay in `libdragonsys.a`. At the same time, we want the "real" assert code to be in debug.c and it must be able to call the console library etc., so it must stay in `libdragon`. To break this dependency, we add the function pointer indirection.

Notice that since we are using a modern linker, we could ask the linker to solve circular dependencies; for reasons that elude me (maybe performance), that's not on by default, so if we start requiring that, it would break all linker scripts in the world, with strange errors about missing symbols that nobody would know how to debug or fix. 

So at the end of the day, it's better to keep it like it is right now, and respecting the split between low-level system code and higher level functionalities.

tests/testrom.c
rasky: The test code is for correctness; we can't benchmark SD accesses because it will be dependent on the actual speed of the physical SD being inserted and also the status of the FAT partition in it.

tests/test_debug.c
rasky: The code is trying to stress `fat_disk_read_64drive` / `fat_disk_write_64drive` which have complicated logic between aligned and non-aligned reads/writes. So what I'm trying to test is the glue between FATFS and newlib. I can't be sure to cover everything because obviously it depends on fatfs internals and how aggressively it internally caches reads/writes, but I figure it out this is better than doing nothing.

src/usb.c
rasky: Done & merged already: https://github.com/buu342/N64-UNFLoader/pull/55

include/console.h
rasky: yeah good idea, I've changed it.

src/console.c
anacierdem: We are already changing it a little so should not be an issue IMO

src/system.c
anacierdem: I now realized why, thanks. Maybe link time optimizations can handle that and remove the additional call

rasky:
I've just added two asserts in core libdragon:

 * enable_interrupts() without disable_interrupts() (would cause all interrupts to stay enabled forever, causing random crashes very hard to debug)
 * dma_read() / dma_write() with len=0 (would cause a hard crash)

I'm going to double check the FatFS buffer alignment soon which is the last open point here.

anacierdem:
We have decided to merge this one as is for now. We will fix the remaining bugs later. Debugging library should be assumed alpha at this stage.

PR #131, opened 2021-06-05
Adding restart_timer and new_timer_stopped functions

stefanmielke:
These functions allowed me to add a new timer that was stopped and that could be restarted later as callbacks to actions.


anacierdem:
Instead of splitting the initializer into two, I'd suggest having a new flag (e.g `TF_ENABLED`) causing `__proc_timers` to skip processing it. With the current implementation `new_timer_stopped` will create a dangling timer if not explicitly handled. Also adding a new flag will make testing new functionality much easier.

anacierdem (changes requested):
Instead of splitting the initializer into two, I'd suggest having a new flag (e.g `TF_ENABLED`) causing `__proc_timers` to skip processing it. With the current implementation `new_timer_stopped` will create a dangling timer if not explicitly handled. Also adding a new flag will make testing new functionality much easier. Also @rasky is very accustomed to the implementation, and might have an opinion on this.

rasky:
Yeah I'd probably use a flag as well to set the timer as stopped, but I'd check that in `new_timer` and simply avoid registering the timer. `restart_timer` looks fine,

anacierdem:
What should it mean to restart a timer multiple times? Should we really register a few?

rasky:
Well it should probably be a nop but at the same time we have the same behavior in start_timer today. So I guess restart_timer is coherent with start_timer

stefanmielke:
Made the changes to new_timer_stopped. With this change we could also include a "pause/resume" functionality later.

anacierdem:
An additional test case would be really nice. It would also make the usage more visible.

stefanmielke:
Fixed a few problems and added a couple tests. Also made a few changes to tests that were having issues to make them pass correctly.

rasky:
Thanks, looks fine to me!

PR #133, opened 2021-06-07
Improve fastpath of dfs_read and add dfs_rom_addr

rasky:
dfs_read improvements:

In dfs_read, there's a fastpath that bypasses the internal file cache
and DMA the data directly into the output buffer. The existing code
assumed that this was possible only when both the RDRAM and ROM addresses
and the size were all multiple of 8.

This is wrong: PI is able to DMA from any even ROM address; also, the DMA
length must follow some weird rules (in general only even lengths, but
small lengths are supported). The constraint of 8-bytes aligned only
stands for RDRAM.

dfs_rom_addr:

For very advanced cases where you need to squeeze every bit of
performance, it might be needed to bypass the whole DFS layer and
physically access the ROM. One scenario is to run largish DMA transfers
in background (maybe triggered under interrupt).

Now that in Dragon FS 2.0 files are laid out linearly, this is indeed
possible. Add a function dfs_rom_addr that returns the initial ROM
address of a file (without opening it), so that it can be manually
accessed via io_read and dma_read.

This should not be required in most cases, but it's still useful to have.

### DIFF
--- a/src/dragonfs.c
+++ b/src/dragonfs.c
@@ -956,17 +956,26 @@ int dfs_read(void * const buf, int size, int count, uint32_t handle)
     if (!to_read)
         return 0;
 
-    /* If the destination pointer, ROM location and the amount of data to read
-       are all 8-bytes aligned, we can DMA directly into the destination buffer.
-       This is much faster than using cached data, so bypass also our local
-       cache as well. */
-    uint32_t align = ((uint32_t)buf | (file->cart_start_loc + file->loc) | to_read) & 0xF;
-    if ((align&7) == 0)
+    /* Fast-path. If possibly, we want to DMA directly into the destination
+     * buffer, without using any intermediate buffers. The rules are convoluted
+     * because we try to squeeze maximum performance here and thus we rely also
+     * on undocumented behaviors of PI DMA.
+     * The rules we follow are:
+     *
+     *   * The RDRAM destination pointer must be 8-bytes aligned.
+     *   * The ROM location must be 2-bytes aligned.
+     *   * The length must be either less than 0x7F (all values accepted),
+     *     or even.
+     */
+    bool rom_aligned = (file->loc & 1) == 0;
+    bool ram_aligned = ((uint32_t)buf & 7) == 0;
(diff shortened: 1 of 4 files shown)


rasky:
Rebased to avoid conflicts

src/dragonfs.c
```diff
@@ -1037,6 +1046,44 @@ int dfs_size(uint32_t handle)
...
+/**
+ * @brief Return the physical address of a file (in ROM space)
+ *
+ * This function should be used for highly-specialized, high-performance
+ * use cases. Using dfs_open / dfs_read is generally acceptable
+ * performance-wise, and is easier to use rather than managing
+ * direct access to PI space.
+ * 
+ * Direct access to ROM data must go through io_read or dma_read. Do not
+ * dereference directly as the console might hang if the PI
```
anacierdem: if the PI ?

tests/testrom.c
```diff
@@ -295,6 +298,7 @@ int main() {
 
 	int64_t total_time = TIMER_MICROS(stop-start) / 1000000;
 
-	printf("\nTestsuite finished in %02lld:%02lld\n", total_time%60, total_time/60);
+	console_set_debug(true);
```
anacierdem, line 301: IIRC this will not work on an ed64 without something listening on the USB. I'm not sure of the cause b/c I know it is not a technical requirement. Let's merge this and I will try to fix it later.

anacierdem (approved):

src/dragonfs.c
rasky: ... is busy :) will add this.

tests/testrom.c
rasky: It's a leftover of the debugging library PR. The ed64 backend in usb.c doesn't have a timeout. It might require a modification in `usb_everdrive_wait_pidma` or `usb_everdrive_usbbusy` or both, to add a simple counter-based timeout like in `usb_64drive_waitidle`. If you could do this it would be great.

rasky:
Pushed fix to comment.

PR #135, opened 2021-06-11
Implement EEPROM helpers to read/write buffers

meeq:
This is a stripped-down version of #125 that ended up being "the simplest thing that could possibly work" for the project I was working on.

What I really wanted was an easy way to read and write structs in/out of EEPROM. The filesystem approach introduced a lot of extra complexity and an abstraction layer that leaked pretty badly in pathological cases.

I'd like to take another swing at the filesystem approach because I think it's clever and useful, but I need to write better tests for it and solve for some edge-cases before I'd call it "shippable".

### DIFF
--- a/src/controller.c
+++ b/src/controller.c
@@ -133,15 +133,11 @@ static void __controller_exec_PIF( const void *inblock, void *outblock )
 }
 
 /**
- * @brief Probe the EEPROM interface
+ * @brief Probe the EEPROM interface on the cartridge.
  *
- * Probe the EEPROM to see if it exists on this cartridge.
- *
- * @retval #EEPROM_16k The cartridge has a 16 kilobit EEPROM
- * @retval #EEPROM_4k The cartridge has a 4 kilobit EEPROM
- * @retval #EEPROM_NONE The cartridge does not have an EEPROM
+ * @return Which EEPROM type was detected on the cartridge.
  */
-int eeprom_present()
+eeprom_type eeprom_present()
 {
     static const unsigned long long SI_eeprom_status_block[8] =
     {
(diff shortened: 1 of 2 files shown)


src/controller.c
```diff
@@ -227,11 +230,109 @@ void eeprom_write(int block, const uint8_t * const data)
...
+		bytes_left -= (EEPROM_BLOCK_SIZE - block_offset);
+		while (block_offset < EEPROM_BLOCK_SIZE)
+		{
+			*dest++ = buf[block_offset++];
+		}
+	}
+	// Read whole blocks at a time
+	while (bytes_left >= EEPROM_BLOCK_SIZE)
+	{
+		eeprom_read(current_block++, buf);
```
anacierdem, line 269: We should be able to read directly into `dest` here as well IINM. Maybe GCC will already optimize it out, but I'm not sure if it serves readability as well

anacierdem (commented):
I liked this leaner approach much better indeed! Looks OK to me as the API is simple and the implementation can be changed later on. A few notes;

- Adding tests for the new functions would be awesome as they will make their usage transparent for newcomers and can pinpoint subtle bugs or areas to improve.
- We may keep the read bytes (or even the whole currently valid range) in a cache and speed up consecutive reads to further simplify consuming these methods. This is something that can be implemented later as the API is already clean though.
- I have a minor comment to remove the additional `memcpy`

Let me know what you think!

rasky:
My take on this is that removing the stdio integration from #125 is probably the best overall solution. We still get most of the benefits of the filesystem implementation like static declaration of files, checksums, etc. but just with functions to read/write each file in a single go. 

The stdio integration is the part that was raising more concerns because it's hard to really simulate a full file system over a small EEPROM.

meeq:
Closing in favor of https://github.com/DragonMinded/libdragon/pull/125

PR #144, opened 2021-06-23
n64tool: return error in case the seek offset required backward seek.

### LINKED ISSUE #141: recurse_path hangs trying to open file.
Adapting the spritemap example about how to create an dfs filesystem on a different project, the function `dfs_open()` hangs on an infinite loop when executing `recurse_path`.

Even if I did something wrong (i.e., failed to initialize the dfs or it wasn't properly created), it should never hang on an infinite loop there. An error should be returned.

[edit]
The cause of the issue is `dfs_init()` failing to initialize the dfs, probably because the dfs offset is incorrect. How can I know exactly where the filesystem is located in memory?

rasky:
For instance, if the user asks to put a DFS at 1M, but the code/data is
2M, it simply can't be done. Instead of silently putting the DFS at 2M,
better return an error and let the user handle it, especially since
the seek offset needs to be passed to dfs_init() so the user needs
to be aware of the correct value.

Fixes #141

### DIFF
--- a/tools/n64tool.c
+++ b/tools/n64tool.c
@@ -165,7 +165,12 @@ int output_zeros(FILE *dest, int amount)
 		/* Don't support odd word alignments */
 		return -1;
 	}
-	
+	if(amount < 0)
+	{
+		/* Can't backward seek */
+		return -1;
+	}
+
 	int i;
 	while (amount > 0) {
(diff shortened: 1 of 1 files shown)


meeq (approved):
Seems reasonable 👍

PR #145, opened 2021-06-23
Make n64tool print dfs DEADBEEF address

giulianobelinassi:
This PR adds a logic for n64tool to print the initialization address required by `dfs_init()`. It also fixes the following minor issues I found in the code:

- The `zero` array was not being properly initialized. Using array[N] = {0} do not initialize all its elements to zero, but only the first element. The reason why this array was being initialized with 0 is because on x86_64, the .data segment is always initialized with 0. This is a latent bug that may show up in other platforms.
-  Reduced the number of calls to `fprintf` to one in `print_message`
-  Removed all trailing whitespaces and tabs.

### DIFF
--- a/tools/n64tool.c
+++ b/tools/n64tool.c
@@ -39,7 +39,7 @@
 /* Easier to write from here */
 static int title[TITLE_SIZE];
 static int wrote_title = 0;
-static unsigned char zero[1024] = {0};
+static unsigned char zero[1024];
 
 void print_usage(char *prog_name)
 {
@@ -230,6 +230,9 @@ int main(int argc, char *argv[])
 	/* Set default title */
 	memset(title, 0x20, TITLE_SIZE);
 	memcpy(title, DEF_TITLE, (strlen(DEF_TITLE) > TITLE_SIZE) ? TITLE_SIZE : strlen(DEF_TITLE));
+
+	/* Initialize zero array */
+	memset(zero, 0x00, 1024*sizeof(*zero));
 
 	if(argc <= 1)
 	{
--- a/tools/n64tool.c
+++ b/tools/n64tool.c
@@ -43,18 +43,21 @@ static unsigned char zero[1024] = {0};
 
 void print_usage(char *prog_name)
 {
-	fprintf(stderr, "Usage: %s [-b] -l <size>B/K/M -h <file> -o <file> -t <title> <file> [[-s <offset>B/K/M] <file>]*\n\n", prog_name);
-	fprintf(stderr, "This program appends a header to an arbitrary number of binaries,\n");
-	fprintf(stderr, "the first being an Nintendo64 binary and the rest arbitrary data.\n\n");
-	fprintf(stderr, "\t-b\t\tByteswap the resulting output.\n");
-	fprintf(stderr, "\t-l <size>\tForce output to <size> bytes.\n");
-	fprintf(stderr, "\t-h <file>\tUse <file> as header.\n");
-	fprintf(stderr, "\t-o <file>\tOutput is saved to <file>.\n");
-	fprintf(stderr, "\t-t <title>\tTitle of ROM.\n");
-	fprintf(stderr, "\t-s <offset>\tNext file starts at <offset> from top of memory.  Offset must be 32bit aligned.\n");
-	fprintf(stderr, "\t\t\tB for byte offset.\n");
-	fprintf(stderr, "\t\t\tK for kilobyte offset.\n");
-	fprintf(stderr, "\t\t\tM for megabyte offset.\n");
+	const char *message = ""
+	"Usage: %s [-b] -l <size>B/K/M -h <file> -o <file> -t <title> <file> [[-s <offset>B/K/M] <file>]*\n\n"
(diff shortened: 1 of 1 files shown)


tools/n64tool.c
```diff
@@ -39,33 +40,41 @@
 /* Easier to write from here */
 static int title[TITLE_SIZE];
 static int wrote_title = 0;
-static unsigned char zero[1024] = {0};
+static unsigned char zero[1024];
```
rasky, line 43: Actually, this is guaranteed by C standard. If you initialize less elements, the rest of the array is guaranteed to be zero initialized. It's not a latent bug, it's the way C works.

rasky:
I'm not sure I understand the rationale of this.

DragonFS is a file. N64tool adds it to the ROM at the requested position, that the user specified on the command line. There is no need to search for it: n64tool put it where the user requested.

There is a bug though: if you request a position that is not possible because it's too small, n64tool doesn't error out. I've fixed this (see #144). Once that is fixed, n64tool will always follow what the user asks, or error out.

So I'm not sure we need to print anything, as what we print would be exactly the same that was specified on the command line.

tools/n64tool.c
giulianobelinassi: True. Testing with:
```
void sink(void *);

int test()
{
    int arr[16] = {0};
    sink(arr);
}
```
will output
```
test:
.LFB0:
	.cfi_startproc
	pushq	%rbp
	.cfi_def_cfa_offset 16
	.cfi_offset 6, -16
	movq	%rsp, %rbp
	.cfi_def_cfa_register 6
	subq	$64, %rsp
	movq	$0, -64(%rbp)
	movq	$0, -56(%rbp)
	movq	$0, -48(%rbp)
	movq	$0, -40(%rbp)
	movq	$0, -32(%rbp)
	movq	$0, -24(%rbp)
	movq	$0, -16(%rbp)
	movq	$0, -8(%rbp)
	leaq	-64(%rbp), %rax
	movq	%rax, %rdi
	call	sink@PLT
	nop
	leave
```
Therefore, the array was indeed initialized with 0.

giulianobelinassi:
> I'm not sure I understand the rationale of this.
> 
> DragonFS is a file. N64tool adds it to the ROM at the requested position, that the user specified on the command line. There is no need to search for it: n64tool put it where the user requested.
> 
> There is a bug though: if you request a position that is not possible because it's too small, n64tool doesn't error out. I've fixed this (see #144). Once that is fixed, n64tool will always follow what the user asks, or error out.
> 
> So I'm not sure we need to print anything, as what we print would be exactly the same that was specified on the command line.

Well, the print would only give confirmation to the user that the DragonFS was detected at address X. Perhaps this should be changed to a "verbose" option instead of a "quiet"?

rasky:
There are other points to take into account:

 • n64tool is used to build N64 ROMs concatenating arbitrary files. DragonFS is just an option. Not everybody is using it, people can simply concatenate binary files and use their offset to access them. So adding a feature that is strictly dependent on DragonFS seems wrong
 • I don't see the need to scan the whole file that we just built. If you want to print the actual offset of concatenated files, just print the file position as it is written. But again, that position was requested by the user... it's not something that n64tool calculates by itself. I understand that you hit a bug that created problems and you got the idea that you don't have a way to know the position; but that bug is going to be fixed. But in general, it's the user that decides the position and n64tool just obeys.
 • Obviously there might be false positives. The string `DEAD BEEF` could compare elsewhere in the ROM, and in that case the address printed by `n64tool` would be wrong. So we're opening the door to a confirmation features that can indeed be wrong.
 • if you want to find out the position of a DragonFS inside a ROM, you can already use one of the many tools available for instance:

```
↪ xxd -u mvs64-mslug.z64 | grep "DEAD BEEF"
00101000: DEAD BEEF FFFF FFFF 4472 6167 6F6E 4653  ........DragonFS
```

so maybe you could add something like this to your Makefile after `n64tool`

DragonMinded:
Yeah, I'm not sure how I feel on this PR as well. There's no limit to the number of DragonFS filesystems that you can place in a ROM and its up to you to request which one you want. You can also choose no filesystem, or a different filesystem altogether coming from a flashcart, or write your own new filesystem and register it with newlib. I could even see a future update to allow multiple initialization calls to `dfs_init` with unique prefixes. It was never intended for there to be guaranteed a DragonFS filesystem or even the number of them. Also, 0xDEADBEEF is a fairly common initializer. Just because you found it does not mean there's a DragonFS filesystem at that location or that the location you found is the only location. There's more checking done in the init function and that shouldn't be replicated here. n64tool just concatenates ROMs with a (supposed) guaranteed offset. The fact that it does not error when it cannot provide that offset is a bug and should be fixed. At most, stdout/stderr printing in a verbose mode should just echo the sections that it has added.

giulianobelinassi:
Thinking through this, it makes sense for n64tool to be completely independent from dragonfs. I am closing this even through this commit have some code cleanup, but this can be pushed in other time.

anacierdem:
@giulianobelinassi feel free to open another PR

PR #146, opened 2021-06-23
Add missing extern "C" in debug.h

giulianobelinassi:
When compiling any .cpp file which calls any function defined
in debug.h, the linker fails to find its symbol because GCC
thinks its name is mangled. This commit fixes that by declaring
these symbols as having "C" names.


rasky (approved):
Thanks!

PR #147, opened 2021-06-24
Fix clipped RDP rectangle drawing

meeq:
Fixes this issue with Flappy Bird, which uses clipped rectangles for the pipes and backgrounds as they scroll off-screen:

Turns out making an unsigned integer negative yields *strange* results.


meeq:
This was broken by https://github.com/DragonMinded/libdragon/pull/70

anacierdem:
I think [this](https://stackoverflow.com/questions/8026694/c-unary-minus-operator-behavior-with-unsigned-operands) is relevant for an explanation.

vieux:
@meeq thanks for fixing this 👍

PR #148, opened 2021-06-24
Ensure C++ global constructors are not garbage collected by ld

### LINKED ISSUE #143: Linking with -Wl,--gc-sections results in C++ global ctors never being called
With the following code:
```
#include <cstdio>
#ifdef N64
#include <libdragon.h>
#endif

class Test
{
	public:
	const char *str;
	int a; int b;

	Test(const char *str1, int a1, int b1)
	: str(str1)
	, a(a1)
	, b(b1)
	{}
};

Test test_ctor("Hello", 42, 1379);

int main()
{
  #ifdef N64
  static resolution_t res = RESOLUTION_320x240;
  static bitdepth_t bit = DEPTH_32_BPP;

  init_interrupts();
  console_init();
  rdp_init();
  controller_init();
  timer_init();
  debug_init_isviewer();

  console_set_render_mode(RENDER_AUTOMATIC);
  #endif

  printf("First print\n");
  printf("Test: %s, %d, %d\n", test_ctor.str, test_ctor.a, test_ctor.b);
  return 0;
}
```
Compiling for x86_64 (g++ 10.2.1, glibc 2.31, linux 5.10.0-7) with `g++ a,cpp -Wl,--gc-sections` and executing, the output is:
```
First print
Test: Hello, 42, 1379
```
Compiling for Nintendo 64 (libdragon) with:
```
mips64-elf-g++ a.cpp -DN64 -o a.elf -ldragon -lc -lstdc++ -ldragonsys -Tn64.ld -Wl,--gc-sections
mips64-elf-objcopy a.elf a.bin -O binary                       
n64tool -l 1M -h /usr/local/mips64-elf/lib/header -o a.z64 a.bin
chksum64 a.z64                                                 
```
The output is:
```
First print
Test: (null), 0, 0
```
This issue vanishes if -Wl,--gc-sections is removed from the compilation flags. It means that the code used to initialize `test_ctor` was somehow removed when garbage collecting the functions.

meeq:
Resolves https://github.com/DragonMinded/libdragon/issues/143

@giulianobelinassi

### DIFF
--- a/n64.ld
+++ b/n64.ld
@@ -67,7 +67,7 @@ SECTIONS {
 
     .ctors : {
         __CTOR_LIST__ = .;
-        *(.ctors)
+        KEEP(*(.ctors))
         __CTOR_END__ = .;
     } > mem
 


giulianobelinassi:
Works fine for the testcase and my application. :)

PR #150, opened 2021-06-26
Fix fopen returning seemingly valid ptr if file not exist

### LINKED ISSUE #149: `fopen` returns seemingly valid pointer if file not found
With the following patch for spritemap.c:
```
diff --git a/examples/spritemap/spritemap.c b/examples/spritemap/spritemap.c
index 1174e52..46047f5 100644
--- a/examples/spritemap/spritemap.c
+++ b/examples/spritemap/spritemap.c
@@ -24,13 +24,19 @@ int main(void)
     rdp_init();
     controller_init();
     timer_init();
+    debug_init_isviewer();
 
     /* Read in single sprite */
     int fp = dfs_open("/mudkip.sprite");
     sprite_t *mudkip = malloc( dfs_size( fp ) );
     dfs_read( mudkip, 1, dfs_size( fp ), fp );
     dfs_close( fp );
-    
+
+    FILE* fptr = fopen("rom://should_not_exist", "r");
+    if (!fptr)
+        debugf("File not found\n");
+    fclose(fptr);
+
     fp = dfs_open("/earthbound.sprite");
     sprite_t *earthbound = malloc( dfs_size( fp ) );
     dfs_read( earthbound, 1, dfs_size( fp ), fp );
```
Clearly, `should_not_exist` is not present in the DragonFS. However, running with:
```
cen64 -is-viewer ~/bagunca/n64pif/pifdata.bin spritemap.z64
```
The __file not found__ message is never print, and a seemly valid (but invalid) pointer is returned, where it should have returned `NULL`.

giulianobelinassi:
Fixes #149

### DIFF
--- a/src/system.c
+++ b/src/system.c
@@ -883,7 +883,7 @@ int open( char *file, int flags, int mode )
  
             void *ptr = fs->open( file + __strlen( filesystems[mapping].prefix ), flags );
 
-            if( ptr )
+            if( ptr && ((int32_t) ptr) != -2 )
             {
                 /* Create new internal handle */
                 handles[i].fileno = __get_new_handle();


src/system.c
```diff
@@ -883,7 +883,7 @@ int open( char *file, int flags, int mode )
  
             void *ptr = fs->open( file + __strlen( filesystems[mapping].prefix ), flags );
 
-            if( ptr )
+            if( ptr && ((int32_t) ptr) != -2 )
```
DragonMinded, line 886: This seems to be a very specific magic value testing against a particular implementation of a filesystem and probably wouldn't work given any other circumstances. Did you track down why we're getting `-2` returned in the first place?

giulianobelinassi (commented):
Posting here because my message got stuck in `pending` state on comment section

The -2 value came from dfs_open() returning an error. I did not look deeply into the code to check if any other function had to modify this value.

rasky:
I think this was already fixed here: https://github.com/DragonMinded/libdragon/commit/86c824cf13f3948843a11c2af7bbed4509a99ce4

giulianobelinassi:
Already fixed by **86c824c**

PR #151, opened 2021-06-29
First kernel prototype

rasky:
This PR introduces a multithreading kernel to libdragon. Having threads
is a very good way to make good use of the limited resources available
on the console as it makes it easier to switch away and do something
while a DMA is in progress or while the RSP is crunching. In theory,
everything can be done without threads, but in practice we posit that most
homebrew developers will be better served with a threading module.

In my upcoming audio library, threading Is basically necessary as processing
audio requires switching between CPU and RSP several times, and doing
that only with interrupts gets complicated fast. On the contrary, with threads,
it's very easy to keep audio processing in background.

This PR just introduces the kernel. It doesn't attempt to revisit all existing 
libdragon APIs to do some thread-aware yield instead of spin-looping. That's
something for another day. Also, we will have to discuss whether it makes sense
for APIs to automatically use threads if available and fallback to spinloops, or
it's better to have a different set of APIs. A good example to reason about is
`dma_read` or even `wait_ms`.

## Description of the kernel module

This module implements a hybrid cooperative/preemptive multi-threaded
kernel for parallel execution of code.

The kernel is "hard real-time": any thread with high priority that is
ready (not blocked) will always have priority over lower priority threads.
This is required to allow threads to implement operations that require
low latency, for instance preparing audio when the AI interrupt fires.

The kernel is not fully preemptive; in particular, there is no timer
interrupt that switches among ready threads at a fixed interval. This
is not deemed necessary as applications will usually have a low number
threads that are mostly blocked waiting for specific events like
hardware interrupts or background activities like RSP ucode.

An event, represented by a kevent_t instance, is a signal that can
be broadcasted to many different threads. A thread can use
kthread_wait_event to wait for a specific event to fire. The kernel
library includes a set of events bound to MI peripheral interrupts,
so that it's possible for threads to wait for specific interrupts and
act after them.

Typically, applications will keep the main thread at low priority
for the main business logic, and will spawn a few threads to handle
high-priority / low-latency responses to hardware events. Libdragon
itself offers libraries that need threading to provide a easy-to-use
API.

A context-switch between threads can happen in the following situations:

  * Preemptively: when a hardware interrupt generates a event that
    wakes up a thread (see kevent). This is the most common scenario:
    an event connected to an interrupt is fired, and a higher priority
    thread waiting for the event is woken up.

  * Cooperatively: when a thread explicitly calls a blocking function
    like kthread_sleep or kthread_wait_event to wait for an event.
    The kernel will put the current thread to sleep and schedule the
    highest-priority thread among the ready ones, including threads
    that have a priority lower than the one that goes to sleep. When
    the event is fired, the lower-priority thread will be preempted
    to give back control to the thread that just woke up.

  * Cooperatively: when a thread explicitly calls kthread_yield to
    leave the CPU to other threads that might be ready at the same time.
    Notice that, given the hard real-time guarantee, this only makes
    sense if there are multiple ready threads at the same priority level,
    as calling kthread_yield will not schedule a lower-priority thread.
    It is normally not necessary to call thread_yield.

In addition to threads (kthread_t) and events (kevent_t), the kernel
library offers a message passing communication primitive called "message
box" (kmbox_t). A mailbox is a circular buffer of messages called
"mails"; any number of threads can try to send mails into a mailbox,
and any number of threads can try to read mails from a mailbox. Once a
thread receives a mail from a mailbox, the mail is removed and other
threads will not see it anymore. 

Mailboxes can be useful as communicative primitives among threads
(eg: to send tasks to other threads), or a synchronization primitive
to wait for multiple events (see kmbox_attach_event).

### DIFF
--- a/include/kernel.h
+++ b/include/kernel.h
@@ -93,6 +93,11 @@
  * #kthread_new returns a pointer to the thread (a pointer
  * to kthread_t) that can be used to externally manage the
  * thread like changing its priority or killing it.
+ * 
+ * NOTE: this structure should be considered internal and subject
+ * to future changes. It is exposed just for debugging purposes
+ * and should not be relied upon. Please do not change any
+ * field without going through a function of the public API.
  */
 typedef struct kthread_s
 {
@@ -137,6 +142,11 @@ typedef struct kthread_s
  * simply return a failure in that case.
  *
  * Threads can receive mails using #kmbox_recv or #kmbox_try_recv.
+ *
+ * NOTE: this structure should be considered internal and subject
+ * to future changes. It is exposed just for debugging purposes
+ * and should not be relied upon. Please do not change any
+ * field without going through a function of the public API.
  */
 typedef struct kmbox_s
 {
@@ -169,6 +179,11 @@ typedef struct kmbox_s
  *
  * It is also possible to attach a mailbox to an event using #mbox_attach_event.
  * This is useful if a thread needs to wait for multiple events.
+ *
+ * NOTE: this structure should be considered internal and subject
+ * to future changes. It is exposed just for debugging purposes
+ * and should not be relied upon. Please do not change any
+ * field without going through a function of the public API.
  */
 typedef struct kevent_s
 {
@@ -354,7 +369,7 @@ const char* kthread_name(kthread_t *th);
  *
  * If the thread needs to handle multiple events and/or a more
  * elaborate communication with other threads, you can use
- * #thread_listen_event instead.
+ * #kmbox with #kmbox_attach_event instead.
  *
  * @param[in] evt
  *            The event to wait for
--- a/include/kernel.h
+++ b/include/kernel.h
@@ -1,0 +1,529 @@
+/**
+ * @file kernel.h
+ * @brief Multi-threading kernel
+ * @ingroup kernel
+ */
+#ifndef KERNEL_H
+#define KERNEL_H
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+
+#include "exception.h"
+
+/**
+ * @defgroup kernel Multi-threading kernel
+ * @ingroup libdragon
+ * @brief A multi-threading kernel that can be used by libdragon applications.
+ *
+ * This module implements a hybrid cooperative/preemptive multi-threaded
+ * kernel for parallel execution of code.
+ *
+ * The scheduler uses a very simple logic:
+ * 
+ *   * A thread is "ready" whenever is able to run, that is, it is not waiting
+ *     for some event or otherwise sleeping. A ready thread can technically be
+ *     scheduled at any time, whenever the scheduler decides so.
+ *   * When a thread switch happens, the scheduler selects the ready thread which
+ *     has the highest priority.
+ *   * If there are multiple threads with the same priority (higher than any
+ *     other ready thread), the scheduler will round-robin among them.
+ * 
+ * A thread switch can happen in a few specific situations:
+ * 
+ *   * The active thread explicitly starts waiting for some event, or sleeps
+ *     for a definite amount of time. In general, whenever a kthread_* function
+ *     is invoked, it might cause a context switch.
+ *   * An interrupt occurs, which triggers an event, which wakes up some thread
+ *     whose priority is higher than the currently running thread.
+ * 
+ * Given the above rules, we can say that the kernel is "hard real-time": any
+ * thread with high priority that is ready (not blocked) will always have
+ * priority over lower priority threads. This is required to allow threads to
+ * implement operations that require low latency, for instance preparing audio
+ * when the AI interrupt fires.
+ *
+ * Also, the kernel is not fully preemptive; in particular, there is no timer
+ * interrupt that switches among ready threads at a fixed interval. This
+ * is not deemed necessary as applications will usually have a low number
+ * threads that are mostly blocked waiting for specific events like
+ * hardware interrupts or background activities like RSP ucode.
+ * 
+ * An event, represented by a #kevent_t instance, is a signal that can
+ * be broadcasted to many different threads. A thread can use
+ * #kthread_wait_event to wait for a specific event to fire. The kernel
+ * library includes a set of events bound to MI peripheral interrupts,
+ * so that it's possible for threads to wait for specific interrupts and
+ * act after them.
+ *
+ * Typically, applications will keep the main thread at low priority
+ * for the main business logic, and will spawn a few threads to handle
+ * high-priority / low-latency responses to hardware events.
+ *
+ * In addition to threads (#kthread_t) and events (#kevent_t), the kernel
+ * library offers a message passing communication primitive called "mail
+ * box" (#kmbox_t). A mailbox is a circular buffer of messages called
+ * "mails"; any number of threads can try to send mails into a mailbox,
+ * and any number of threads can try to read mails from a mailbox. Once a
+ * thread receives a mail from a mailbox, the mail is removed and other
+ * threads will not see it anymore. 
+ *
+ * Mailboxes can be useful as communicative primitives among threads
+ * (eg: to send tasks to other threads), or a synchronization primitive
+ * to wait for multiple events (see #kmbox_attach_event).
+ *
+ * @{
+ */
+
+/**
+ * @brief a kernel thread for parallel execution
+ *
+ * This structure represents a thread that can be scheduled
+ * for parallel execution. Create a thread with #kthread_new,
+ * allocating the required memory for the stack.
+ *
+ * Normally, there's no need to explicitly manage the lifetime
+ * or state of a thread. Once the thread is created, it is
+ * immediately started, and the kernel will scheduled it when
+ * required. If the thread exits (by simply returning by its
+ * entry point function), it will be cleaned up and the memory
+ * released.
+ *
+ * #kthread_new returns a pointer to the thread (a pointer
+ * to kthread_t) that can be used to externally manage the
+ * thread like changing its priority or killing it.
+ */
+typedef struct kthread_s
+{
+	/** Pointer to the top of the stack, which contains the thread register state */
+	reg_block_t* stack_state;
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
+	/** Size of the stack in bytes */
+	int stack_size;
+	/** Name of thread (for debugging purposes) */
+	const char *name;
+	/** Internal flags */
+	uint8_t flags;
+	/** Priority of the thread (0=lowest, use positive numbers only) */
+	int8_t pri;
+	/** Intrusive link to next thread in a waiting list */
+	struct kthread_s *next;
+	/** Entry point function for the thread */
+	void (*user_entry)(void*);
+	/** Custom argument to be passed to the entry point */
+	void *user_data;
+	/** Pointer to the stack buffer */
+	void *stack;
+} kthread_t;
+
+/**
+ * @brief A kernel mailbox.
+ *
+ * A mailbox is a circular buffer of messages called
+ * "mails"; any number of threads can try to send mails into a mailbox,
+ * and any number of threads can try to read mails from a mailbox. Once a
+ * thread receives a mail from a mailbox, the mail is removed and other
+ * threads will not see it anymore.
+ *
+ * A mailbox can be allocated either on the heap (#mbox_new) or on
+ * the stack (#mbox_new_stack). In both cases, the size of the buffer
+ * (maximum number of mails that can be kept within the mailbox)
+ * must be specified.
+ *
+ * Threads can send mails using #kmbox_send or #kmbox_try_send: the former
+ * will block the thread if the mailbox is full, while the later will
+ * simply return a failure in that case.
+ *
+ * Threads can receive mails using #kmbox_recv or #kmbox_try_recv.
+ */
+typedef struct kmbox_s
+{
+	/** Read index of the circular buffer */
+	uint8_t r;
+	/** Write index of the circular buffer */
+	uint8_t w;
+	/** Size of the buffer (maximum number of mails) */
+	uint8_t max;
+	/** Number of events this mailbox is attached to */
+	uint8_t evts;
+	/** Linked list of threads waiting on this mailbox */
+	kthread_t *wait_list;
+	/** Circular buffer of mails */
+	void* mails[0];
+} kmbox_t;
+
+/** Maximum number of mailboxes that can be attached to a single event. */
+#define MAX_MAILBOXES_PER_EVENT 8
+
+/**
+ * @brief a kernel event.
+ *
+ * An event, represented by a #kevent_t instance, is a signal that can
+ * be broadcasted to many different threads. A thread can use
+ * #kthread_wait_event to wait for a specific event to fire. The kernel
+ * library includes a set of events bound to MI peripheral interrupts,
+ * so that it's possible for threads to wait for specific interrupts and
+ * act after them.
+ *
+ * It is also possible to attach a mailbox to an event using #mbox_attach_event.
+ * This is useful if a thread needs to wait for multiple events.
+ */
+typedef struct kevent_s
+{
+	/** Mailboxes attached to this event */
+	kmbox_t *mboxes[MAX_MAILBOXES_PER_EVENT];
+} kevent_t;
+
+
+/** @brief Event generated when a SP interrupt is triggered */
+extern kevent_t KEVENT_IRQ_SP;
+/** @brief Event generated when a SI interrupt is triggered */
+extern kevent_t KEVENT_IRQ_SI;
+/** @brief Event generated when a AI interrupt is triggered */
+extern kevent_t KEVENT_IRQ_AI;
+/** @brief Event generated when a VI interrupt is triggered */
+extern kevent_t KEVENT_IRQ_VI;
+/** @brief Event generated when a PI interrupt is triggered */
+extern kevent_t KEVENT_IRQ_PI;
+/** @brief Event generated when a DP interrupt is triggered */
+extern kevent_t KEVENT_IRQ_DP;
+/** @brief Event generated when the RESET button is pushed (pre-NMI exception) */
+extern kevent_t KEVENT_RESET;
+
+
+/** 
+ * @brief Initialize the multi-threading kernel
+ *
+ * The current execution context becomes the main thread of the
+ * program, with priority set to 0 (you can change priority
+ * to the main thread as well using #thread_set_pri).
+ *
+ * The main thread uses the original stack allocated for the
+ * whole process, so it is technically unbounded (or limited by
+ * heap size).
+ *
+ * @return a pointer to the main thread.
+ *
+ */
+kthread_t* kernel_init(void);
+
+/** 
+ * @brief Shutdown the multi-threading kernel
+ *
+ * This function is mostly useful for testing purposes.
+ * Since the kernel does not keep track of all created threads,
+ * this function should be called only when all created threads
+ * are exited or have been killed.
+ */
+void kernel_close(void);
+
+/**
+ * @brief Create a new thread
+ *
+ * Create a new thread, with a specified stack size and priority.
+ * The thread is immediately made ready after creation, so if it has
+ * a priority higher than or equal to the current thread, it will be
+ * scheduled immediately, before thread_new returns.
+ *
+ * The thread will begin execution from the specified entry point
+ * function. If the function ever returns, the thread is automatically
+ * killed.
+ *
+ * @param[in] name
+ *            Name of the thread (for debugging purposes)
+ * @param[in] stack_size
+ *            Size of the stack in bytes. Minimum suggested size
+ *            is 2048.
+ * @param[in] pri
+ *            Priority of the thread (0-127). Negative values are not allowed.
+ *            Higher number means higher priority.
+ * @param[in] user_entry
+ *            Entry point of the thread. This is the function that will
+ *            be called when the thread begins execution. If this function
+ *            ever returns, the thread is automatically killed.
+ * @param[in] user_data
+ *            Argument that will be passed to the entry point of the thread.
+ *
+ * @return    A pointer to the new thread. It is not necessary to store
+ *            this reference if not required; the thread will clean up
+ *            after itself when it exits.
+ */
+kthread_t* kthread_new(const char *name, int stack_size, int8_t pri,
+	void (*user_entry)(void*), void *user_data);
+
+
+/**
+ * @brief Yield execution of the current thread and run the scheduler.
+ *
+ * This function allows the current thread to cooperatively yield
+ * its execution to allow other threads to run.
+ *
+ * The scheduler will switch to the highest priority thread that is
+ * currently ready to run. If no ready thread has a priority higher than
+ * the current thread, the scheduler will switch to a different thread 
+ * of the same priority of the current one (the scheduler will guarantee
+ * a correct round-robin scheduling among threads of the same priority).
+ * If no ready thread has a priority higher or equal than the current
+ * one, the scheduler will reschedule the current thread, that will
+ * continue execution.
+ *
+ * @note The scheduler is semi-preemptive. Any interrupt could cause
+ * a thread switch to happen if the interrupt itself makes a thread
+ * become ready. If you need a block of code to be executed without
+ * any context switch, make sure to disable interrupts.
+ */
+void kthread_yield(void);
+
+
+/**
+ * @brief Sleep for the specified interval, allowing execution of other threads.
+ *
+ * This function will put the current thread to sleep for a specified
+ * time interval, allowing other threads to run.
+ *
+ * The sleeping interval is expressed in hardware ticks. See #TICKS_READ
+ * for an explanation of hardware ticks, and #TICKS_PER_SECOND and #TICKS_FROM_MS
+ * for constants and macros that simplify conversion to standard time units.
+ *
+ * Notice that #wait_ticks and #wait_ms are not threading aware, so they will
+ * just spin-wait while keeping the current thread ready. Other higher-priority
+ * threads might still be scheduled if they are awaken by an interrupt, but in
+ * general thread_sleep is a better primitive if you want the current thread
+ * to yield.
+ *
+ * @note This function requires the timer module, so #timer_init must have
+ *       been called.
+ *
+ * @param[in] ticks
+ *            The number of hardware ticks to sleep.
+ */
+void kthread_sleep(uint32_t ticks);
+
+/** 
+ * @brief Change priority of a thread.
+ * 
+ * Change priority of the specified thread. If the argument is NULL,
+ * this function changes priority to the current thread.
+ *
+ * The change of priority is immediately effective. It may cause
+ * a context switch if the changed thread is ready and its priority
+ * is changed in a way to start/stop it relative to the other ready
+ * threads.
+ *
+ * @param[in]  th
+ *             Reference to the thread (NULL = current thread)
+ * @param[in]  pri
+ *             New priority of the thread (0-127). Higher number means
+ *             higher priority.
+ *
+ */
+void kthread_set_pri(kthread_t *th, int8_t pri);
+
+
+/** 
+ * @brief Kill a thread, aborting its execution.
+ *
+ * The specified thread is aborted, and its memory freed (including
+ * its stack). This function can executed for any thread, including the
+ * current one. The execution will be aborted and the memory released
+ * to the heap for further use.
+ *
+ * @param[in]  th
+ *             Reference to the thread to kill (NULL = current thread)
+ *
+ */
+void kthread_kill(kthread_t *th);
+
+
+/**
+ * @brief Return the name of the specified thread.
+ *
+ * @param[in]  th
+ *             Reference to thread (NULL = current thread)
+ */
+const char* kthread_name(kthread_t *th);
+
+/**
+ * @brief Wait for a single hardware event.
+ *
+ * This functions blocks the current thread until a specified
+ * hardware event happens. The CPU is yielded so that other
+ * threads will be scheduled.
+ *
+ * If the thread needs to handle multiple events and/or a more
+ * elaborate communication with other threads, you can use
+ * #thread_listen_event instead.
+ *
+ * @param[in] evt
+ *            The event to wait for
+ *
+ */
+void kthread_wait_event(kevent_t *evt);
+
+/**
+ * @brief Allocate a new mailbox (on the heap).
+ *
+ * @param[in] max_mails
+ *            Maximum number of pending mails in the mailbox.
+ *
+ * @return The allocated mailbox
+ *
+ * @note See #mbox_new_stack to allocate a mailbox on the stack.
+ */
+kmbox_t* kmbox_new(int max_mails);
+
+/**
+ * @brief Allocate a new mailbox (on the stack).
+ *
+ * @param[in] max_mails
+ *            Maximum number of pending mails in the mailbox.
+ *
+ * @return The allocated mailbox
+ *
+ * @note See #mbox_new to allocate a mailbox on the heap.
+ */
+#define kmbox_new_stack(max_mails) ({ \
+	int sz = sizeof(kmbox_t) + (max_mails)*sizeof(void*); \
+	kmbox_t *mb = alloca(sz); \
+	bzero(mb, sz); \
+	mb->max = max_mails; \
+	mb; \
+})
+
+/**
+ * @brief Free a heap mailbox when it's not needed anymore.
+ *
+ * @note Do not call this function on a mailbox allocated with
+ *       #kmbox_new_stack. As the mailbox is on the stack, there is
+ *       no need to explicitly free it.
+ */
+void kmbox_free(kmbox_t *mbox);
+
+/** @brief Return true if the mailbox is empty */
+bool kmbox_empty(kmbox_t *mbox);
+
+/** @brief Return true if the mailbox is full */
+bool kmbox_full(kmbox_t *mbox);
+
+/** 
+ * @brief Try sending a mail to a mailbox. 
+ *
+ * This function tries to send a mail to a mailbox. If the
+ * mailbox is full, it returns false without blocking.
+ *
+ * @param[in] mbox The mailbox to send the mail to
+ * @param[in] mail The mail to send (must not be NULL)
+ *
+ * @return true if the mail was sent, false if the mailbox was full
+ */
+bool kmbox_try_send(kmbox_t *mbox, void *mail);
+
+/** 
+ * @brief Send a mail to a mailbox. 
+ *
+ * This function sends a mail to a mailbox. If the
+ * mailbox is full, it will block until another thread
+ * receive a pending mail, so that the mailbox has space
+ * for the new mail.
+ *
+ * @param[in] mbox The mailbox to send the mail to
+ * @param[in] mail The mail to send (must not be NULL)
+ *
+ */
+void kmbox_send(kmbox_t *mbox, void *mail);
+
+/** 
+ * @brief Try receiving a mail from a mailbox. 
+ *
+ * This function tries to receive a mail to a mailbox. If the
+ * mailbox is empty, it will return NULL without blocking.
+ *
+ * @param[in] mbox The mailbox to receive the mail from
+ *
+ * @return the received mail, or NULL if the mailbox was empty
+ *
+ */
+void* kmbox_try_recv(kmbox_t *mbox);
+
+/** 
+ * @brief Receive a mail from a mailbox. 
+ *
+ * This function receives a mail from a mailbox. If the
+ * mailbox is empty, it will block until a mail is available.
+ *
+ * @param[in] mbox The mailbox to receive the mail from
+ *
+ * @return the received mail
+ *
+ */
+void* kmbox_recv(kmbox_t *mbox);
+
+
+/**
+ * @brief Attach an event to a mailbox, so that a mail arrives when the event
+ *        is triggered.
+ *
+ * This functions connects a mailbox to the specified event, so that
+ * when the event is triggered, a mail is posted to the
+ * mailbox. The thread can then poll the mailbox to see when the
+ * event is arrived. The received mail will be the event itself.
+ *
+ * This function is useful for threads that need to react to multiple
+ * events and/or handle messages from other threads. If you just
+ * want to wait for a single interrupt, it is easier to simply
+ * call #thread_wait_event.
+ *
+ * Example:
+ *
+ *     kmbox_t *mbox = kmbox_new_stack(4);
+ *     kmbox_attach_event(mbox, &KEVENT_IRQ_SP);
+ *     kmbox_attach_event(mbox, &KEVENT_IRQ_DP);
+ *
+ *     void *mail;
+ *     while ((mail = kmbox_recv(mbox)))
+ *     {
+ *        if (mail == &KEVENT_IRQ_SP)
+ *        {
+ *            [...]
+ *        }
+ *        else if (mail == &KEVENT_IRQ_DP)
+ *        {
+ *            [...]
+ *        }
+ *     }
+ *
+ *     // Always detach mailboxes before freeing them!
+ *     kmbox_detach_event(mbox, &KEVENT_IRQ_SP);
+ *     kmbox_detach_event(mbox, &KEVENT_IRQ_DP);
+ *
+ *
+ * @note It is mandatory to detach all attached events before
+ *       destroying a mailbox.
+ *
+ * @note If the mailbox happens to be full when the event is triggered,
+ *       the event will be missed. When you attach a mailbox to an event,
+ *       make always sure to have enough space in the mailbox.
+ *
+ * @param[in] mbox The mailbox that will receive the event
+ * @param[in] evt The event to attach to
+ */
+void kmbox_attach_event(kmbox_t *mbox, kevent_t *evt);
+
+
+/**
+ * @brief Detach an event from a mailbox.
+ *
+ * @note It is mandatory to detach all attached events before
+ *       destroying a mailbox.
+ *
+ * @param[in] mbox The mailbox that is receiving the event
+ * @param[in] evt The event to detach from
+ */
+void kmbox_detach_event(kmbox_t *mbox, kevent_t *evt);
+
+
+/** @} */
+
+#endif /* KERNEL_H */
--- a/src/inthandler.S
+++ b/src/inthandler.S
@@ -134,13 +134,17 @@ exception_critical:
 	nop
 
 exception_syscall:
-	# Syscall exception
+	# Run the syscall exception handler. This one can optionally
+	# switch to a different thread: the return value is the new
+	# stack pointer to use.
 	jal __onSyscallException
 	addiu a0, sp, 32
 
-	j end_interrupt
-	nop
-
+	# Exit the exception and restore *all* registers including caller-saved
+	# ones, because if the syscall handler switched to a different thread,
+	# the new thread might have different values in those registers.
+	j end_exception_restore_all
+	addiu sp, v0, -32
 
 exception_coprocessor:
 	# Extract CE bits (28..29) from CR
@@ -264,9 +268,60 @@ notcart:
 	jal __MI_handler
 	addiu a0, sp, 32
 
-	# No more interrupts to process, we can exit
-	# (fallthrough)
+	# No more interrupts to process
 	sw zero, interrupt_exception_frame
+
+	# Check if some interrupt handler requested
+	# a reschedule to happen
+	la t0, __isr_force_schedule
+	lbu t1, (t0)
+	beqz t1, end_interrupt
+	sb zero, (t0)
+
+thread_schedule_from_interrupt:
+	# Save full exception context in preparation for stack switching
+	mfc0 t0, C0_SR
+	or t0, SR_CU1
+	mtc0 t0, C0_SR
+	jal save_fpu_regs
+	move a0, sp
+	jal finalize_exception_frame
+	nop
+
+	# Call the scheduler
+	jal __kthread_syscall_schedule
+	addiu a0, sp, 32
+	addiu sp, v0, -32
+	# fallthrough to end_exception_restore_all
+
+end_exception_restore_all:
+	# Restore also caller-saved registers. These are only saved during an exception
+	# (by finalize_exception_frame), and we need to restore them only in case of
+	# kernel stack switch (because otherwise, the C handler for an exception will
+	# never modify them without restoring them too, so it would be useless to
+	# restore them here).
+	ld $16,(STACK_GPR+16*8)(sp)   # S0
+	ld $17,(STACK_GPR+17*8)(sp)   # S1
+	ld $18,(STACK_GPR+18*8)(sp)   # S2
+	ld $19,(STACK_GPR+19*8)(sp)   # S3
+	ld $20,(STACK_GPR+20*8)(sp)   # S4
+	ld $21,(STACK_GPR+21*8)(sp)   # S5
+	ld $22,(STACK_GPR+22*8)(sp)   # S6
+	ld $23,(STACK_GPR+23*8)(sp)   # S7
+	ld $28,(STACK_GPR+28*8)(sp)   # GP
+	ld $30,(STACK_GPR+30*8)(sp)   # FP
+	ldc1 $f20,(STACK_FPR+20*8)(sp)
+	ldc1 $f21,(STACK_FPR+21*8)(sp)
+	ldc1 $f22,(STACK_FPR+22*8)(sp)
+	ldc1 $f23,(STACK_FPR+23*8)(sp)
+	ldc1 $f24,(STACK_FPR+24*8)(sp)
+	ldc1 $f25,(STACK_FPR+25*8)(sp)
+	ldc1 $f26,(STACK_FPR+26*8)(sp)
+	ldc1 $f27,(STACK_FPR+27*8)(sp)
+	ldc1 $f28,(STACK_FPR+28*8)(sp)
+	ldc1 $f29,(STACK_FPR+29*8)(sp)
+	ldc1 $f30,(STACK_FPR+30*8)(sp)
+	ldc1 $f31,(STACK_FPR+31*8)(sp)
 
 end_interrupt:
 	mfc0 t0, C0_SR
@@ -400,3 +455,49 @@ inthandler_end:
 	.p2align 2
 	.lcomm interrupt_exception_frame, 4
 
+	ld k1,STACK_LO(k0)
+	mtlo k1
+
+	ld k1,STACK_HI(k0)
+	mthi k1
+
+	ldc1 $f0,(STACK_FPR+0*8)(k0)
+	ldc1 $f1,(STACK_FPR+1*8)(k0)
+	ldc1 $f2,(STACK_FPR+2*8)(k0)
+	ldc1 $f3,(STACK_FPR+3*8)(k0)
+	ldc1 $f4,(STACK_FPR+4*8)(k0)
+	ldc1 $f5,(STACK_FPR+5*8)(k0)
+	ldc1 $f6,(STACK_FPR+6*8)(k0)
+	ldc1 $f7,(STACK_FPR+7*8)(k0)
+	ldc1 $f8,(STACK_FPR+8*8)(k0)
+	ldc1 $f9,(STACK_FPR+9*8)(k0)
+	ldc1 $f10,(STACK_FPR+10*8)(k0)
+	ldc1 $f11,(STACK_FPR+11*8)(k0)
+	ldc1 $f12,(STACK_FPR+12*8)(k0)
+	ldc1 $f13,(STACK_FPR+13*8)(k0)
+	ldc1 $f14,(STACK_FPR+14*8)(k0)
+	ldc1 $f15,(STACK_FPR+15*8)(k0)
+	ldc1 $f16,(STACK_FPR+16*8)(k0)
+	ldc1 $f17,(STACK_FPR+17*8)(k0)
+	ldc1 $f18,(STACK_FPR+18*8)(k0)
+	ldc1 $f19,(STACK_FPR+19*8)(k0)
+	ldc1 $f20,(STACK_FPR+20*8)(k0)
+	ldc1 $f21,(STACK_FPR+21*8)(k0)
+	ldc1 $f22,(STACK_FPR+22*8)(k0)
+	ldc1 $f23,(STACK_FPR+23*8)(k0)
+	ldc1 $f24,(STACK_FPR+24*8)(k0)
+	ldc1 $f25,(STACK_FPR+25*8)(k0)
+	ldc1 $f26,(STACK_FPR+26*8)(k0)
+	ldc1 $f27,(STACK_FPR+27*8)(k0)
+	ldc1 $f28,(STACK_FPR+28*8)(k0)
+	ldc1 $f29,(STACK_FPR+29*8)(k0)
+	ldc1 $f30,(STACK_FPR+30*8)(k0)
+	ldc1 $f31,(STACK_FPR+31*8)(k0)
+
+	lw k1, STACK_FC31(k0)
+	ctc1 k1, $f31
+
+	.set noat
+	ld $1,(STACK_GPR+1*8)(k0)
+	eret
+	nop
--- a/src/kernel.c
+++ b/src/kernel.c
@@ -208,7 +208,7 @@ reg_block_t* __kthread_syscall_schedule(reg_block_t *stack_state)
 				// the thread must already have been added to a waiting list,
 				// otherwise it wouldn't ever be scheduled again.
 				assert(th_cur->flags & TH_FLAG_INLIST);
-				assert(*(uint32_t*)th_cur->stack_state->epc == 0x4C);
+				assertf(*(uint32_t*)th_cur->stack_state->epc == 0x0000004C, "invalid opcode found by __kthread_syscall_schedule:\nexpected 0x0000004C (SYSCALL 0x1), found: %08lx", *(uint32_t*)th_cur->stack_state->epc);
 				th_cur->stack_state->epc += 4;
 			}
 			else
--- a/src/kernel.c
+++ b/src/kernel.c
@@ -1,0 +1,645 @@
+#include <libdragon.h>
+#include "kernel.h"
+#include "kernelinternal.h"
+#include "timer.h"
+#include <assert.h>
+#include <stdlib.h>
+#include <memory.h>
+
+#define DEBUG_KERNEL   0
+
+#define STACK_COOKIE   0xDEADBEEFBAADC0DE
+#define STACK_GUARD    64
+
+/** @brief Read the current value of the gp register */
+#define REG_GP()     ({ uint64_t gp; __asm("move %0, $28": "=r" (gp)); gp; })
+
+/**
+ * @brief Execute a context switch.
+ *
+ * This forces a context switch to happen, by invoking the syscall 0x1. The
+ * syscall will be intercepted by the generic interrupt handler (inthandler.S)
+ * that will call #__kthread_syscall_schedule.
+ *
+ * This macro is very low level and is called as part of higher-level primitives
+ * that force a context switch like #thread_yield.
+ *
+ * Do not call this under interrupt; use #KTHREAD_SWITCH_ISR instead.
+ */
+#define KTHREAD_SWITCH()      __asm("syscall 0x1")
+
+/**
+ * @brief Executed a context switch, during an ongoing interrupt service routing.
+ *
+ * This forces a context switch to happen just like #KTHREAD_SWITCH, but it is
+ * the right version to call if the code is currently servicing an interrupt.
+ */
+#define KTHREAD_SWITCH_ISR()  ({ __isr_force_schedule = true; })
+
+/** @brief Main thread */
+kthread_t th_main;
+/** @brief Pointer to the current thread */
+kthread_t *th_cur;
+/** @brief Pounter to the idle thread */
+kthread_t *th_idle;
+/** @brief List of ready threads */
+static kthread_t *th_ready;
+/** @brief Number of live thread */
+static int th_count;
+
+kevent_t KEVENT_IRQ_AI = {0};
+kevent_t KEVENT_IRQ_DP = {0};
+kevent_t KEVENT_IRQ_SP = {0};
+kevent_t KEVENT_IRQ_SI = {0};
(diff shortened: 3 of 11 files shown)


joeldipops:
Incredible stuff Rasky.
My initial and superficial feedback is just about naming consistency, internally and with libdragon.  
Within kernal.h, the same concept is referred to as a message box, a mail box and as kmbox.  I think a single name should be applied throughout.   Libdragon usually spells things out in full, so personally I like `kernel_mailbox_t` etc.  

The `k` prefix also seems a bit inconsistent with libdragon's naming, but I'm undecided on whether every function here being called kernel_blah_blah_blah is a better solution.

`kmbox_empty` reads a bit ambiguously to me.  In my projects these would be is_empty and is_full, though the only places libdragon uses that convention are from my own PRs, so I don't have much to justify myself there.

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+typedef struct kthread_s
+{
+	/** Pointer to the top of the stack, which contains the thread register state */
+	reg_block_t* stack_state;
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
+	/** Size of the stack in bytes */
+	int stack_size;
+	/** Name of thread (for debugging purposes) */
+	char name[17];
```
danbolt: Nitpick, but would you feel comfortable having the `17` here `#define`d as a constant? This might help with copying from `strncpy`-style functions.

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ * general thread_sleep is a better primitive if you want the current thread
+ * to yield.
+ *
+ * @note This function requires the timer module, so #timer_init must have
+ *       been called.
+ *
+ * @param[in] ticks
+ *            The number of hardware ticks to sleep.
+ */
+void kthread_sleep(uint32_t ticks);
```
danbolt, line 317: Just a nitpick to do with naming, but I'm not 100% sure  if `uint32_t ticks` and `wait_ticks` are the same?

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ * @param[in] max_mails
+ *            Maximum number of pending mails in the mailbox.
+ *
+ * @return The allocated mailbox
+ *
+ * @note See #mbox_new to allocate a mailbox on the heap.
+ */
+#define kmbox_new_stack(max_mails) ({ \
+	int sz = sizeof(kmbox_t) + (max_mails)*sizeof(void*); \
+	kmbox_t *mb = alloca(sz); \
```
danbolt, line 404: I just realized now that libdragon has `alloca` and I love it :heart:

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ * @brief Allocate a new mailbox (on the stack).
+ *
+ * @param[in] max_mails
+ *            Maximum number of pending mails in the mailbox.
+ *
+ * @return The allocated mailbox
+ *
+ * @note See #mbox_new to allocate a mailbox on the heap.
+ */
+#define kmbox_new_stack(max_mails) ({ \
```
danbolt, line 402: At a later point, it might be good to have some sample code for `kmbox_new_stack` for new programmers that aren't as familiar with the return value of a C macro. They trip me up sometimes!

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ *     {
+ *        if (mail == &KEVENT_IRQ_SP)
+ *        {
+ *            [...]
+ *        }
+ *        else if (mail == &KEVENT_IRQ_DP)
+ *        {
+ *            [...]
+ *        }
+ *     }
```
danbolt, line 510: Not to speak for every programmer, but I like this a lot. Or, it feels comfortably straightforward if someone's making a game by themselves, but also easy to put into something larger.

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ *            Priority of the thread. Negative values are not allowed.
+ *            Higher number means higher priority.
+ * @param[in] user_entry
+ *            Entry point of the thread. This is the function that will
+ *            be called when the thread begins execution. If this function
+ *            ever returns, the thread is automatically killed.
+ * @param[in] user_data
+ *            Argument that will be passed to the entry point of the thread.
+ *
+ * @return    A reference to the thread. It is not necessary to store
```
danbolt:
```suggestion
 * @return    A pointer to the new thread. It is not necessary to store
```

Just another naming nitpick.

src/kernel.c
```diff
@@ -0,0 +1,640 @@
...
+	// |             |                        |             |
+	// | Stack guard |   Stack                | kthread_t   |
+	// |             |                        |             |
+	//
+	//
+	// For the kthread structure, we will use the final part of
+	// the buffer: since the stack grows downward, this makes sure that
+	// overflowing the stack doesn't corrupt the thread structure, which might
+	// make debugging more difficult and might even cause the kernel to crash.
+	void *thmem = malloc(STACK_GUARD + stack_size + sizeof(kthread_t));
```
danbolt, line 323: I'd be curious to hear your thoughts on returning `NULL` if `malloc` returns `NULL`.

I worry it might make the kernel API more complicated, but I can also see homebrew programmers coming from a wide background with varying familiarity of embedded programming. They might benefit from the function "failing gracefully".

An assertion might be a better middle ground.

src/kernel.c
```diff
@@ -0,0 +1,640 @@
...
+void kernel_close(void)
+{
+	assert(__kernel);
+	assertf(th_cur == &th_main, "kthread_close can only be called from main thread");
+
+	// Kill the idle thread to release the memory
+	kthread_kill(th_idle);
+	th_idle = NULL;
+
+	assertf(th_count == 1, "not all threads were killed");
```
danbolt, line 300: :+1:

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ * The thread is immediately made ready after creation, so if it has
+ * a priority higher than or equal to the current thread, it will be
+ * scheduled immediately, before thread_new returns.
+ *
+ * The thread will begin execution from the specified entry point
+ * function. If the function ever returns, the thread is automatically
+ * killed.
+ *
+ * @param[in] name
+ *            Name of the thread (for debugging purposes)
```
danbolt, line 250: Could `name` be optional? Would it be alright if a programmer pass in `NULL` and the kernel set a default name?

src/kernel.c
```diff
@@ -0,0 +1,640 @@
...
+	kthread_t *th = thmem + STACK_GUARD + stack_size;
+
+	// Initialize the thread structure
+	memset(th, 0, sizeof(kthread_t));
+	th->stack = thmem;
+
+	strlcpy(th->name, name, sizeof(th->name));
+	th->user_entry = user_entry;
+	th->user_data = user_data;
+	th->pri = pri;
```
danbolt, line 334: Since negative values aren't allowed, would it be worthwhile to assert if `pri` is less than zero?

danbolt:
I love this so much! `kernel.h` is a lot easier to read and parse through than libultra's `OSMesgQueue` documentation.

I've tried to look for things that might make the API a little easier to handle for a programmer new to low-level C programming. Stuff like `kthread_new` returning `NULL` if we're out of memory, or optional thread names. The N64Brew Game Jam had a variety of developers with different experience levels, and I'd love to advocate for ease-of-use on their behalf. If this is able to empower someone to make something new for the Nintendo 64, I'm all for it.

include/kernel.h
sp1187: Is it necessary for the name for be a char array at all rather than a pointer to a string? I expect run-time generated names to be pretty rare, and making the thread struct smaller might be better for the cache, especially if the name string is stored in a rodata section further away  and never or rarely accessed.

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
+	/** Size of the stack in bytes */
+	int stack_size;
+	/** Name of thread (for debugging purposes) */
+	char name[17];
+	/** Internal flags */
+	uint8_t flags;
+	/** Priority of the thread (0=lowest, use positive numbers only) */
+	int pri;
```
sp1187: How many priority levels are needed really? Guess this variable could be smaller.

include/kernel.h
```diff
@@ -0,0 +1,531 @@
...
+ */
+typedef struct kevent_s
+{
+	/** Mailboxes attached to this event */
+	kmbox_t *mboxes[MAX_MAILBOXES_PER_EVENT];
+} kevent_t;
+
+
+/** @brief Event generated when a SP interrupt is triggered */
+extern kevent_t KEVENT_IRQ_SP;
```
sp1187, line 196: Would it make sense to add events for non-RCP interrupts such as the timer interrupt (IP7) or the pre-NMI interrupt (IP4)?

include/kernel.h
buu342: Libultra only uses 256, which I think is more than enough. Though I think having it as a short (2 bytes) would be better as it 32-Bit aligns the struct without any padding :smiley:

rasky:
> Within kernal.h, the same concept is referred to as a message box, a mail box and as kmbox. I think a single name should be applied throughout. Libdragon usually spells things out in full, so personally I like `kernel_mailbox_t` etc.

Any suggestion on whether it's better to use "message box" or "mail box" in the first place? I'd go with mailbox just because it's shorter.

> The `k` prefix also seems a bit inconsistent with libdragon's naming, but I'm undecided on whether every function here being called kernel_blah_blah_blah is a better solution.

Yeah that's the reason why I haven't put the full `kernel_` in front of everything. If we don't like the `k`, I guess a reasonable alternative is to drop the kernel dependency altogether and simply have `thread_new`, etc.

> `kmbox_empty` reads a bit ambiguously to me. In my projects these would be is_empty and is_full, though the only places libdragon uses that convention are from my own PRs, so I don't have much to justify myself there.

No strong opinions here. In general I prefer shorter to longer, so I tend to err on the side of dropping `is` from booleans.

include/kernel.h
rasky: Good point. I was in general concerned about stack overlows corrupting `kthread_t` instances to the point of making debugging difficult so an array of chars survives better than a pointer, but I think this concern is now obsolete as I have simply moved the `kthread_t` instead after rather than before the stack. I'll change this to a pointer.

include/kernel.h
rasky: Not sure I understand your comment... `wait_ticks()` is a libdragon function defined in `n64sys.h`. They do use the same clock.

include/kernel.h
rasky: There's an example in the documentation of `kmbox_attach_event`.

include/kernel.h
rasky: Applied.

src/kernel.c
rasky: I don't like returning NULL here because the return error can safely be ignored most of the times. In generale, I don't expect N64 games to gracefully handle dynamic end-of-memory situations, so I will just add an assert here to simplify debugging.

include/kernel.h
rasky: After changing name to be a pointer, it now supports NULL. In debug logs you get `(null)` but that's expected I guess.

include/kernel.h
rasky: Changed to `int8_t`.

include/kernel.h
rasky: Thanks, I've added pre-NMI.

Timer interrupt is more delicate; in general, timers are used by kernels for internal bookkeeping (eg: preemptive multitasking, or thread sleeping while yielding). libdragon already offers a good timer library to register callbacks, so I feel like user-level code in a thread shouldn't mess with timer interrupt. So I'd rather not expose it at this level.

rasky:
> I've tried to look for things that might make the API a little easier to handle for a programmer new to low-level C programming. 

Absolutely! In the debugging library PR, I've started also adding asserts here and there in libdragon in situations that have bitten me in the past and in which debugging was hard (eg: calling `dma_read` with size 0 used to freeze the console, and it took me a couple of debugging days to find the bug in a complex code where the size was dynamically calculated depending on all kind of async events). This is also why I made sure that now we can actually call `assert()` in the first place and see the output on the screen (and the debug spew).

I'm all for helping programmers with debugging features especially in an embedded context where there are many hurdles.

In this PR, as you might have seen, I've added stack smashing protection. In my experience, the most common case of instability brought by threads in an embedded system is programmers overflowing the stacks without realizing them. So this kernel adds some cookies at the end of the stack and every time the stack is switched away, it verifies if the cookies were corrupted. 

I've also played a bit with TLBs; it could be possible to map the stack using TLB so that every single memory access outside the allocated range would immediately trigger an exception. The main issue is that TLBs have only a few fixed sizes (for our purposes: 4K, 16K, 64K), so you would need to map a combination of those to cover the stack (for instance, for a 28K stack, one would need 1 16K TLB + 3 4K TLBs), and in general you would be limited to 4K increments at that point.

What do people think about this? Is it worth it, compared to what is already implemented here?

src/kernel.c
rasky: In theory yes, but I use -1 internally for the idle thread, and I would have to have an internal API to create a thread without this assert. It can be done of course, but I'm not sure it's worth it. It's not that -5 doesn't work, it's just that it's never scheduled because priority is too little. Opinions?

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * #kthread_new returns a pointer to the thread (a pointer
+ * to kthread_t) that can be used to externally manage the
+ * thread like changing its priority or killing it.
+ */
+typedef struct kthread_s
+{
+	/** Pointer to the top of the stack, which contains the thread register state */
+	reg_block_t* stack_state;
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
```
sp1187, line 107: This field might be made smaller as well.

rasky: Yeah but that is a snapshot of the global state (to fake TLS) and the global state is int. Changing it here would require changing the global state. I can easily think that 255 nesting levels ought to be enough, but I'd rather not mess with it in the context of this PR.

src/kernel.c
danbolt: Interesting! If that's the case, things are probably fine the way they are. The documentation in `kernel.h` lets the programmer know what's up, and it's not like they'll get a difficult or confusing debugging situation with something like -5. 

Having two initialization functions for threads sounds a little more complicated!

include/kernel.h
danbolt: Ahhh, that makes more sense. This is good; I was a little confused.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ *     an event connected to an interrupt is fired, and a higher priority
+ *     thread waiting for the event is woken up.
+ * 
+ *   * Cooperatively: when a thread explicitly calls a blocking function
+ *     like #kthread_sleep or #kthread_wait_event to wait for an event.
+ *     The kernel will put the current thread to sleep and schedule the
+ *     highest-priority thread among the ready ones, including threads
+ *     that have a priority lower than the one that goes to sleep. When
+ *     the event is fired, the lower-priority thread will be preempted
+ *     to give back control to the thread that just woke up.
```
anacierdem: So we always give control to a just woken up thread regardless of its priority right? What should happen when there are more than one thread need to wake up to the same event (say A and B)? From what I understand from the general layout, the one with the highest priority should wake up (let's say thread A). Then thread B would not get scheduled if there are higher priority threads that will get scheduled if thread A cooperatively yileds. This might not be a practical case but I'm speculating to understand/predict the edge cases.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * entry point function), it will be cleaned up and the memory
+ * released.
+ *
+ * #kthread_new returns a pointer to the thread (a pointer
+ * to kthread_t) that can be used to externally manage the
+ * thread like changing its priority or killing it.
+ */
+typedef struct kthread_s
+{
+	/** Pointer to the top of the stack, which contains the thread register state */
```
anacierdem, line 104: Maybe we should not call this stack to reduce confusion

include/kernel.h
rasky: > So we always give control to a just woken up thread regardless of its priority right?

I think the documentation might be confusing.

Let's says you have three ready threads A (priority 5), B (priority 3), C (priority 3). A is running by definition (since a higher-priority ready thread is always scheduled, winning against lower priority threads). If A explicitly goes to sleep (with either kthread_sleep or kthread_wait_event), at that point the scheduler will schedule a lower priority thread, so either B or C. Notice that if A goes to sleep, *by definition* a lower-or-equal priority thread is scheduled; it's impossible that a higher priority thread is scheduled because if there was a higher priority thread D ready while A was running, A would have been stopped immediately to run D.

Let me know if now it's more clear and how can we make the documentation more clear (though keeping it concise of course).

> What should happen when there are more than one thread need to wake up to the same event (say A and B)? From what I understand from the general layout, the one with the highest priority should wake up (let's say thread A).

All threads are waken up, they all become ready (both A and B). Then the scheduler gives CPU to the currently ready thread that has the highest priority. This could be A like your in example or maybe it could be C (the thread that triggered the event) that might have a higher priority than both A and B. In this case, C keeps running after triggering the event, but A and B are both ready; as soon as C goes to sleep for any reason, they might be scheduled.

include/kernel.h
anacierdem: So should we understand that scheduling a thread which has its interrupts enabled will enable the interrups? I haven't checked the implementation yet on purpose and this made me also think we may need to keep track of all global state per thread e.g `interrupt_disabled_tick`?

include/kernel.h
rasky: I was thinking about this name a few days ago. This variable is the one that basically contains the thread SP register *while the thread is sleeping*. It gets written when the scheduler puts the thread to sleep, and then it's read to restore the SP register when the thread is scheduled. Its type is `reg_block_t*` because we know how to the top of the stack looks like when the interrupt is triggered (just like when an exception is triggered).

Do you think `save_sp` would be better?

Also: for reducing future bugs, we might want to write NULL in this variable when the thread is scheduled. In fact, while the thread is running, this points to some point in the stack that contains nobody knows what (it used to contain the exception stack during exception, but now it contains random data).

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+	reg_block_t* stack_state;
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
+	/** Size of the stack in bytes */
+	int stack_size;
+	/** Name of thread (for debugging purposes) */
+	const char *name;
+	/** Internal flags */
```
anacierdem, line 112: Are these for internal use only? Should we prefix with `__`?

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+	uint8_t flags;
+	/** Priority of the thread (0=lowest, use positive numbers only) */
```
anacierdem, line 114: As it is an `int8_t` now, I think we can remove positive numbers warning.

include/kernel.h
rasky: > So should we understand that scheduling a thread which has its interrupts enabled will enable the interrupts?

Well, in general, this is not a user-visible behavior. Most code runs with interrupts enabled. If the user disable interrupts, it's usually to run some short critical code that cannot be interrupted, and in fact with interrupts disabled the kernel scheduler can not run! So disabling interrupts also has the effect of disabled thread preemption, which is a good default of existing code (in modern kernels, disabling interrupts and disabling preemptions are two different concepts, but in our simple case I think it's OK to conflate the two). So from an user standpoint, when a thread is scheduled, interrupts are always enabled.

In the kernel implementation itself, though we need to force thread switching while interrupts are disabled. This is just an internal implementation detail, but when we do, we absolutely need to restore enabled interrupts for scheduled threads (and then restore disabled interrupts when we get back to the switching point!).

> we may need to keep track of all global state per thread e.g `interrupt_disabled_tick`

In the specific case of `interrupt_disabled_tick`, that is OK to keep as a global state as `timer_ticks` uses it just as a "new enough" point in time. But in general yes, we might of course find out in future that there is some global state that should actually be thread-local storage. I would say this is unlikely when dealing with hardware libraries that deal with hardware peripherals (as the peripheral is just one and has just one state, irrespective of the running thread), but it's not impossible. If we find out other thread-local storage, we would have to deal with it. We can't stuff everything in kthread_t, but for now it's a fast, simple way.

include/kernel.h
rasky: To tell you the truth, the whole `kthread_t` structure should be opaque from a formal API standpoint. User code should just manipulate "handles" to threads (so opaque pointers), and surely not mess with the thread structure contents. 

I resisted the temptation of hiding this in an internal header because in general I think that with libdragon we're generally better off keeping the door open to advanced hacking / tweaking of libraries, rather than trying to force a pure API that the users then need to hack around when it doesn't do everything possible. But I'm happy to be talked otherwise for the kernel. See also `timer_link_t` which is in `timer.h` but I fail to see good reasons for it, if not for a general "open" philosophy.

include/kernel.h
anacierdem: Maybe we can state the rules of the threading system somewhere and verify the documentation makes sense/is consistent considering the rules. Most of this is already stated in the docs, but having them like this might help;

- When a "switch" occurs, the library switches to the thread with the highest priority from the `ready` pool, if any.
- A "switch" accurs when;
  - the active thread go to sleep yielding via a `kthread_` call in its entry point function or;
  - an interrupt occurs (which may put another thread into the ready pool)

I'm not sure if this is complete yet but would help a lot understanding it, at least for me.

Then it is the consumer's responsibility to make sure a thread will eventually get shceduled etc.

include/kernel.h
anacierdem: So the only thing we care is SP and we keep the struct to help with the offset from what I understand from your explanation. If that's the case, we may change it to a `uint64_t` and actually write the contents via `offsetof` etc.

Or if that's not something practical, what about `last_interrupt_frame` maybe?

include/kernel.h
anacierdem: I agree on ideally we should keep it internal. I'm not against keeping it open such that consumers can implement their own methods around them if need be. On the other hand, with an implementation so tightly coupled with the libdragon and its ex. vectors etc., I don't think it will be of much help.

We can still open the core helpers/mechanisms to the user (like we did for example with cop0/cop1 macros and custom ex handler registration etc.) to enable tweaking. As an example, RDP implementation can consist of many small pieces of helper funcs. / macros instead of a fully closed api like `rdp_enable_primitive_fill`. Rather it can implement the ring buffer and provide an API to enqueue commands. Then we could have a `enable_primitive_fill` "helper" instead, that we can enqueue.

A similar approach would not probably be easy at the exception/kernel level though. Instead we may try adding a "hook" where we actually schedule the threads, such that third party code can implements its own "kernel extensions" but that would potentially bring the performance down and would be a mess in practice. Instead, one should provide her/his own `inthandler` and kernel modules if s/he wants to tweak it. I think we should aim for this kind of an architecture for long term expandibility/tweaking. This may not be easy considering how little time we can spare for this project (except for @rasky :)) though.

In short, I think it is ok to keep it open but I see little value doing so for such a highly tightly coupled portion of the library.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * this function should be called only when all created threads
+ * are exited or have been killed.
+ */
+void kernel_close(void);
+
+/**
+ * @brief Create a new thread
+ *
+ * Create a new thread, with a specified stack size and priority.
+ * The thread is immediately made ready after creation, so if it has
```
anacierdem, line 241: Would there be any advantages of having the ability to mark it as not ready yet?

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+typedef struct kthread_s
+{
+	/** Pointer to the top of the stack, which contains the thread register state */
+	reg_block_t* stack_state;
+	/** Current state of interrupt depth for this thread */
+	int interrupt_depth;
+	/** Size of the stack in bytes */
+	int stack_size;
+	/** Name of thread (for debugging purposes) */
+	const char *name;
```
anacierdem, line 111: What about keeping an id instead? I have seen in the examples that we even used userdata for that. Why not even remove this alltogether and provide a common userdata struct as a helper? This way we can kep this as minimal as possible and make it more extensible.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ *            be called when the thread begins execution. If this function
+ *            ever returns, the thread is automatically killed.
+ * @param[in] user_data
+ *            Argument that will be passed to the entry point of the thread.
+ *
+ * @return    A pointer to the new thread. It is not necessary to store
+ *            this reference if not required; the thread will clean up
+ *            after itself when it exits.
+ */
+kthread_t* kthread_new(const char *name, int stack_size, int8_t pri,
```
anacierdem, line 268: I'd vote for removing the name from here. I think we can make this more generic without a name. Consider;

    typedef struct kudata_s
    {
	    uint8_t id;
            const char *name;
            // anything I want as a consumer
    } kudata_t;
    
    kudata_t thread_data_1 = { .id = 0, .name="thread1" };
    kudata_t thread_data_2 = { .id = 0, .name="thread2" };
    
    kthread_new(2048, 2, thread_entry, &thread_data_1);
    kthread_new(2048, 2, thread_entry, &thread_data_2);

If we need the name as a lower-level debugging tool, we can enable it behind a define IMO.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * @brief Yield execution of the current thread and run the scheduler.
+ *
+ * This function allows the current thread to cooperatively yield
+ * its execution to allow other threads to run.
+ *
+ * The scheduler will switch to the highest priority thread that is
+ * currently ready to run. If no ready thread has a priority higher than
+ * the current thread, the scheduler will switch to a different thread 
+ * of the same priority of the current one (the scheduler will guarantee
+ * a correct round-robin scheduling among threads of the same priority).
```
anacierdem, line 282: What about the case where the thread creates a new thread of the same priority? Will it get scheduled after all of them in case all the remaining threads are of same priority?

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * If no ready thread has a priority higher or equal than the current
+ * one, the scheduler will reschedule the current thread, that will
+ * continue execution.
+ *
+ * @note The scheduler is semi-preemptive. Any interrupt could cause
```
anacierdem, line 287: Ok this is useful information, adding it to the rules list above.

include/kernel.h
```diff
@@ -0,0 +1,534 @@
...
+ * Notice that #wait_ticks and #wait_ms are not threading aware, so they will
+ * just spin-wait while keeping the current thread ready. Other higher-priority
+ * threads might still be scheduled if they are awaken by an interrupt, but in
+ * general thread_sleep is a better primitive if you want the current thread
+ * to yield.
+ *
+ * @note This function requires the timer module, so #timer_init must have
+ *       been called.
+ *
+ * @param[in] ticks
```
anacierdem, line 314: This won't guarantee a re-scheduling after the time is elapsed though right? o/w there should be another rule for switching threads, other than priority.
We would probably need to change priority if we want that behavior.

include/kernel.h
rasky: > So the only thing we care is SP and we keep the struct to help with the offset from what I understand from your explanation. If that's the case, we may change it to a uint64_t and actually write the contents via offsetof etc.

Uhm I'm not sure I follow. A `thread_t` is a thread descriptor. For a sleeping thread, we need to keep its stack pointer saved in the descriptor, otherwise we have no way to schedule it again. We could keep the stack pointer as `uint64_t` since it's a R4300 register after all, but we do access the contents of the stack while scheduling, since we know how the top of the stack looks like (it's a `regblock_t*`). 

Notice that we also have a `void* stack` variable in `thread_t` which is the bottom of the stack (the beginning of the memory area). Maybe we could have `void* stack_start` and `regblock_t *stack_cur`?

include/kernel.h
rasky: OK I agree with you. Let's keep it this way.

include/kernel.h
rasky: The first iteration of the API was actually creating "paused" threads that the user had to unpause. I found it a little overwhelming and without any specific benefit. I think it's fine to create them ready most of the time, since you can always create a thread with a very low priority (so that it's not scheduled even if it's ready) and then later bump it up.

include/kernel.h
rasky: I think that `name` is very useful for debugging. For instance, if the thread overflows its stack, you get a nice error screen saying that thread "foo" has overflown its stack of size N by at least X bytes. That is incredibly useful. 

At some point in the future, we might get the exception screen to also show the thread in which the exception was thrown, and we could even let the exception kill the thread but let the other threads continue running (libultra does this IIRC). 

I actually fail to see what we gain by removing the thread name.

Also I'm not sure about how this is related to `userdata`. `userdata` is just a generic argument to the user entry point, and it's actually a pretty common pattern in C for doing generic functors. This allows the caller to reuse the same function for multiple threads, passing some "context" to it (eg: the audio channel to play, or something).

include/kernel.h
rasky: This kind of construct requires the user to define a structure to be able to pass custom data to the thread entry point. Let's say I have a audio entry point which is `play_sound_effect` and I have `sound_t` as descriptor for a sound. Right now, I can do something like `kthread_new("audio", 4096, PRI_AUDIO, play_sound_effect, (void*)my_sound);`. With your proposal  I need to define my own `kudata_sound_effect_t` structure which inherits `kudata_t` and contains an additional field of a type `sound_t*` and then pass it `kthread_new`. It looks too much verbose to me.

include/kernel.h
rasky: This is answered in `kthread_new` documentation:

> * The thread is immediately made ready after creation, so if it has
> * a priority higher than or equal to the current thread, it will be
> * scheduled immediately, before thread_new returns.

include/kernel.h
anacierdem: I haven't read the implementation yet and if it is integrated in such a way, it seems totally ok to have a name. I thought it was just a name the current thread can query for debugging on the consumer side, then it becomes equivalent to userdata.

include/kernel.h
rasky: > This won't guarantee a re-scheduling after the time is elapsed though right? 

It guarantees that the thread becomes ready again after the time is elapsed. Whether it is scheduled or not at that point depends on what are threads are ready. Consider also that even if it's scheduled again at that point, it might be switched off one microsecond later once a higher priority thread becomes ready. So in general `sleep` in a threading context means `sleep at least`, given that any thread can be interrupted by any other higher priority thread at any point.

include/kernel.h
anacierdem: If we are accessing other things from the top of the stack, we better have the regblock sruct, it'll be clearer for me when I delve more into the implementation probably. Let me think of a name after fully understanding it 🙂

include/kernel.h
anacierdem: Agreed. Knowing `name` is also integrated with the current implementation, this is just fine. I originally thought it was for user code debugging and was superficial. Maybe we can note the name is for integrated debugging extensions.

include/kernel.h
rasky: I've updated the documentation. Hopefully is more clear now. Please let me know.

rasky:
> > Within kernal.h, the same concept is referred to as a message box, a mail box and as kmbox. I think a single name should be applied throughout. Libdragon usually spells things out in full, so personally I like `kernel_mailbox_t` etc.
> 
> Any suggestion on whether it's better to use "message box" or "mail box" in the first place? I'd go with mailbox just because it's shorter.

I removed all references to "message box", it's now called "mailbox" everywhere. The type name is still `kmbox_t` but I can be talked into `kmailbox_t` if people feel strongly.

include/kernel.h
anacierdem: The rules above makes is really clear now.

anacierdem:
I am towards removing the k/kernel prefix too, but wont be something I will insist on. TLB idea looks really good, though I think we can implement it later on.

include/kernel.h
```diff
@@ -0,0 +1,529 @@
...
+ *     kmbox_detach_event(mbox, &KEVENT_IRQ_SP);
+ *     kmbox_detach_event(mbox, &KEVENT_IRQ_DP);
+ *
+ *
+ * @note It is mandatory to detach all attached events before
+ *       destroying a mailbox.
+ *
+ * @note If the mailbox happens to be full when the event is triggered,
+ *       the event will be missed. When you attach a mailbox to an event,
+ *       make always sure to have enough space in the mailbox.
```
anacierdem, line 522: Should we then disable the interrupts and query it with `kmbox_full` ?

src/inthandler.S
```diff
@@ -11,134 +11,163 @@
...
-	/*Save $2 before using*/
-	sd $2,save02
-
 	.set noat
-	/*Fetch exception pc off cop#0*/
-	mfc0 $2,C0_EPC
-
-	la k1,save01
-	sd $1,(k1)
+	.set noreorder
```
anacierdem: Without a matching `.set reorder`, we may be disabling them in unexpected ways.

src/inthandler.S
```diff
@@ -11,134 +11,163 @@
...
-	mfc0 k1,C0_SR
-	la $1,saveSR
-	sw k1,($1)
-	la $1, ~1
-	and k1,$1
-	mtc0 k1,C0_SR
-	.set at
+#define EXCEPTION_CODE_SYS_CALL      (8<<2)
+#define EXCEPTION_CODE_BREAKPOINT    (9<<2)
+#define EXC_STACK_SIZE (544+32)
```
anacierdem: A comment on the size would be nice IMHO

src/inthandler.S
rasky: Isn't this per-file?

src/inthandler.S
rasky: Done. I've also added a static assert that `sizeof(reg_block_t) == 544`.

include/kernel.h
rasky: I think it's up to the receiver thread to create a mailbox which is large enough to buffer pending events that might be received. In general, I would say that most threads will wait only on 1 event (1 type of interrupt), and then process it as soon as it happens and before it happens again.

On the sender side, we're already under interrupt. So if a AI interrupt happens, the code will notify all pending threads that were waiting for an AI interrupt. But if one of those mailboxes happen to be full, there's nothing we can do in the context of the AI interrupt.

src/inthandler.S
anacierdem: I'm not 100% sure on that. I suspect they might get appended to each other in the end.

rasky: Each assembly file generates its own object file. Disassembling inthandler.o I don't see any special symbols being emitted related to the state of `.set noreorder` so I would say that this is local state of the assembler that is not persisted across different files.

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+ *
+ * This macro is very low level and is called as part of higher-level primitives
+ * that force a context switch like #thread_yield.
+ *
+ * Do not call this under interrupt; use #KTHREAD_SWITCH_ISR instead.
+ */
+#define KTHREAD_SWITCH()      __asm("syscall 0x1")
+
+/**
+ * @brief Executed a context switch, during an ongoing interrupt service routing.
```
anacierdem: nit: routine?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+ * This forces a context switch to happen just like #KTHREAD_SWITCH, but it is
+ * the right version to call if the code is currently servicing an interrupt.
+ */
+#define KTHREAD_SWITCH_ISR()  ({ __isr_force_schedule = true; })
+
+/** @brief Main thread */
+kthread_t th_main;
+/** @brief Pointer to the current thread */
+kthread_t *th_cur;
+/** @brief Pounter to the idle thread */
```
anacierdem: nit: pointer

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+ */
+#define KTHREAD_SWITCH()      __asm("syscall 0x1")
+
+/**
+ * @brief Executed a context switch, during an ongoing interrupt service routing.
+ *
+ * This forces a context switch to happen just like #KTHREAD_SWITCH, but it is
+ * the right version to call if the code is currently servicing an interrupt.
+ */
+#define KTHREAD_SWITCH_ISR()  ({ __isr_force_schedule = true; })
```
anacierdem, line 40: This seemed like complicating things a little to me. Why can't we just syscall again?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+
+	uint64_t *s = (uint64_t*)th_main.stack;
+	for (int i=0;i<STACK_GUARD/8;i++)
+		s[i] = STACK_COOKIE;
+
+	// The main thread is the currently scheduled one.
+	th_cur = &th_main;
+
+	// Allocate the idle thread
+	th_idle = kthread_new("idle", 4096, -1, __kthread_idle, 0);
```
anacierdem, line 285: This will have the highest priority as it is unsigned now?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	if (DEBUG_KERNEL) debugf("thread end: %s[%p]\n", th_cur->name, th_cur);
+	kthread_kill(NULL);
+}
+
+void __kthread_check_overflow(kthread_t *th)
+{
+	// If the current stack pointer is beyond the end of the stack,
+	// this is a stack overflow even without checking the integrity of the
+	// guard.
+	assertf((void*)th->stack_state >= th->stack + STACK_GUARD,
```
depp: Cast to `uintptr_t` or `intptr_t` instead of `void*`. C compilers can optimize out bounds checks if you write them using pointer arithmetic, because out-of-bounds pointers aren’t supposed to ever exist (UB).

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	th->pri = pri;
+
+	// Initialize the stack guard
+	uint64_t *s = (uint64_t*)th->stack;
+	for (int i=0;i<STACK_GUARD/8;i++)
+		s[i] = STACK_COOKIE;
+
+	// Top of the stack is the end of the stack area, so where the kthread_t
+	// structure is. Remember that the stack grows downward.
+	void *top_stack = (void*)th;
```
depp: Cast is unnecessary.

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	// SR: start in exception mode (we will be within an exception during scheduling),
+	// and with interrupts enabled.
+	th->stack_state->sr = C0_STATUS() | C0_STATUS_EXL | C0_STATUS_IE;
+
+	// GP: this is needed by the compiler to point to the global memory pool,
+	// so let's initialize to the same value it has in this thread.
+	th->stack_state->gpr[28] = REG_GP();
+
+	// SP: top of stack
+	th->stack_state->gpr[29] = (int64_t)(int32_t)top_stack;
```
depp, line 366: Probably should be `= (intptr_t)top_stack;`

depp (commented):
It looks like a thread can be in one of four states: running, runnable, waiting, and zombie. It might be easier to understand if these were enumerated:

```
#define TH_STATE_RUNNING 1
#define TH_STATE_RUNNABLE 2
...
```

I think this would be easier to understand than using the ZOMBIE / INLIST flags.

E.g. Any runnable thread can be woken up by `kthread_yield()`. Any waiting thread is blocked on a timer or resource, will be made runnable when the timer fires / resource becomes available.

src/inthandler.S
```diff
@@ -11,134 +11,175 @@
...
+	move sp, k0
+
+	andi t0, k1, 0xff
+	beqz t0, justaninterrupt
+	nop
+	bne t0, EXCEPTION_CODE_SYS_CALL, critical_exception
+	nop
+
+thread_schedule:
+	addiu a0, sp, 32
```
anacierdem: We may reference the above comment for `32` here as well

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+ * @return The stack pointer containing the register block of the thread to
+ *         be activated.
+ **/
+reg_block_t* __kthread_syscall_schedule(reg_block_t *stack_state)
+{
+	if (th_cur)
+	{
+		// For debugging purposes: check if the current thread has overflown
+		// its allocated stack.
+		th_cur->stack_state = stack_state;
```
anacierdem: This seems to be the only place where we update this pointer, maybe we should have a warning to notify we should not use it somewhere else? Then do we really need it on the thread struct I'm not sure, maybe we can keep this as an argument only.

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+			// syscall forcing a thread switch (THREAD_SWITCH), or an interrupt
+			// that awakes a higher priority thread (THREAD_SWITCH_ISR).
+			if (C0_GET_CAUSE_EXC_CODE(stack_state->cr) == EXCEPTION_CODE_SYS_CALL)
+			{
+				// Skip syscall instruction when this thread will be scheduled
+				// again. Notice that, given that this is an explicit syscall,
+				// the thread must already have been added to a waiting list,
+				// otherwise it wouldn't ever be scheduled again.
+				assert(th_cur->flags & TH_FLAG_INLIST);
+				assert(*(uint32_t*)th_cur->stack_state->epc == 0x4C);
```
anacierdem: I didn't get this one

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	uint64_t *s = (uint64_t*)th->stack;
+	for (int i=0;i<STACK_GUARD/8;i++)
+		s[i] = STACK_COOKIE;
+
+	// Top of the stack is the end of the stack area, so where the kthread_t
+	// structure is. Remember that the stack grows downward.
+	void *top_stack = (void*)th;
+
+	// Put the initial regblock at the top of the stack. This is the
+	// block that will be popped when the thread is first scheduled, and needs to
```
anacierdem: This is where the int handler restore values from right? Maybe `the thread is first scheduled` is not the best way to explain.

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	// initialize all the registers and jump to the entry point.
+	th->stack_state = (reg_block_t*)(top_stack - sizeof(reg_block_t));
+	assert(((uint32_t)th->stack_state & 7) == 0);
+	memset(th->stack_state, 0, sizeof(reg_block_t));
+
+	// EPC: __thread_boot function (which will then invoke the user entry point)
+	th->stack_state->epc = (uint32_t)&__kthread_boot;
+
+	// SR: start in exception mode (we will be within an exception during scheduling),
+	// and with interrupts enabled.
```
anacierdem, line 358: Don't we [disable](https://github.com/DragonMinded/libdragon/pull/151/files/d6217842197c60c0d49cdf01d21d4b7cf1bdc1dc#diff-614949a9360542b842db8214ef209cb3f02ef100c151b77932bd9d1a80aba7c6R123) them at the moment?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	th->stack_state->sr = C0_STATUS() | C0_STATUS_EXL | C0_STATUS_IE;
+
+	// GP: this is needed by the compiler to point to the global memory pool,
+	// so let's initialize to the same value it has in this thread.
+	th->stack_state->gpr[28] = REG_GP();
```
anacierdem, line 363: Does this ever change? If not we may always restore in the inthandler to save a few cycles?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	if (DEBUG_KERNEL) debugf("[kernel] sleeping %ld %s[%p]\n", ticks, th->name, th);
+	disable_interrupts();
+
+	// Start a timer for the specified delay.
+	timer_link_t timer;
+	start_timer(&timer, ticks, TF_ONE_SHOT, cb);
+
+	// Mark the thread as being in a list to avoid the scheduler to panic.
+	// This is not "technically true", but it's "semantically true": when
+	// we say "a thread is in a list" it means that it's registered somewhere
```
anacierdem, line 462: Minor nit: maybe we can call it `TH_FLAG_OWNED`, what do you think?

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	kthread_yield();
+}
+
+void kthread_sleep(uint32_t ticks)
+{
+	kthread_t *th = th_cur;
+
+	// Timer callback. This will be invoked when the timer elapses after the
+	// requested delay.
+	void cb(int ovlf)
```
anacierdem, line 443: Stack trampolines are pretty useful here!

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	// Mark the thread as being in a list to avoid the scheduler to panic.
+	// This is not "technically true", but it's "semantically true": when
+	// we say "a thread is in a list" it means that it's registered somewhere
+	// so that there's a way to wake it up in the future. In this case,
+	// the timer callback will do it, so we can reassure the kernel scheduler
+	// that everything is fine.
+	th->flags |= TH_FLAG_INLIST;
+
+	// Context switch
+	KTHREAD_SWITCH();
```
anacierdem, line 469: How do we prevent re-scheduling ourselves here? e.g when we are the highest priority thread and we were in the ready list already? I think I'm missing something here.

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	return mbox->mails[mbox->w] != NULL;
+}
+
+/**
+ * @brief  Enqueue a mail into a mailbox (lower-level primitive).
+ *
+ * @note   This function must be called with interrupts disabled, as it
+ *         manipulates the ready list.
+ *
+ * @return 0 if the mail was not sent (mbox full)
```
anacierdem, line 517: Still maybe nice to enumarate these even if they are internal

src/kernel.c
```diff
@@ -0,0 +1,642 @@
...
+	return ret != 0;
+}
+
+void kmbox_send(kmbox_t *mbox, void *mail)
+{
+	disable_interrupts();
+	while (!kmbox_try_send(mbox, mail))
+	{
+		__thlist_add(&mbox->wait_list, th_cur);
+		KTHREAD_SWITCH();
```
anacierdem, line 557: What would happen if the cuurent thread is the highest priority thread and no threads in the wait list? Will it get re-scheduled over and over as there are no other threads that can get scheduled to free up space? I think this corresponds to `test_kernel_mbox_2` doing a fourth initial `kmbox_send(mbox, &msg4);`. I'll try this for myself :)

include/kernel.h
anacierdem: The only name coming to my mind is `regs_when_scheduled_wrt_sp` :) If we were also including the 32 bytes I may call it just `sp_when_scheduled` but not being exactly sp and a `regblock_t` could not come up with a better alternative. What if we change `__kthread_syscall_schedule` paramter to be just the sp pointer without the offset, name it `sp_when_scheduled` and extract the `regblock_t` from it in the function? Might be a little more descriptive IMO but this is just a minor detail, I wouldn't care at all if it was not in the public API.

tests/test_kernel.c
```diff
@@ -0,0 +1,444 @@
...
+
+	kthread_t *rth2 = kthread_new("read_kthread_2", 2048, 4, read_kthread_2, NULL);
+	DEFER(kthread_kill(rth2));
+
+	// Go to sleep. Now the RSP will run, trigger the two reader threads, and
+	// they will register themselves in the called array.
+	kthread_sleep(TICKS_FROM_MS(5));
+
+	// Make sure the order is correct (first the higher priority thread).
+	uint8_t exp1[] = { 2, 1, 0 };
```
anacierdem, line 431: I expected `{ 2, 0 }` because of;
> Once a thread receives a mail from a mailbox, the mail is removed and other threads will not see it anymore

anacierdem (approved):
I went over the implementation as well and it looks really good! I have learned a lot from this.
The mail box implementation was a little complicated to grasp but I'm pretty sure I cannot easily come up with something better :)
I think the API is great and the implementation seems tweekable later on if we ever need (e.g stack guarding with TLB etc.) On the other hand considering other comments in the PR I still feel like exposing the thread structure might not be a great idea. e.g think about removing the flags and the structure will not be the same anymore. Still, I'm ok to keep it open only with a remark stating it is internal and open to possible future changes.

include/kernel.h
```diff
@@ -0,0 +1,529 @@
...
+/**
+ * @brief Wait for a single hardware event.
+ *
+ * This functions blocks the current thread until a specified
+ * hardware event happens. The CPU is yielded so that other
+ * threads will be scheduled.
+ *
+ * If the thread needs to handle multiple events and/or a more
+ * elaborate communication with other threads, you can use
+ * #thread_listen_event instead.
```
anacierdem: There is no `thread_listen_event` should use kmbox instead

anacierdem:
Another idea: what about exposing a define to prevent the kernel checks? In case someone does not use it or provides their own alternative implementation.

anacierdem:
I haven't verified this yet (may need a separate test) but I suspect that a stack callback in a thread may not be able to work properly. Think about this code sample;
```
void func_th(void *arg) {
	volatile bool cb1_called = false;

	void cb1(int ovlf) {
		cb1_called = true;
	}

	timer_init();

	timer_link_t *t1 = new_timer(TICKS_FROM_MS(2), TF_ONE_SHOT, cb1);

        kthread_set_pri(NULL, 2);

	while (!cb1_called) {}

        // I think this will never be able to continue execution
}

int main() {
	kernel_init();

	kthread_set_pri(NULL, 5);
	kthread_new("test1", 2048, 10, func_th, NULL);

        while(1){}
}
```
- `func_th` will get immediately scheduled and register a timer callback yielding itself
- main thread starts the idle loop, we are in the main context now
- timer expires, stores registers in the main stack
- then it calls `cb1` before we schedule the secondary `func_th` thread
- the callback potentially clobbers a main thread register/memory (register are safe though)
- `func_th` is scheduled back, `cb1_called` was not set and it hangs.

I may be totally off with this but seems like it needs a test of its own. If this is not working as expected, we need to declare global functions incompatible as threads.

anacierdem:
GNU stack function extension seems to be smart enough to keep sp and arrange the trampoline for any parameters. Still might be a good idea to add tests at some point.

src/kernel.c
rasky: We could in theory, but it would mean we could suspend a thread while it's under interrupt. I'm not sure of the ramification of this. I like the idea that when an interrupt fires, it is being served right away; the fact that somewhere in the interrupt service routine, somebody might switch away from it and the service routine might be suspended for an indefinite period of time seems to break the basic assumption that interrupts are super high priority, even higher than the highest thread. It looks dangerous.

Also, there's a small performance/memory benefit to get, as `KTHREAD_SWITCH_ISR` doesn't cause a second exception but causes the current exception to be used for thread switching as soon as the interrupt is done with. Nothing incredible, but still sounds like a easy gain. 

In general, this might make thing a little bit more complicated in there kernel itself, but it should be mostly transparent to users.

For comparison, FreeRTOS for instance has many duplicated APIs in this same way (normal and ISR version). Nintendo kernel got away with the problem by not allowing to write interrupt service routines in the first place (interrupts are just used to switch threads and fire events, business logic is only in threads); it's maybe a easier design but surely less efficients. I like how we are able to do both with this PR (eg: keeping the AI fed via ISR seems like a good design, irrespective of the fact that you might have a audio thread to handle playback).

src/kernel.c
rasky: It's not unsigned, it's a `int8_t`.

src/kernel.c
rasky: The comment was wrongly placed (fixed), but we do need to save it. It's our SP pointer, that is the only place where we remember where our thread stack is in memory.

src/kernel.c
rasky: Thread switching happens in two ways: explicit syscall (`KTHREAD_SWITCH()`) or an interrupt fired and somehow the service routine woke up a higher priority thread, that is then activated (`KTHREAD_SWITCH_ISR()`).

This if/else is basically handling the two cases differently. In the first case (the quoted one), it asserts that the thread has been put in a waiting list. If it had not, it is a programming bug in the kernel code; it means somebody is calling `KTHREAD_SWITCH` without first putting the thread in a waiting list, which means that the thread would be "leaked": nobody would kept its pointer around and there would be no way to scheduled it again.

Does this make more sense to you? If so, how do you suggest to improve the comment?

src/kernel.c
rasky: Tried to rephrase, let me know if it reads better.

src/kernel.c
anacierdem: Actually I didnt get the second assertion with `0x4C` o/w Im ok with the comment

src/kernel.c
rasky: This is the value that will be restored [when the ISR exits](https://github.com/DragonMinded/libdragon/commit/d8c15022088b980fd1de017869d1a2371d0a04da#diff-614949a9360542b842db8214ef209cb3f02ef100c151b77932bd9d1a80aba7c6R239)

src/kernel.c
rasky: It is part of the C ABI. It never changes while C code is running, but in theory there could always be assembly code that temporarily use it otherwise, and then restore it before going back to C. I think it's a rare chance. If this happens BTW even the current code is wrong: the ISR should put the "standard" value in GP (`_gp`, as initialized by `entrypoint.S`) before calling the ISR C routines, otherwise the C code would be called with a wrong GP code. I've fixed this now by adding a `la gp, _gp`.

If we want to go to your optimized route, then we would have to document that nobody can ever touch GP, not even in assembly code. That seems something easy to forget about, and at the same time easy to cause nasty bugs. I am not sure it's worth it.

src/kernel.c
anacierdem: This totally make sense! Also I don't think it is something we cannot change if need be as it is completely internal.

src/kernel.c
rasky: I would like to attempt @depp's suggestion to use explicit thread states, first.

src/kernel.c
anacierdem: I don't know what you have changed so I need to go over it again sometime.

src/kernel.c
rasky: We haven't put the current thread in the ready list, so the scheduler can't schedule it right away. The thread is instead stored in the stack trampoline, so the scheduler doesn't see it. When the timer fires, the callback will put the thread pointer back in the ready list, and at that point the scheduler will be free to schedule it.

src/kernel.c
rasky: I don't get this... they are already enumerated, aren't they?

src/kernel.c
anacierdem: Yeah that's my bad, forget about it when I couldn't find any reference :)

src/kernel.c
rasky: Yes. If the highest priority thread tries to write to a full mbox, it would keep spinning there. The only way to get out is that an interrupt fires and wakes up a thread that was waiting on the interrupt (kevent) and that thread is even higher priority so it gets scheduled, and it pulls from the mbox.

We might in theory detect this looping scenario, but I'm not sure what we would do in that case. Even if the above description sounds far fetched, it's not *that* weird. Think a main thread enqueuing RSP jobs for a RSP thread. The main thread pushes many jobs and at some point fills the mbox. The RSP threads pop them one at a time, schedule them in the RSP, and then wait execution (wait for RSP interrupt). I'm not sure there is nothing inherently wrong with this. And if the mbox is full, the main thread hasn't really anything else better to do than (spin)waiting.

tests/test_kernel.c
rasky: there are two mailbox here `m1` and `m2`, and both of them are attached to the same event. Events are broadcast, so all mboxes attached to an event receive it. This is by design to allow multiple threads to wait for the same interrupt.

src/kernel.c
anacierdem: GCC does not allow changing gp via inline assembly. I'm not sure on individually compiled files but shouldn't such a code already restore it anyways? Nothing important though, was just an idea.

rasky:
> Another idea: what about exposing a define to prevent the kernel checks? In case someone does not use it or provides their own alternative implementation.

Done.

src/kernel.c
anacierdem: I meant keeping the values in macros/enums and using them

src/kernel.c
rasky: `gpr[29]` is a `int64_t`, while `top_stack` is a 32-bit pointer. I need to convert it to `int64_t` in the same way MIPS assembly treats 32-bit values in 64-bit registers, which is by sign extending them. I don't want to rely on any property of integer promotion to casually match what the MIPS architecture does.

src/kernel.c
rasky: Ah sorry. 0x0000004C is the opcode for SYSCALL 0x1. I'm just checking that EPC is pointing to the expected syscall instruction, as a consistency check. I'll add a comment

src/kernel.c
rasky: I've pushed the change separately. The rebase is just a rebase, it doesn't affect this PR

src/kernel.c
rasky: If you write an assembly file like inthandler.S, surely you can change GP in there (like entrypoint.S does). Obviously such code would restore it afterwards, but if an interrupt happens while GP had the wrong value, the inthandler would call C code with the wrong GP, causing disasters. This is now fixed.

anacierdem:
> > Another idea: what about exposing a define to prevent the kernel checks? In case someone does not use it or provides their own alternative implementation.
> 
> Done.

Actually I was trying to reference `if (__kernel)`s but the ability to disable cookie checks is also a good idea :)

rasky:
> Actually I was trying to reference `if (__kernel)`s but the ability to disable cookie checks is also a good idea :)

Oh :) For that, I'd wait to see how things shape up. The hope is that there shouldn't be ifs in user code. For instance, code would call `dma_read` and would get the correct behavior whether it's using the kernel or not. Obviously this is making things more complicated for us (it would be easier to simply mandate kernel usage, like Nintendo did), but it might be worthwhile for a while, as we're exposing the real benefits of the kernel.

rasky:
I've realized that this PR is missing the part about making newlib safe for threading.

I think there are 3 things to do:

  1. keep one copy of the reentrancy structure (`struct reent`) in the TCB, and switch the global reent pointer (`_impure_ptr`) in the scheduler any time a thread switches. This allows each thread to have its own errno, and in general its own global context as required by libc. This is easy and has ~overhead, it just adds ~2K of RAM per each thread. BTW there is also a `__REENT_SMALL` define that we can investigate while compiling newlib to reduce this RAM overhead per thread.
  2. provide a reentrant-safe version of the syscalls in `system.c`, instead of the non-reentrant versions we're providing now. This is something we could do always, even if the kernel is not used, because it doesn't give overhead (in fact, it removes useless wrappers that newlib has to create for each syscall). We also need to rebuild newlib with macro `REENTRANT_SYSCALLS_PROVIDED`.
  3. implement locks as required by `sys/lock.h`. This would be easy if I had implemented mutex/semaphore in the kernel... which I haven't. I think for the audio library I will end up needing a mutex anyway, so I might as well work on it.

If somebody wants to help, I think 2 is an easy pick. It's a semi-mechanical change.

src/kernel.c
```diff
@@ -0,0 +1,652 @@
...
+			// that awakes a higher priority thread (THREAD_SWITCH_ISR).
+			if (C0_GET_CAUSE_EXC_CODE(stack_state->cr) == EXCEPTION_CODE_SYS_CALL)
+			{
+				// Skip syscall instruction when this thread will be scheduled
+				// again. Notice that, given that this is an explicit syscall,
+				// the thread must already have been added to a waiting list,
+				// otherwise it wouldn't ever be scheduled again.
+				assert(th_cur->flags & TH_FLAG_INLIST);
+				assertf(*(uint32_t*)th_cur->stack_state->epc == 0x0000004C, "invalid opcode found by __kthread_syscall_schedule:\nexpected 0x0000004C (SYSCALL 0x1), found: %08lx", *(uint32_t*)th_cur->stack_state->epc);
+				th_cur->stack_state->epc += 4;
```
anacierdem, line 212: This is not correct for branch delay slot case. We should also consider `C0_CAUSE_BD` and find the point to continue in that branch. Or we can simply hardcode a `nop` before the syscall as a temporary solution.

src/kernel.c
```diff
@@ -0,0 +1,652 @@
...
+ * This forces a context switch to happen, by invoking the syscall 0x1. The
+ * syscall will be intercepted by the generic interrupt handler (inthandler.S)
+ * that will call #__kthread_syscall_schedule.
+ *
+ * This macro is very low level and is called as part of higher-level primitives
+ * that force a context switch like #thread_yield.
+ *
+ * Do not call this under interrupt; use #KTHREAD_SWITCH_ISR instead.
+ */
+#define KTHREAD_SWITCH()      __asm("syscall 0x1")
```
anacierdem, line 32: It might be a good idea to wrap this with `MEMORY_BARRIER();`s to prevent unexpected behavior in case the compiler tries to move this around. Maybe this should be the default practice for all our assembly macros. I remember you telling the tick read getting misplaced as well right?

anacierdem (commented):
I think I have found a bug in case the syscall is placed in a branch delay slot. This is not something I have reproduced, but t is a correctness issue.

src/kernel.c
rasky: True. But the kernel is the only one to generate syscalls, and GCC can't possibly place them in a delay slot. Anyway, we can add an assert here (better safe than sorry).

src/kernel.c
anacierdem: Or rather we can add a memory clobber I suppose.

rasky:
Status update: this PR would be ready to go. I'm going to test it on a real world scenario (the audio library -- creating a audio thread) before actually merging this, so that we can verify that the API is sound and working.

As for the newlib reentrancy issue, this can be fixed in a followup.

anacierdem:
I just have one thing in my mind:
- It'd be nice if there was a way to disable preemption when we need so that we can use non-entrant things in threads. It's been a long time and I don't remember the code ATM. If just wrapping with disable/enable interrupts will do it let's merge this once you are content. Or is there a better way?

rasky:
For now you can disable interrupts for that. But to do the audio thread support, I'll add proper preemption disable, which is actually quite easy to implement (i think).

rasky:
This is now being merged to the preview branch of libdragon to begin experimenting with it. It's still too early to actually use it in production (see https://github.com/DragonMinded/libdragon/pull/151#issuecomment-906407018) but the preview branch is fine for this kind of improvements.

PR #152, opened 2021-07-07
Add support for RTC status/read/write commands

meeq:
The Joybus commands to read/write to RTC are based off of the [UltraPIF implementation](https://github.com/jago85/UltraPIF_MCU/blob/master/Src/main.c) (thanks @jago85!) and 64drive (thanks marshallh!)

Calling `rtc_init` will detect if an RTC is available, and if so will enable it and hook into newlib's `gettimeofday`. From there, software can either call `rtc_get` to read the current time into an `rtc_time_t` struct or use the ISO C time functions such as `time`. To adjust the current time, first call `rtc_is_writable` to determine if RTC writes are supported, and if so call `rtc_set` with an updated `rtc_time_t` struct.

Introduces a new `rtctest` example that demonstrates reading and writing the date/time from the cartridge RTC:

RTC Status | Screenshot
--- | ---
RTC Missing | rtc-missing
RTC Read-Only | rtc-nowrite
RTC Read-Write | rtctest-present

### Compiled `rtctest` ROM:
[rtctest.zip](https://github.com/DragonMinded/libdragon/files/6851355/rtctest.zip)

This ROM has been successfully tested on 64drive HW2 🎉 

This ROM does not allow changing the date/time on EverDrive64 X7, since the firmware does not support the RTC write SI command.

This ROM also does not allow setting the date/time on any emulator I have tested, since none of them handle the RTC write SI command. (Except my branch of cen64 that adds support for RTC writes: https://github.com/n64dev/cen64/pull/205)


jago85:
My RTC emulation may be useful for implementing the write RTC feature.
https://github.com/jago85/UltraPIF_MCU/blob/3b185ce4b2afd8d0489c992061ab43a8f39b064c/Src/main.c#L617

It's been some time since I experimented with this. But the procedure should be as follows:
- Write command is 0x08
  - Block 0 is the control register
  - Block 2 is the RTC data
- Stop RTC and disable write protection for block 2 (write to status register)
  - 08 00 01 04 00 00 00 00 00 00
- Write the new time to block 2
- Start RTC and enable write protection for block 2
  - 08 00 03 00 00 00 00 00 00 00

jago85:
I've tested the ROM and it's hanging at "Waiting for RTC to stop...". (using the UltraPIF RTC emulation.) I've noticed that the control register seems to be in the wrong byte order. I'm receiving 0x0400 instead of 0x0004. I think the problem is here: https://github.com/DragonMinded/libdragon/blob/867d7ae6c3914718556851e3f2c548f58dd72aec/src/rtc.c#L189

meeq:
@jago85 I followed the write commands from your previous comment:

> 08 00 01 04 00 00 00 00 00 00

So that should actually be:

> 08 00 01 00 04 00 00 00 00 00

Right?

src/rtc.c
```diff
@@ -0,0 +1,341 @@
...
+        0,
+        1
+    };
+    static uint64_t output[8];
+
+    controller_exec_pif( input, output );
+
+    uint8_t * recv_buf = (uint8_t *)&output[1];
+
+    return (((uint16_t)recv_buf[1]) << 8) | recv_buf[0];
```
meeq: This doesn't seem right to me, but appears to be correct based on the UltraPIF implementation: https://github.com/jago85/UltraPIF_MCU/blob/3b185ce4b2afd8d0489c992061ab43a8f39b064c/Src/main.c#L584-L585

@jago85 Are you sure that the control block bytes are swapped when reading vs writing?

jago85: You're right. This seems to be a bug in my implementation. I'll fix this and check if Animal Forest is happy. 
The memset in my code is also strange^^

jago85: I've checked with a real RTC. Of course, the byte order for reading and writing is the same. I'll fix the Read Control Register feature in my implementation. So I will keep interpreting the data as big endian uint16_t. Could also be the other way around, but this is the simplest change for me.
Animal Forest is also accepting the changes. It seems like the Control Register is only read to perform a Read-Modify-Write operation.

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+    data[0] = byte_to_bcd(rtc_time->sec);
+    data[1] = byte_to_bcd(rtc_time->min);
+    data[2] = byte_to_bcd(rtc_time->hour) + 0x80;
+    data[3] = byte_to_bcd(rtc_time->day);
+    data[4] = byte_to_bcd(rtc_time->week_day);
+    data[5] = byte_to_bcd(rtc_time->month + 1);
+    data[6] = byte_to_bcd(year);
+    data[7] = byte_to_bcd(year / 100);
+
+    joybus_rtc_write( 2, (uint64_t *)data );
```
rasky: You can't cast a uint8_t array to uint64_t because you have no guarantees on the alignment.

Do the opposite: declare a uint64_t and then cast it to uint8_t* to access it byte-wise

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ */
+static bool joybus_rtc_time_equal( const rtc_time_t * a, const rtc_time_t * b )
+{
+    return (
+        a->year  == b->year  &&
+        a->month == b->month &&
+        a->day   == b->day   &&
+        a->hour  == b->hour  &&
+        a->min   == b->min
+        /* Skip sec comparison in case the clock ticked between RTC reads */
```
rasky: Technically, min could tick as well if sec was 59. Since this function is not very well defined semantically speaking, I'd inline it in its only caller, where ignoring seconds is OK because it is set as 3 (so it's unlikely to reach 60 before we read it... though multi threading might interfere)

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+#define JOYBUS_RTC_WRITE_DELAY 20
+
+/**
+ * @brief The RTC is ready for block 2 to be read.
+ */
+#define JOYBUS_RTC_STATUS_READY 0x00
+/**
+ * @brief The RTC is ready for block 2 to be written.
+ */
+#define JOYBUS_RTC_STATUS_BUSY  0x80
```
rasky: Comment and variable name don't seem to match

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+#define JOYBUS_RTC_CONTROL_MODE_SET 0x0004
+/**
+ * @brief Control mode for normal operation of the RTC.
+ *
+ * Clears the busy bit sets the write-protection bits.
+ *
+ * 0x0100 is the block 1 write-protection bit.
+ * 0x0200 is the block 2 write-protection bit.
+ *
+ * RTC block 1 has an unknown purpose and is typically unimplemented.
```
rasky: Unimplemented by whom? Emulators? I'd mention it explicitly.

Do we know the actual RTC chip used by animal crossing? Its data sheet might provide some insights even di the communication is shielded by the PIF

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+
+#include "libdragon.h"
+
+/**
+ * @defgroup rtc Real-Time Clock Subsystem
+ * @ingroup libdragon
+ * @brief Joybus real-time clock interface.
+ *
+ * The Joybus real-time clock is an accessory that was only included on
+ * one official cartridge that was only available in Japan. The Joybus RTC
```
rasky: IIUC, it was an official option supported by Nintendo, implemented in PIF firmware, but only animal crossing decided to use it in its cartridge. I'd specify this, because otherwise it sounds like a game-specific hardware extension which would have very little relevance

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * is accessed through the serial interface (SI) similar to EEPROM.
```
rasky: ... and joystick

src/rtc.c
rasky: I'd also specify here in this first paragraph that it's supported by some development kits (64drive / ed64) and some emulators (list here which)

include/rtc.h
```diff
@@ -0,0 +1,57 @@
...
+#endif
+
+bool joybus_rtc_is_present( void );
+bool joybus_rtc_is_busy( void );
+bool joybus_rtc_is_writable( void );
+
+void joybus_rtc_run( void );
+void joybus_rtc_set( const rtc_time_t * rtc_time );
+
+void joybus_rtc_read_time( rtc_time_t * rtc_time );
```
rasky: I think this should be symmetric with the set command. Either set/get or read/write.

include/rtc.h
```diff
@@ -0,0 +1,57 @@
...
+    uint8_t sec;
+    /** @brief Day of week. [0-6] (Sun-Sat) */
+    uint8_t week_day;
+} rtc_time_t;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+bool joybus_rtc_is_present( void );
```
rasky: Do we really need the joybus prefix in all these names?

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * A real RTC uses a battery to power a clock that runs even when the N64
+ * is powered-off and maintains the date, time, and day of the week.
```
rasky: Maybe I'd move this as the very first paragraph to quickly describe what a RTC is.

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * To check if the real-time clock is available, call #joybus_rtc_is_present.
+ * If detected, it is a good idea to ensure the RTC control block is in a
+ * known-good state by calling the #joybus_rtc_run function before reading
+ * the current time with #joybus_rtc_read_time for the first time.
+ *
+ * The simplest method to change the time is by calling #joybus_rtc_set,
+ * which is a high-level convenience function that prepares the RTC and
+ * writes the time with the proper delays.
```
rasky: I'd say this paragraph could be simplified since there's only one way now, and also details about delays are irrelevant in a general overview.

I'd mention timing, though. Maybe people won't expect a write to take many milliseconds

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * Many emulators do not actually implement the RTC write command and will
+ * always respond with the host system's current local time. Some emulators
+ * and flash carts support writing to RTC but will not persist the date/time
+ * after resetting or powering-off.
+ *
+ * The only way to check if writes are actually supported is to write a
+ * time to the RTC and read the time back out. Many emulators that do
+ * support RTC reads will silently ignore RTC writes. You should detect
+ * whether writes are supported using #joybus_rtc_is_writable so that you can
+ * conditionally show the option to change the time if it's supported.
```
rasky: Again, mention timing.

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * which is a high-level convenience function that prepares the RTC and
+ * writes the time with the proper delays.
+ *
+ * In general, since only one game ever used Joybus RTC (and that game was
+ * later re-released on the GameCube in English), RTC support by emulators
+ * and flash carts tends to be incomplete, inaccurate, or non-existent.
+ * Many emulators do not actually implement the RTC write command and will
+ * always respond with the host system's current local time. Some emulators
+ * and flash carts support writing to RTC but will not persist the date/time
+ * after resetting or powering-off.
```
rasky: This paragraph is a little generic. Given that there's a handful of supported cartridges / emulators that are relevant, I'd simply list them all with their current status (mentioning version number, or a general "as of July 2021")

rasky: I'd drop the very first sentence. Don't create excuses for emulator writers 😃

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * The only way to check if writes are actually supported is to write a
+ * time to the RTC and read the time back out. Many emulators that do
+ * support RTC reads will silently ignore RTC writes. You should detect
+ * whether writes are supported using #joybus_rtc_is_writable so that you can
+ * conditionally show the option to change the time if it's supported.
+ *
+ * The RTC sends and receives the date/time in bytes of binary-coded decimal
+ * numbers. This subsystem handles decoding and encoding the date/time into
+ * a more generally-useful struct called #rtc_time_t.
```
rasky: The internal protocol is not very useful in this overview, so I'd remove this, though I'm not strong on this. Up to you.

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * RTC is present on the cartridge and a status byte to indicate whether the
+ * RTC is in the "busy" state.
+ *
+ * Generally, you should not need to call this function directly. Prefer
+ * calling the high-level #joybus_rtc_is_present and #joybus_rtc_is_busy
+ * functions which handle the magic numbers associated with the response.
+ *
+ * @return the RTC status response data
+ */
+uint16_t joybus_rtc_status( void )
```
rasky: Static?

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * @param[in]   block
+ *              Which RTC block to read from (0-2)
+ *
+ * @param[out]  data
+ *              Destination pointer for the RTC block data
+ *
+ * @return the busy status of the real-time clock
+ */
+uint8_t joybus_rtc_read( uint8_t block, uint64_t * data )
```
rasky: Static

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * @param[in]   block
+ *              Which RTC block to write to (0-2)
+ *
+ * @param[out]  data
+ *              RTC block data to write
+ *
+ * @return the busy status of the real-time clock
+ */
+uint8_t joybus_rtc_write( uint8_t block, const uint64_t * data )
```
rasky: Static

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+        0x0000000002090700,
+        0xffffffffffffffff,
+        0xfffe000000000000,
+        0,
+        0,
+        0,
+        0,
+        1
+    };
+    static uint64_t output[8];
```
rasky: Why this has to be static?

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+        0x000000000A010800,
+        0xffffffffffffffff,
+        0xfffe000000000000,
+        0,
+        0,
+        0,
+        0,
+        1
+    };
+    static uint64_t output[8];
```
rasky: Same as above

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+
+/**
+ * @brief Read the status of the real-time clock.
+ *
+ * Some flash carts require the RTC to be explicitly enabled.
+ * Some emulators and flash carts do not support RTC at all.
+ *
+ * @return whether the Joybus RTC is present on the cartridge.
+ */
+bool joybus_rtc_is_present( void )
```
rasky: Can you please add rough timing in milliseconds to all public function documentation in this file?

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * handle the delays and calibration propagation. Call #joybus_rtc_status or
+ * #joybus_rtc_is_busy to determine which control mode the RTC is in.
+ *
+ * @param[out]  control
+ *              Destination pointer for the RTC control data.
+ *
+ * @param[out]  calibration
+ *              Destination pointer for the RTC calibration data.
+ */
+void joybus_rtc_read_control( uint16_t * control, uint32_t * calibration )
```
rasky: Static

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+
+    if( control != NULL ) *control = data >> 48;
+    if( calibration != NULL ) *calibration = data;
+}
+
+/**
+ * @brief Read the current date/time from the real-time clock.
+ *
+ * Your code should call this once per frame to update the rtc_time_t
+ * data structure.
```
rasky: Once per frame seems overkill to me... what's the timing of this function?

Anyway I think this should be called once at boot and then an internal time should be calculated using timer_ticks(). This is why I suggest hooking this up to newlib, so that we can provide the correct abstraction to calculate real time much faster than polling joybus

include/rtc.h
```diff
@@ -0,0 +1,57 @@
+/**
+ * @file rtc.h
+ * @brief Real-Time Clock Subsystem
+ * @ingroup rtc
+ */
```
rasky, line 5: Do users need to call controller_init() to be able to use rtc functions?

include/controller.h
```diff
@@ -195,6 +195,7 @@ extern "C" {
 #endif
 
 void controller_init( void );
+void controller_exec_pif( const void *inblock, void *outblock );
```
rasky: Maybe this should be called joybus_exec() and moved to joybus.c file, together with a non-installed joybus.h header, included by controller.c and rtc.c

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ *
+ * This function is a good idea to call immediately after verifying
+ * that the RTC is present and before the first call to
+ * #joybus_rtc_read_time so that you can be sure the RTC is ready.
+ *
+ * This function will take approximately 25 milliseconds to complete.
+ *
+ * It is not necessary to call this after calling #joybus_rtc_set.
+ */
+void joybus_rtc_run( void )
```
rasky: I understand the comment you wrote but since I don't understand the control mode thing, this API looks strange to me.

Maybe we should just have a bool rtc_init() that returns false if the RTC doesn't exist, and initializes it otherwise, so that it's ready to run.

Maybe this init function should also detect immediately if the RTC is writable... this would conflate a slow operation into a init() function which runs at a boot time where milliseconds don't possibly count much

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+void joybus_rtc_set( const rtc_time_t * write_time )
+{
+    uint32_t calibration;
+    /* Read the calibration data from the control block */
+    joybus_rtc_read_control( NULL, &calibration );
+    /* Prepare the RTC to write the time, preserving the calibration data */
+    joybus_rtc_write_control( JOYBUS_RTC_CONTROL_MODE_SET, calibration );
+    wait_ms( JOYBUS_RTC_WRITE_DELAY );
+    /* Check the RTC status to make sure RTC block 2 writes are supported */
+    if( joybus_rtc_is_busy() )
```
rasky: What happens if this if is not taken? It looks to me the function fails, but it's not reported. Also the RTC is left in write mode?

src/rtc.c
```diff
@@ -0,0 +1,518 @@
...
+ * calling the high-level #joybus_rtc_run and #joybus_rtc_set functions which
+ * handle the delays and calibration propagation.
+ *
+ * @param[in]  control
+ *             The control block data
+ *
+ * @param[in]  calibration
+ *             The opaque calibration data from #joybus_rtc_read_control
+ */
+void joybus_rtc_write_control( uint16_t control, uint32_t calibration )
```
rasky: Static

rasky (changes requested):
Thanks! This is spectacular work, and the fact that you also allowed progresses to be made in emulators, everdrive, and even ultrapif is simply outstanding!

I've made a first review from the phone. The code seems ok apart from a potential alignment bug, so I focused on documentation and API ergonomic.

src/rtc.c
meeq: Interestingly, as far as I can tell this comment is correct. When the RTC is in "set mode", you have to wait for the busy status before you can write to it. Once the RTC is put back into "run mode" you have to wait for the busy status to clear before continuing.

The name "busy" is consistent with [other documentation I have found](https://github.com/BluRosie/doubutsu-no-mori/blob/default/docs/disassembled%20RTC%20access%20routines.txt)

src/rtc.c
meeq: I have not found any documentation about what block 1 is for. No emulator implements it (that I have found), the UltraPIF and 64drive also do not implement it.

As far as I can tell, Animal Forest does not ever access block 1, so its purpose is a complete mystery. I'll do a bit more research and see if I can figure out what it is from the actual component.

include/rtc.h
meeq: Nope. `controller_init` is only needed for actual controller inputs, not SI peripherals.

src/rtc.c
meeq: I'm totally on-board with this 👍🏻

include/controller.h
meeq: I'm totally on-board with this 👍🏻

src/rtc.c
meeq: Some emulators (Mupen64Plus and cen64 [without my patch]) don't implement the RTC busy status at all and also don't support writes. This just bails on the write if the busy status is unimplemented.

The reason it fails silently is because I have to treat all RTC write errors as silent failures. The intent is that the developer calls `joybus_rtc_is_writable` to determine if writes are implemented. Somewhere in the git history I had this function return a bool of whether the write was successful, but it ended up basically being an implementation of `joybus_rtc_is_writable` inside `joybus_rtc_set`.

I'm open to suggestions about how to handle these silent write failures better

src/rtc.c
meeq: I copied this pattern from `controller.c` -- I have no other reasoning for it being `static`

src/rtc.c
meeq: That recommendation is based off of what Animal Forest does. 🤷🏻

include/rtc.h
meeq: We do if we ever want to support the N64DD RTC, which has a completely different interface.

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

include/rtc.h
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: I mentioned it here because I want this to be a resource for people to understand how it actually works as well as how to use the APIs that we provide.

I don't think it's hurting anything to keep it, and I think it helps contextualize what the library is actually doing for them.

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/e46fbb07a6cbb01a98f05a2d3b170d4ff1860b77

src/rtc.c
meeq: I included the first sentence to set expectations for homebrew developers that depending on RTC functionality will inevitably limit where their software will correctly run (at the moment). RTC is currently an esoteric novelty that really needs more momentum behind it before emulators are incentivized to fully support it.

src/rtc.c
jago85: If I remember correctly, block 1 is an 8 byte battery backed SRAM.
I could do some tests if needed. I have a real RTC and with the UltraPIF and a debugger I can directly send commands to the RTC and see the results.

src/rtc.c
rasky: If the bit is really called "busy", maybe its semantic meaning is reversed? I think it's normal to wait for a busy bit to be *off*, or of a ready bit to be *on*. I think calling it "ready" is more clear if we wait for it before writing.

include/rtc.h
rasky: I see. OTOH I'm not sure if the average libdragon user is interested in which RTC there is, rather than "a RTC", so maybe the n64dd rtc should be implemented behind the same API?

src/rtc.c
rasky: ok

src/rtc.c
rasky: ok

src/rtc.c
rasky: Oh I see. I think it's not required (or I can't see how it should be required). Since it's already difficult to check this PR, maybe we should remove it from controller.c first and check it there.

src/rtc.c
rasky: Still, I think it's not a good suggestion at least for now. 

First, we don't expose an API to be able to fetch RTC in background. I don't know timing for `joybus_rtc_get`, but you need to wait for a full joybus handshake so it can't be very fast. I'm sure Animal Forest was using background fetching (issue the command, then wait for interrupt and process it). Once this library will be threading compatibly, maybe it'd make more sense.

Second, LibUltra doesn't have a 64-bit timer. We do expose a 64-bit timer via `timer_ticks()`. So with LibUltra, you have no way to keep up with the current time, because the 32-bit COP0 counter overflows too often. What I would expect is that our `gettimeofday` implementation would transparently read RTC once (at boot or first call), and then store the current time as a 64-bit offset, so that later it is sufficient to add that offset to `timer_ticks()` to obtain the correct time, and then use newlib functions to convert it into calendar dates when required.

include/rtc.h
rasky: Good

src/rtc.c
```diff
@@ -0,0 +1,528 @@
...
+ * must be explicitly enabled in the OS or with a ROM header configuration;
+ * RTC will not be detected if the EEPROM save type is used; RTC writes are
+ * not supported through the SI, so changing the time must be done in the OS.
+ * UltraPIF fully implements an emulated RTC that can be accessed even when
+ * the cartridge does not contain a Real-Time Clock or battery.
+ *
+ * The only reliable way to check if writes are actually supported is to write
+ * a time to the RTC and read the time back out. Many emulators that do
+ * support RTC reads will silently ignore RTC writes. You should detect
+ * whether writes are supported using #joSybus_rtc_is_writable so that you can
```
meeq: Note to self: typo

src/rtc.c
```diff
@@ -0,0 +1,528 @@
...
+ */
+
+#include "libdragon.h"
+
+/**
+ * @defgroup rtc Real-Time Clock Subsystem
+ * @ingroup libdragon
+ * @brief Joybus real-time clock interface.
+ *
+ * The Joybus real-time clock is an peripheral that uses a battery to power
```
meeq: Note to self: "a cartridge peripheral"

src/rtc.c
```diff
@@ -0,0 +1,528 @@
...
+ * 
+ * Some notable examples of RTC support (as of July 2021):
+ * 64drive hw2 fully implements RTC including writes, but requires hard-coded
+ * delays after setting the time (see #JOYBUS_RTC_WRITE_FINISHED_DELAY).
+ * EverDrive64 3.0 and X7 partially support the RTC, with caveats: The RTC
+ * must be explicitly enabled in the OS or with a ROM header configuration;
+ * RTC will not be detected if the EEPROM save type is used; RTC writes are
+ * not supported through the SI, so changing the time must be done in the OS.
+ * UltraPIF fully implements an emulated RTC that can be accessed even when
+ * the cartridge does not contain a Real-Time Clock or battery.
```
meeq: Note to self: "does not support the real-time clock peripheral"

src/rtc.c
meeq: I think the most accurate name for this status is actually "STOPPED".

This is consistent with the [documentation of the `0x0004` control register bit as "disable clock"](https://github.com/BluRosie/doubutsu-no-mori/blob/default/docs/Clock%20Access.txt)

src/rtc.c
```diff
@@ -0,0 +1,528 @@
...
+    uint32_t calibration;
+    /* Read the calibration data from the control block */
+    joybus_rtc_read_control( NULL, &calibration );
+    /* Prepare the RTC to write the time, preserving the calibration data */
+    joybus_rtc_write_control( JOYBUS_RTC_CONTROL_MODE_SET, calibration );
+    wait_ms( JOYBUS_RTC_WRITE_BLOCK_DELAY );
+    /* Check the RTC status to make sure RTC block 2 writes are supported */
+    if( !joybus_rtc_is_busy() ) return false;
+    /* Write the updated time to RTC block 2 */
+    joybus_rtc_write_time( write_time );
```
meeq, line 750: Note to self: recalculate day of week from year/month/day before writing to RTC

src/rtc.c
meeq: @jago85 is the RTC actually a feature that was always intended to be supported by the PIF, or is it a game-specific convention that's mostly implemented on the cartridge circuitry? From what I can tell, it behaves a lot like EEPROM (RTC is on the same PIF channel as EEPROM) but responds to different command identifiers.

jago85: Oh, I really don't know when they were planning this. Your question is about the original PIF, isn't it? 
Yes, the RTC is completely implemented inside the cart, just using the SI to transport the data. The PIF doesn't care what is connected to the bus and what the commands are good for. It just shifts the data out to and in from one of the five SI channels. So from the PIF's point of view, the interface is pretty generic.
I've heared that Animal Forest was originally planned for the 64DD. As it flopped they ported the game to cartridge and needed a replacement RTC. So they designed an RTC chip with SI interface. And it was designed with the ability to coexist with an EEPROM chip on the same bus.

src/rtc.c
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/6424043b8c17697de4377b9e4c6c6009b89ac490

include/rtc.h
meeq: I have removed the joybus prefix from the API; most of the static methods still have the joybus_ prefix if they are actually joybus-specific. If/when we add support for the N64DD RTC, it can use the same interface and the LibDragon RTC subsystem can abstract away the differences

include/controller.h
meeq: Addressed in https://github.com/DragonMinded/libdragon/pull/152/commits/d198b72f90aa4b7d832146d3d78998e99a0db48b

src/rtc.c
meeq: The actual component does not appear to be an off-the-shelf RTC; it's branded `Nintendo RTCK-NUS` and `A0108C`. Not surprisingly, I am not able to find a datasheet based on that. 
animalforestrtc

src/rtc.c
meeq: Fixed

src/rtc.c
meeq: Fixed

meeq:
@rasky This is ready for review again.

I have integrated the RTC Subsystem with newlib's `gettimeofday` and added a demo using ISO C time functions to the `ctest` example.

I have also refactored the RTC Subsystem in anticipation for supporting the 64DD RTC (cen64 has a partial implementation of it that can read time, so we have a starting point).

Conceivably, we could also support a "Simulated RTC" that would have to be set on startup and maintain the date/time as long as the system is powered on. This would be "better than nothing" for players who want to play RTC-enabled homebrew but don't actually have RTC-capable hardware.

The `rtc_get` function now has a cache that invalidates after 1 second so that homebrew can read the clock every frame without paying the cost of hitting the SI just to see if the clock ticked.

I considered implementing your suggestion of only reading the RTC once during initialization, but it got complicated. If we do end up implementing the "Simulated RTC", we will basically end up with what you proposed though.

I need to put together some timing tests to figure out what the real cost of calling each of these API methods is -- unfortunately I don't have an RTC on my flash cart, so I'll have to enlist some EverDrive and 64drive owners in Discord to get me hard numbers.

Of course, this will all be better once we integrate the Joybus commands with your fancy new kernel. 😄 

Please let me know if you have any other suggestions here. Thank you for your feedback!

Makefile
```diff
@@ -66,8 +66,10 @@ install: libdragon.a libdragonsys.a
 	install -m 0644 include/debug.h $(INSTALLDIR)/mips64-elf/include/debug.h
 	install -m 0644 include/usb.h $(INSTALLDIR)/mips64-elf/include/usb.h
 	install -m 0644 include/console.h $(INSTALLDIR)/mips64-elf/include/console.h
+	install -m 0644 include/joybus.h $(INSTALLDIR)/mips64-elf/include/joybus.h
```
rasky, line 69: Can we hold on installing this (and making `joybus_exec` public API)? I think the API might change as part of the integration with the kernel, so if we can hold up a little on this, we can avoid later breaking things that start depending on it.

We can keep the include file internal, so that only libdragon itself can use `joybus_exec` for now.

src/rtc.c
```diff
@@ -0,0 +1,792 @@
...
+ * @brief Real-time clock interface.
+ *
+ * The Joybus real-time clock is a cartridge peripheral that uses a battery
+ * to power a clock that tracks the date, time, and day of the week. The
+ * real-time clock keeps running even when the N64 is powered-off. The
+ * Joybus RTC is accessed through the serial interface (SI) similar to EEPROM
+ * and controllers. The Joybus RTC was only ever available on one official
+ * cartridge that was only available in Japan: Dōbutsu no Mori (Animal Forest).
+ * There is also a real-time clock included in the N64DD hardware, which uses
+ * a different interface and is not currently supported by libdragon.
```
rasky, line 27: I think here we should at least mention that RTC is supported by 64drive and Everdrive (and emulators), so it can still be useful for homebrew development, otherwise it seems weird to an user why libdragon should care about a peripheral that was used once in history.

src/rtc.c
```diff
@@ -0,0 +1,792 @@
...
+ *
+ * To check if the real-time clock is available, call #rtc_init.
+ * To read the current time from the real-time clock, call #rtc_get.
+ * To check if the real-time clock supports writes, call #rtc_is_writable.
+ * To write a new time to the real-time clock, call #rtc_set.
```
rasky, line 35: I would mention here that you can call `time(NULL)` to get the current time. Something like `call #rtc_get or time(NULL)`.

src/rtc.c
```diff
@@ -0,0 +1,792 @@
...
+bool rtc_get( rtc_time_t * rtc_time )
+{
+    /* libdragon currently only supports getting the time for Joybus RTC! */
+    if( rtc_present() != RTC_JOYBUS ) return false;
+
+    /* This should be overwritten the first time this function is called */
+    static rtc_time_t cache_time = { 2000, 0, 1, 0, 0, 0, 6 };
+
+    uint32_t current_ticks = get_ticks();
+    int32_t distance = TICKS_DISTANCE( current_ticks, rtc_get_cache_ticks );
```
rasky: The prototype is `TICKS_DISTANCE(from, to)`, so I think your usage here is reversed?

src/rtc.c
```diff
@@ -0,0 +1,792 @@
...
+ * the cache validity.
+ * 
+ * Cache will invalidate every #RTC_GET_CACHE_INVALIDATE_TICKS.
+ * Cache will also invalidate when the tick counter overflows.
+ * Calling #rtc_set will also invalidate the cache.
+ *
+ * @param[out]  rtc_time
+ *              Destination pointer for the RTC time data structure
+ * 
+ * @return whether the rtc_time destination pointer data was modified
```
rasky, line 691: This should mention how slow is this function (at least whenever it really reads the RTC).

src/rtc.c
```diff
@@ -0,0 +1,792 @@
...
+}
+
+/**
+ * @brief High-level convenience helper to set the RTC date/time.
+ *
+ * Prepares the RTC for writing, sets the new time, and resumes the clock.
+ *
+ * This process will take approximately 570 milliseconds to complete.
+ * Unfortunately, the best way to ensure that writes to the RTC have
+ * actually finished is by waiting for a fixed duration. Emulators may not
```
rasky, line 726: This sentence is a little confusing to me. Does this mean that the user has to wait a fixed duration after calling `rtc_set`? I guess the meaning is that 570ms includes a conservative delay, but maybe it should be clarified.

rasky (approved):
> I have integrated the RTC Subsystem with newlib's gettimeofday and added a demo using ISO C time functions to the ctest example.

Great thanks!

> Conceivably, we could also support a "Simulated RTC" that would have to be set on startup and maintain the date/time as long as the system is powered on. This would be "better than nothing" for players who want to play RTC-enabled homebrew but don't actually have RTC-capable hardware.

One option I thought of is that we could add a feature to the debugging library to read the current time via USB (extending the UNFLoader protocol, so to be able to read the time on the host PC). But given that both 64drive hw2 and Everdrive X7 have a RTC already, it looks like this feature would be only useful to 64drive hw1 owners. Not sure it's worth it.

> The rtc_get function now has a cache that invalidates after 1 second so that homebrew can read the clock every frame without paying the cost of hitting the SI just to see if the clock ticked.

I think the cache is a reasonable compromise. What I suggested would still be superior because it would basically means that `rtc_get / time(NULL)` is always ~free, but I can see that it's going to be more complicated.

> Of course, this will all be better once we integrate the Joybus commands with your fancy new kernel. 😄

Sure! Joybus is pretty slow in general so doing it in background sounds like a good idea anyway.

Makefile
meeq: I'm on the fence about this. On the one hand, I agree that it's probably a good idea not to expose APIs that are subject to change in the near future, but I also see immense utility in exposing the Joybus commands outside of libdragon (for example, this Randnet Keyboard test ROM: https://github.com/meeq/libdragon/commit/44aff08151a3e8d667e14dff59c806f7acc4ae30)

I think there's going to be a huge shake-up in many of the APIs once the kernel lands, so I think we need to have a clear plan for the post-kernel transition process that either minimizes breakage or draws hard lines on where breakage is okay.

src/rtc.c
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/152/commits/544baa50639e55196c93c17c97b8cda1365f7bfc

src/rtc.c
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/152/commits/544baa50639e55196c93c17c97b8cda1365f7bfc

src/rtc.c
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/152/commits/544baa50639e55196c93c17c97b8cda1365f7bfc

src/rtc.c
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/152/commits/544baa50639e55196c93c17c97b8cda1365f7bfc

src/rtc.c
meeq: Fixed by using `timer_ticks` instead of `get_ticks`

jago85:
I've tested the new ROM again and can confirm that it works with the real RTC. It's working with the emulation of the UltraPIF and the Brutzelkarte too. Good job! :smile:

rasky:
After a chat with @meeq and his latest rework for using the 64-bit timer for cache invalidation, this PR is ready to go in for me.

@anacierdem @DragonMinded the only minor point left here is whether we want to expose `joybus_exec` right now, on the verge of possibly/probably having to change it again to be able to run joybus commands in background with the kernel. I defer to your judgment, it's a minor point for me.

anacierdem:
Im ok to have it exposed with a comment to say it maybe subject to change.

examples/rtctest/Makefile
```diff
@@ -0,0 +1,62 @@
...
+CHKSUM64 = $(SDK_DIR)/bin/chksum64
+N64TOOL = $(SDK_DIR)/bin/n64tool
+
+# LibDragon Flags
+OUT_SIZE = 1052672B # 52672B HEADER + 1M (minimum) ROM
+N64TOOLFLAGS = -l $(OUT_SIZE) -h $(ROM_HEADER) -t $(PROG_TITLE)
+
+# Byte-swap ROM
+ifeq ($(N64_BYTE_SWAP),true)
+ROM_EXTENSION = .v64
```
anacierdem: We can change these to match the new version as well now

src/rtc.c
```diff
@@ -0,0 +1,803 @@
...
+    if( rtc_time->day > days_in_month ) rtc_time->day = days_in_month;
+
+    /* Recalculate the day of the week based on calendar date */
+    month = month + 1; /* Adjust month to 1-indexed */
+    if( month < 3 )
+    {
+        month = month + 12;
+        year = year - 1;
+    }
+    rtc_time->week_day = (
```
anacierdem, line 657: Can you provide a reference for this I could not find this exact method 🙂

anacierdem (approved):
A few minor things. Will merge when you say ok

examples/rtctest/Makefile
```diff
@@ -0,0 +1,62 @@
...
+LDFLAGS = --library=dragon --library=c --library=m --library=dragonsys
+LDFLAGS += --library-path=$(SDK_LIB_DIR) --script=n64.ld --gc-sections
+
+# LibDragon binaries
+ROM_HEADER = $(SDK_LIB_DIR)/header
+CHKSUM64 = $(SDK_DIR)/bin/chksum64
+N64TOOL = $(SDK_DIR)/bin/n64tool
+
+# LibDragon Flags
+OUT_SIZE = 1052672B # 52672B HEADER + 1M (minimum) ROM
```
sp1187: This explanation is incorrect. Rather the minimum size is 1 MiB = 1048576 bytes (not decimal megabytes) and the header size is 4096 bytes.

meeq: You’re right, that was a gnarly workaround for a bug in n64tool that has now been fixed. Good catch.

src/rtc.c
meeq: It’s based on a snippet that @networkfusion sent to me — he said it was from a homework problem for one of his Computer Science classes 😄  

I’ve verified that it’s correct, at least for the date ranges that are supported by RTC.

networkfusion: Yes, written by me when in college as part of a paper to check if a calendar algorithm was incorrect and prove why. (I will see if I can dig out the original source!)

networkfusion: Attribution would be welcome 😉 
[WPFOdrinalCalculator.zip](https://github.com/DragonMinded/libdragon/files/6872689/WPFOdrinalCalculator.zip)

networkfusion: (BTW, my teacher thought that there was nothing wrong with the original algorithm until he saw my paper 🤣 )

PR #153, opened 2021-07-07
Refactor n64tool, add ed64romconfig binary

meeq:
Enhanced `n64tool` with support for long command-line flags, better error messages, and support for 1M ROM size. Removed support for byte-swapped V64 ROM format.

```
Usage: n64tool [-t <title>] -l <size>B/K/M -h <file> -o <file> <file> [[-s <offset>B/K/M] <file>]*

This program creates an N64 ROM from a header and a list of files,
the first being an Nintendo64 binary and the rest arbitrary data.

Command-line flags:
        -t, --title <title>    Title of ROM (max 20 characters).
        -l, --size <size>      Force ROM output file size to <size> (min 1 megabyte).
        -h, --header <file>    Use <file> as IPL3 header.
        -o, --output <file>    Save output ROM to <file>.
        -s, --offset <offset>  Next file starts at <offset> from top of memory. Offset must be 32-bit aligned.

Size/offset suffix notation:
        B for bytes.
        K for kilobytes.
        M for megabytes.
```

Created a separate `ed64romconfig` binary that can be run after `n64tool` (but before `chksum64`) that sets the [EverDrive64 ROM Header Configuration](https://github.com/N64-tools/ED64/blob/develop/docs/rom_config_database.md#developer-override):

```
Usage: ed64romconfig [-r] [-c] [-w <savetype>] <file>

This program takes a big-endian N64 ROM and sets the header so that
EverDrive64 will respect the declared save type, RTC, and region-free
settings without needing to create a save_db.txt entry for it.
See: https://github.com/krikzz/ED64/blob/master/docs/rom_config_database.md#developer-override

Command-line flags:
        -w, --savetype <type>  Declare cartridge save type.
        -c, --rtc              Declare real-time clock support.
        -r, --regionfree       Declare region-free ROM.

Supported cartridge save types:
        none        Game does not save or uses Controller Pak.
        eeprom4k    Game saves to 4 kilobit EEPROM.
        eeprom16k   Game saves to 16 kilobit EEPROM.
        sram256k    Game saves to 256 kilobit SRAM
        sram768k    Game saves to 768 kilobit SRAM
        sram1m      Game saves to 1 megabit SRAM
        flashram    Game saves to 1 megabit FlashRAM
```

Refactored all of the `examples` Makefiles to follow standard conventions, add comments for special-cases, and use consistent patterns.

#### Premise
Big-endian (`.z64`) byte ordering is the canonical Nintendo 64 ROM format. Little-endian (`.n64`) and byte-swapped (`.v64`) formats are confusing and only exist for historical reasons that are no longer relevant for the overwhelming majority of use-cases. If there is any legitimate reason to prefer publishing a byte-swapped ROM instead of big-endian, I could not find it.

#### Argument
libdragon has a responsibility to create authentic artifacts that target real Nintendo 64 hardware. Nintendo 64 ROM files are interpreted in a big-endian environment. There should be no question by homebrew developers as to which ROM format to use. 

Developers should not be distributing ROMs that require "unmangling" in order to memory-map the data properly. Flash carts and emulators should not have to make assumptions about how to interpret a Nintendo 64 ROM file based on hard-coded magic numbers of the first 4 bytes of the ROM header.

The most developer-friendly approach is to deprecate the `.n64` and `.v64` formats and strongly encourage everyone to use `.z64` going forward.

#### Counter-Arguments

* "The physical Mask ROM byte order is not big-endian"
  * **Response:** The intended representation of the ROM data is clearly big-endian:
    * The N64's MIPS CPU is big-endian.
    * The console's memory map and DMA read the cartridge ROM in big-endian.
    * The game title in the ROM header is only readable in big-endian.
    * The official Nintendo SDK emits ROM files in big-endian.
* "What about the theoretical homebrew developer still using a Doctor V64?"
  * **Response:** If such person exists, they can convert the `.z64` to `.v64` as part of their build process. When they go to share their work to others, they should publish the `.z64` ROM file.
* "Byte-swapped ROM support already existed; keeping it around costs us nothing"
  * **Response:** The maintenance burden is not zero: `n64tool`'s byte-swap flag was responsible for two subtle easy-to-miss bugs while modifying the code. `n64tool` does not need the additional complexity of byte-swapping. The feature is completely orthogonal to the other functionality of `n64tool`. There are plenty of other tools that solve the generalized problem of byte-swapping a file.
* "This could cause breakage for existing workflows" 
  * **Response:** This is the strongest argument against my proposal, and my only hesitation to removing this feature. I do not want to break anyone's build. However, I do want to encourage anyone currently building a byte-swapped ROM to start publishing big-endian (`.z64`) instead, and I don't think the solution to continue generating V64 is difficult, onerous, or unreasonable. `n64tool` will suggest the fix if the `-b` flag is used, so affected developers will know exactly how to embrace this change. 

#### Compromise

The process of building a ROM file should always be in big-endian byte-order. Byte-swapping can be done as a post-processing step after the big-endian ROM file is finalized.

The refactored `examples` Makefiles still support the `N64_BYTE_SWAP` flag, but the implementation has changed so that it is now a distinct target that depends on the `.z64` build:

```Make
ifeq ($(N64_BYTE_SWAP),true)
$(PROG_NAME).v64: $(PROG_NAME).z64
	dd conv=swab if=$^ of=$@
endif
```

If a developer really wants a `.v64` ROM, they still can easily create one, but it will no longer be handled by `n64tool`.

#### Backlash-anticipation
I understand that libdragon has its own legacy of using the V64 format, and that some individuals wished to continue supporting the byte-swapped format to prevent churn. I welcome anyone to provide opinions on how to improve this proposal.


meeq:
The magic rom configuration has been verified on EverDrive hardware. A warning has been added for the case of EEPROM + RTC which is not supported on real hardware (but works on most emulators).

jago85:
Why do you think the RTC and EEPROM can not work together in real hardware? Did someone test this already? I believe it could be possible. The chips are connected to the same PIF channel. But this is a single wire bi-directional data line. If I remember correctly it is using an open-drain configuration (each chip can only drive the line low). As the chips are using different commands (EEPROM: 0x00, 0x04, 0x05 / RTC: 0x06, 0x07, 0x08), they may work perfectly without any interference. But I never testet it with real chips.

meeq:
@networkfusion tested this on real hardware with an EverDrive X7 and was only able to get the RTC to work if EEPROM wasn’t used. If both EEPROM and RTC were enabled, EEPROM would be detected but RTC would not.

jago85:
Ah ok. So, at least the ED64 X7 doesn't support it. And real PCBs with both chips don't exist (to my knowledge). Would be a nice experiment to build one :)

meeq:
The test ROM in https://github.com/DragonMinded/libdragon/pull/152 will happily detect both EEPROM and RTC together on emulators. As you said, there's no reason from a software-side that they can't co-exist.

I don't know enough about the hardware to say if there's any sort of conflict in actually building an EEPROM+RTC cart.

I'm curious as to why the EverDrive X7 is implemented this way and if the 64drive hw2 has the same limitation. I'll ask Marshall and see if he has any insights.

buu342:
> I'm curious as to why the EverDrive X7 is implemented this way and if the 64drive hw2 has the same limitation.

If I remember correctly from my testing, it _should_ support both...

meeq:
Here is the test ROM that uses the header configuration for EEPROM16Kbit + RTC: https://cdn.discordapp.com/attachments/763114214688555039/862140409907380244/ctest_RTCEEPROM16k.zip

And here is that test ROM running on @networkfusion 's EverDrive X7:
image

meeq:
Update: EEPROM + RTC works on 64drive hw2

tools/n64tool.c
```diff
@@ -19,146 +19,154 @@
...
+	fprintf(stderr, "\n");
+	fprintf(stderr, "Supported cartridge save types:\n");
+	fprintf(stderr, "\tnone\t\tGame does not save or uses Controller Pak.\n");
+	fprintf(stderr, "\teeprom4k\tGame saves to 4 kilobit EEPROM.\n");
+	fprintf(stderr, "\teeprom16k\tGame saves to 16 kilobit EEPROM.\n");
+	fprintf(stderr, "\tsram128k\tGame saves to 128 kilobit SRAM\n");
+	fprintf(stderr, "\tsram256k\tGame saves to 256 kilobit SRAM\n");
+	fprintf(stderr, "\tsram768k\tGame saves to 768 kilobit SRAM\n");
+	fprintf(stderr, "\tflashram\tGame saves to 1 megabit FlashRAM\n");
+}
```
rasky, line 53: Does this still line up correctly? `\t` is 8 on most consoles, and some arguments are larger than 8 chars now.

tools/n64tool.c
```diff
@@ -183,23 +191,26 @@ int output_zeros(FILE *dest, int amount)
...
-	/* Grab number */
-	size = atoi(temp);						
+	if(arg_len < 2)
+	{
+		/* Need at least 2 characters to parse */
+		return -1;
+	}
+
+	/* Blank last character to do an atoi */
+	char * temp = strdup(arg);
```
rasky: I know it was here before, but is this needed? `atoi` stops at the first non-digit anyway, and you can use `strtol` in case you need to access the suffix.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
+			{
+				/* Expected another argument */
+				fprintf(stderr, "ERROR: Expected an argument to size flag\n\n");
+				print_usage(argv[0]);
+				return -1;
+			}
+
+			total_size = parse_bytes(argv[i++]);
+
+			if(total_size <= 0)
```
rasky: I don't know if this is the right place to check, but IPL3 also has a minimum limit of 1028 Kb which should probably be enforced.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
+			if(offset <= 0)
+			{
+				/* Invalid offset */
+				fprintf(stderr, "ERROR: Invalid offset argument\n\n");
+				print_usage(argv[0]);
+				return -1;
+			}
+
+			/* Write out needed number of zeros */
+			size_t num_zeros = offset - total_bytes;
```
rasky: This is casting to unsigned. We're lucky because `output_zeros` accepts `ssize_t`, but I think it's clearer if we define this as signed as well.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
-		
-		if(output_zeros(write_file, num_zeros))
-		{
-			fprintf(stderr, "Couldn't pad image in %s!\n", output);
-			return -1;
-		}
-		
-		fflush(write_file);
-		fclose(write_file);
+		fprintf(stderr, "ERROR: Couldn't pad image in %s!\n", output);
```
rasky: maybe add a `print_usage` here as well?

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
+		 * @see https://github.com/N64-tools/ED64/blob/develop/docs/rom_config_database.md#developer-override
+		 */
+		const char cart_id[CART_ID_SIZE] = {'E', 'D'};
+		fseek(write_file, CART_ID_OFFSET, SEEK_SET);
+		fwrite(cart_id, 1, CART_ID_SIZE, write_file);
+		fseek(write_file, VERSION_OFFSET, SEEK_SET);
+		fwrite(&config, 1, VERSION_SIZE, write_file);
+	}
+
+	fflush(write_file);
```
rasky: `fflush` is implied by `fclose`, so you can safely remove this.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
+			if(!write_file)
+			{
+				fprintf(stderr, "ERROR: Cannot open %s for writing!\n", output);
+				return -1;
+			}
+
+			ssize_t wrote = copy_file(write_file, header, byte_swap);
+
+			/* Since write file wasn't open, we haven't written the header */
+			if(wrote < 0)
```
rasky: Does this ever happen? This test baffles me. Also we're returning -1 without any explanation to the user.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
+	FILE * write_file = NULL;
+	const char * header = NULL;
+	const char * output = NULL;
+	size_t total_size = 0;
+	size_t total_bytes = 0;
+	bool byte_swap = false;
+	bool force_rtc = false;
+	bool region_free = false;
+	uint8_t save_type = SAVETYPE_NOT_SET;
+	char title[TITLE_SIZE];
```
rasky: Can we change this to `TITLE_SIZE+1`, zero-initialize it with `= {0}`, and then put also a terminating zero when a title is parsed from the command line? I know that it's not strictly required, but keeping around non-zero terminated strings is not helping other contributors on this code base.

tools/n64tool.c
```diff
@@ -209,241 +220,338 @@ int get_bytes(char *cur_arg)
...
-	int state = STATE_NONE;
-	int byte_swap = 0;
-	int i;
-	
-	/* Set default title */
+	FILE * write_file = NULL;
+	const char * header = NULL;
+	const char * output = NULL;
+	size_t total_size = 0;
+	size_t total_bytes = 0;
```
rasky: I've been confused in the past by these two variables, especially the fact that both of them get updated as we grow the output file. I think one should just be the final target size, and the other reflect the current size of the output file. Do you agree? Would you mind changing the code accordingly?

rasky (commented):
Thanks for this PR, I especially like that command line parsing is streamlined in this PR. My comments also contain suggestions for further cleanups.

tools/n64tool.c
meeq: Fixed

tools/n64tool.c
meeq: Fixed

tools/n64tool.c
meeq: Improved the readability of this check

tools/n64tool.c
meeq: Fixed

tools/n64tool.c
meeq: Fixed, also improved the messaging here.

tools/n64tool.c
meeq: Fixed.

tools/n64tool.c
meeq: Fixed

tools/n64tool.c
meeq: Excellent suggestion. The `strtol` version is much cleaner.

tools/n64tool.c
meeq: Fixed by using spaces to line things up

examples/eepromfstest/eepromfstest.c
```diff
@@ -94,7 +94,7 @@ static void print_game_high_scores(void)
 static int read_game_high_scores(char * path)
 {
     printf( "Reading '%s'\n", path );
-    const int result = eepfs_read(path, &game_high_scores);
+    const int result = eepfs_read(path, &game_high_scores, sizeof(game_high_scores));
```
rasky, line 97: This is an unrelated change, I think it should not be part of this PR

tools/ed64romconfig.c
```diff
@@ -0,0 +1,197 @@
...
+uint8_t rom_configuration(uint8_t save_type, bool force_rtc, bool region_free)
+{
+	if(!force_rtc && !region_free && save_type == SAVETYPE_NOT_SET)
+	{
+		return ROMCONFIG_NOT_SET;
+	}
+	if(save_type == SAVETYPE_NOT_SET)
+	{
+		fprintf(stderr, "WARNING: RTC/Region-Free declared without save type; defaulting to 'none'\n");
+		save_type = parse_save_type("none");
```
rasky, line 98: Seems like a convoluted way of writing `save_type = SAVETYPE_DISABLED`

tools/ed64romconfig.c
```diff
@@ -0,0 +1,197 @@
...
+			continue;
+		}
+		if(i < argc)
+		{
+			fprintf(stderr, "ERROR: Unexpected extra arguments\n\n");
+			print_usage(argv[0]);
+			return -1;
+		}
+
+		write_file = fopen(arg, "r+b");
```
rasky, line 165: It looks like this could get executed multiple times in a loop. Maybe it should instead be detected as a command line syntax error

tools/ed64romconfig.c
```diff
@@ -0,0 +1,197 @@
...
+	uint8_t save_type = SAVETYPE_NOT_SET;
+	uint8_t config = ROMCONFIG_NOT_SET;
+	int i = 1;
+	const char * arg;
+
+	if(argc <= 1)
+	{
+		/* No way we can have just one argument or less */
+		print_usage(argv[0]);
+		return -1;
```
rasky, line 121: I think in general on UNIX it is more customary to use positive error values for processes. I'd change this and the others to 1.

tools/ed64romconfig.c
```diff
@@ -0,0 +1,197 @@
...
+ */
+uint8_t rom_configuration(uint8_t save_type, bool force_rtc, bool region_free)
+{
+	if(!force_rtc && !region_free && save_type == SAVETYPE_NOT_SET)
+	{
+		return ROMCONFIG_NOT_SET;
+	}
+	if(save_type == SAVETYPE_NOT_SET)
+	{
+		fprintf(stderr, "WARNING: RTC/Region-Free declared without save type; defaulting to 'none'\n");
```
rasky, line 97: Do we really need this warning? By looking at the command line syntax, I would say it's pretty clear the intended behavior. If I asked for just RTC, I probably meant it.

rasky (approved):
Great work! I think this is good to go

examples/eepromfstest/eepromfstest.c
meeq: It's a minor change; I included it because the example was broken, and I needed to test each example's Makefile. The size parameter of of eepfs_read and eepfs_write was a last-second addition to the PR that introduced this example.

tools/ed64romconfig.c
meeq: Oh, that's a relic from before I introduced the `#define` constants -- you're right that it's a sub-optimal way to set that constant.

tools/ed64romconfig.c
meeq: The if block directly above aborts the command if there are any more arguments that would cause the write file to be opened more than once, but I could make this more obvious

anacierdem:
@meeq feel free to open a new pr for any remaining changes you have.

PR #154, opened 2021-07-09
Implement cartridge save type detection

meeq:
* Refactor DMA functions to support SRAM/FlashRAM
  * Note that `dma_read`/`dma_write` have been deprecated and are now aliases to `cart_rom_read`/`cart_rom_write`. This was done so that existing code that depends on DMA transfers would not break.
  * There is now a lower-level `pi_dma_read`/`pi_dma_write` which abstracts the DMA transfer procedure without assuming an address domain.
* Expand ctest to include save type detection
* This branch does not support reading or writing FlashRAM, only detecting and identifying the chip.
  * Additional research is required to handle writes correctly, as there are various differences between the FlashRAM chips.
  * I do not have complete information on the different FlashRAM chips out there, specifically the identifiers for the two different chips labelled `29L1100KC-15B0`. This is a best-effort based on the source code for various emulators, forum posts, Discord conversations, and the firmware for the Sanni Cart Reader. 

Test ROM: [ctest.zip](https://github.com/DragonMinded/libdragon/files/6909987/ctest.zip)

**Note that this test ROM will crash cen64 with a segmentation fault.** See https://github.com/n64dev/cen64/pull/204 for a fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,7 @@ install: libdragon.a libdragonsys.a
 	install -m 0644 include/cop0.h $(INSTALLDIR)/mips64-elf/include/cop0.h
 	install -m 0644 include/cop1.h $(INSTALLDIR)/mips64-elf/include/cop1.h
 	install -m 0644 include/interrupt.h $(INSTALLDIR)/mips64-elf/include/interrupt.h
+	install -m 0644 include/cart.h $(INSTALLDIR)/mips64-elf/include/cart.h
 	install -m 0644 include/dma.h $(INSTALLDIR)/mips64-elf/include/dma.h
 	install -m 0644 include/dragonfs.h $(INSTALLDIR)/mips64-elf/include/dragonfs.h
 	install -m 0644 include/audio.h $(INSTALLDIR)/mips64-elf/include/audio.h
--- a/examples/ctest/ctest.c
+++ b/examples/ctest/ctest.c
@@ -37,10 +37,12 @@ const char * format_save_type( cart_save_type_t save_type )
             return "256 kilobit SRAM detected";
         case SAVE_TYPE_SRAM_768KBIT:
             return "768 kilobit SRAM detected";
-        case SAVE_TYPE_SRAM_1MBIT:
-            return "1 megabit SRAM detected";
         case SAVE_TYPE_FLASHRAM_1MBIT:
             return "1 megabit FlashRAM detected";
+        case SAVE_TYPE_SRAM_1MBIT_CONTIGUOUS:
+            return "1 megabit contiguous SRAM detected";
+        case SAVE_TYPE_SRAM_1MBIT_BANKED:
+            return "1 megabit banked SRAM detected";
         default:
             return "Unknown cartridge save type";
     }
--- a/examples/ctest/ctest.c
+++ b/examples/ctest/ctest.c
@@ -23,53 +23,50 @@ const char * format_type( int accessory )
     }
 }
 
-const char * format_save_type( cart_save_type_t save_type )
-{
-    switch(save_type)
-    {
-        case SAVE_TYPE_NONE:
-            return "No cartridge save detected";
-        case SAVE_TYPE_EEPROM_4KBIT:
-            return "4 kilobit EEPROM detected";
-        case SAVE_TYPE_EEPROM_16KBIT:
-            return "16 kilobit EEPROM detected";
-        case SAVE_TYPE_SRAM_256KBIT:
-            return "256 kilobit SRAM detected";
-        case SAVE_TYPE_SRAM_768KBIT:
-            return "768 kilobit SRAM detected";
-        case SAVE_TYPE_FLASHRAM_1MBIT:
-            return "1 megabit FlashRAM detected";
-        case SAVE_TYPE_SRAM_1MBIT_CONTIGUOUS:
-            return "1 megabit contiguous SRAM detected";
-        case SAVE_TYPE_SRAM_1MBIT_BANKED:
-            return "1 megabit banked SRAM detected";
-        default:
-            return "Unknown cartridge save type";
-    }
-}
-
 const char * format_flashram_type( flashram_type_t flashram )
 {
     switch( flashram )
     {
         case FLASHRAM_TYPE_NONE:
-            return "FlashRAM not detected";
+            return "not detected";
         case FLASHRAM_TYPE_MX29L0000:
-            return "FlashRAM identified: Macronix MX29L0000";
+            return "identified: Macronix MX29L0000";
         case FLASHRAM_TYPE_MX29L0001:
-            return "FlashRAM identified: Macronix MX29L0001";
+            return "identified: Macronix MX29L0001";
         case FLASHRAM_TYPE_MX29L1100:
-            return "FlashRAM identified: Macronix MX29L1100";
+            return "identified: Macronix MX29L1100";
         case FLASHRAM_TYPE_MX29L1101_A:
-            return "FlashRAM identified: Macronix MX29L1101";
+            return "identified: Macronix MX29L1101";
         case FLASHRAM_TYPE_MX29L1101_B:
         case FLASHRAM_TYPE_MX29L1101_C:
-            return "FlashRAM identified: 29L1100KC-15B0 (MX29L1101 compatible)";
+            return "identified: 29L1100KC-15B0 (MX29L1101 compatible)";
         case FLASHRAM_TYPE_MN63F81MPN:
-            return "FlashRAM identified: Matsushita MN63F81MPN";
+            return "identified: Matsushita MN63F81MPN";
         default:
-            return "Unknown FlashRAM detected";
+            return "not identified";
     }
+}
+
+void print_save_type( uint8_t save_type, flashram_type_t flashram )
+{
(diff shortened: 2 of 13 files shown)


meeq:
According to fpgashibe the 768K EEPROM uses 3 SRAM chips bank switched with a logic circuit, so it is not actually a contiguous address space.

Additional research is required to determine how this actually works.

meeq:
I have updated this branch based on the DMA commands used by Dezaemon 3D, cross-referenced with the Verilog used by 64drive.

Bits 19:18 of the DMA cart address are used to select from the three SRAM chips used in the 768Kbit SRAM configuration

jago85:
In case you want to go further with the FlashRAM:

I've had a deeper look at the behaviour of FlashRAM chips and the games at very low level. As I've developed an own simple flash cart (with FlashRAM support) I could use an internal logic analyzer to look at the bus accesses. Also, I did a lot of tests on real FlashRAM chips accessing them using a microcontroller. I have the following chips here: MXL1101, MXL1100 and MN63F81MPN. 

Most emulators interpret the FlashRAM erase and write commands different than the real chips do. Someday I found the so called hkz-libn64. There are a lot of similarities to the emulator implementation. Maybe the emulators implemented the Flash based on this or they simply copied the stuff from each other.

For example the so called EXECUTE command doesn't execute the procedures at all. Games send this command on boot and after erase or write is finished. Don't know what this command is good for, it might be a sort of clean up. Erase process is started with ERASE_MODE command, write process is started with WRITE_OFFSET command. Then, the games read the status register until the procedure is finished (and send this EXECUTE command).

There is also a FULL_CHIP_ERASE command (0x3C000000, must be followed by ERASE_MODE command).

The different flash types return different status codes while writing/erasing and when finished. Also, some Flash types need to get the initial address right shifted by one position.

I have some notes here, but I had to translate them to english. Let me know, if it would be of any use for you.

rasky:
EXECUTE could be what it's usually called COMMIT: it's similar to a lower-level fsync for flash chips. It's required because many chips have an internal cache so writes are usually cached for a while (to speed up, avoid tearing, etc.) until the cache is full or a COMMIT arrives.

jago85:
I experimented a lot with the N64 chips. This command isn't needed at all. The FlashRAM has a 128 byte write buffer which is transfered immediately to the addressed position after the WRITE_OFFSET command. The chips don't accept any command in this state but a Read Status Register until the procedure is finished. Then you can turn off the power and the written data survives without ever sending the EXECUTE command.

But I think I've found the solution: It's the **Read/Write Status Register** command. http://n64devkit.square7.ch/n64man/os/osFlashClearStatus.htm
You can even find the function and the command code 0xd2000000 in the source code leak from last year. (Don't want to link to it, but Google easily finds it online.)
So, if an erase/write procedure fails they could set a bit. To reset the status bits a Clear Status Register command must be sent.

Little correction: The command makes the Status Register accessible. The game then writes 0 to it to clear the status.

meeq:
Marshall has informed me that the 64drive wraps SRAM addresses, which means that the test ROM incorrectly detects 1Mbit SRAM when 768K SRAM is enabled. I need to adjust the detection routines to ensure that reads and writes actually end up at their intended address.

@networkfusion has run the latest test ROM on EverDrive X7 with the latest firmware, and the results are odd:
* Setting the save type to `SRAM 256Kbit` detects 256Kbit, 768Kbit (banked), 768Kbit (contiguous), and 1Mbit (banked) capacities.
* Setting the save type to `SRAM 768Kbit` detects 256Kbit, 768Kbit (banked) and 768Kbit (contiguous) capacities.
* Setting the save type to `SRAM 1Mbit` detects  256Kbit, 768Kbit (banked), 768Kbit (contiguous), 1Mbit (banked), and 1Mbit (contiguous) capacities.

I am very curious to see if the EverDrive also does address wrapping.

meeq:
I have removed support for detecting and using so-called "contiguous SRAM" configurations because as far as I know they are not "canonical" and should be treated as an emulation accuracy bug/quirk. If homebrew developers start to rely upon this behavior, they may find themselves with compatibility issues across the ecosystem.

The only "accurate" way to target SRAM capacity greater than 256Kbits is with the "bank select" bits used by Dezaemon 3D. In theory this pattern supports up to 1 Megabit using four banks of SRAM, but in reality Dezaemon 3D only implemented 768 Kbits across three SRAM banks. I have retained support for the mythical fourth SRAM bank, but strongly discourage relying upon it.

rasky:
Can you clarify why a save type detection feature is useful for homebrew development? What scenario do you expect? 

Focusing on the DMA changes, can you explain the added value of having domain-specific functions? It seems to me that what they do in addition to the "raw" transfers is to mask / clamp the address. What is the situation in which those mask / clamps are required? Buggy code? Shouldn't those be asserts instead?

meeq:
@rasky The unfortunate reality of homebrew on N64 is that ROM configuration is driven by databases of Cartridge IDs and checksums.

This save detection functionality was born out of a realization that homebrew developers cannot rely upon emulators and flash carts to be correctly configured with the desired save type for their ROMs. This scenario is vastly improved in runtimes that respect the EverDrive64 ROM Header Configuration magic values, but most do not.

One simple way to ensure a good player experience for homebrew that wishes to offer save functionality is to check if the desired save type is available and show a message about not being able to save if not.

Some games may be able to compromise/adjust based on available storage:

* Offer 1 save slot for 256Kbit SRAM, 3 for 768Kbit SRAM, or 4 save slots for FlashRAM
* Ask for a Controller Pak instead
* Use different save data structures based on available capacity.
* Inform the player that they can manually set the save type or add it to the emulator/flash cart's save type database.

I'm working on an update to this branch that includes FlashRAM read/write APIs and a big refactoring of the SRAM APIs.

I want to see more homebrew developers take advantage of saves, and I want to make sure players have the best possible experience given the current circumstances.

meeq:
Closing this for now. I will re-open once I have finished refactoring these changes and incorporating the feedback.

sebastianmas95-blip:
> In case you want to go further with the FlashRAM:
> 
> I've had a deeper look at the behaviour of FlashRAM chips and the games at very low level. As I've developed an own simple flash cart (with FlashRAM support) I could use an internal logic analyzer to look at the bus accesses. Also, I did a lot of tests on real FlashRAM chips accessing them using a microcontroller. I have the following chips here: MXL1101, MXL1100 and MN63F81MPN.
> 
> Most emulators interpret the FlashRAM erase and write commands different than the real chips do. Someday I found the so called hkz-libn64. There are a lot of similarities to the emulator implementation. Maybe the emulators implemented the Flash based on this or they simply copied the stuff from each other.
> 
> For example the so called EXECUTE command doesn't execute the procedures at all. Games send this command on boot and after erase or write is finished. Don't know what this command is good for, it might be a sort of clean up. Erase process is started with ERASE_MODE command, write process is started with WRITE_OFFSET command. Then, the games read the status register until the procedure is finished (and send this EXECUTE command).
> 
> There is also a FULL_CHIP_ERASE command (0x3C000000, must be followed by ERASE_MODE command).
> 
> The different flash types return different status codes while writing/erasing and when finished. Also, some Flash types need to get the initial address right shifted by one position.
> 
> I have some notes here, but I had to translate them to english. Let me know, if it would be of any use for you.

Hi! By any chance, do you still have the hkz-libn64? I cant find it anywhere...

jago85:
> > In case you want to go further with the FlashRAM:
> > I've had a deeper look at the behaviour of FlashRAM chips and the games at very low level. As I've developed an own simple flash cart (with FlashRAM support) I could use an internal logic analyzer to look at the bus accesses. Also, I did a lot of tests on real FlashRAM chips accessing them using a microcontroller. I have the following chips here: MXL1101, MXL1100 and MN63F81MPN.
> > Most emulators interpret the FlashRAM erase and write commands different than the real chips do. Someday I found the so called hkz-libn64. There are a lot of similarities to the emulator implementation. Maybe the emulators implemented the Flash based on this or they simply copied the stuff from each other.
> > For example the so called EXECUTE command doesn't execute the procedures at all. Games send this command on boot and after erase or write is finished. Don't know what this command is good for, it might be a sort of clean up. Erase process is started with ERASE_MODE command, write process is started with WRITE_OFFSET command. Then, the games read the status register until the procedure is finished (and send this EXECUTE command).
> > There is also a FULL_CHIP_ERASE command (0x3C000000, must be followed by ERASE_MODE command).
> > The different flash types return different status codes while writing/erasing and when finished. Also, some Flash types need to get the initial address right shifted by one position.
> > I have some notes here, but I had to translate them to english. Let me know, if it would be of any use for you.
> 
> Hi! By any chance, do you still have the hkz-libn64? I cant find it anywhere...

Just found it here:
https://github.com/scummvm/dockerized-bb/tree/master/toolchains/n64/packages/hkz-libn64

sebastianmas95-blip:
Thank u so much! Do You know if its possible to run scummvm -some of them-
without a expansión pak online my nintendo 64?

jago85:
> Thank u so much! Do You know if its possible to run scummvm -some of them- without a expansión pak online my nintendo 64? 

No clue, sorry. I only found the source codes there. Maybe someone else?

sebastianmas95-blip:
No problem, you've already helped me a lot by finding they library, thanks!

PR #160, opened 2021-08-02
n64tool: misc improvements

rasky:
3 improvements in 3 different commits (cc @meeq):

* n64tool: make the -l option optional
Currently, -l is mandatory and contrary to before, n64tool errors out
correctly if the specified size is less then necessary. This means
that now clients are forced to specify a valid size even if they
don't really care about it (like most of the time when developing on
an emulator).
This is a nuisance that is better solved by making the size optional,
and let n64tool build a file as big as necessary, which is what most
users would expect by default.

* n64tool: declared size must be a multiple of 4.
This is required by IPL3 (used in the standard header)

* n64tool: generate a temporary file while working
If n64tool aborts at any point, it can currently leave a partial output
file on the disk. This might confuse users that don't read error messages,
and also breaks standard build system like GNU make that will believe
that the output file has been already generated.
Fix this by going through a temporary file, which is renamed to the
final name in case of success.

### DIFF
--- a/tools/n64tool.c
+++ b/tools/n64tool.c
@@ -123,11 +123,6 @@ ssize_t copy_file(FILE * dest, const char * file)
 
 ssize_t output_zeros(FILE * dest, ssize_t amount)
 {
-	if((amount & 3) != 0)
-	{
-		/* Don't support odd word alignments */
-		return -1;
-	}
 	if(amount < 0)
 	{
 		/* Can't backward seek */
@@ -281,6 +276,10 @@ int main(int argc, char *argv[])
 				fprintf(stderr, "ERROR: Invalid size argument; must be a multiple of 4 bytes\n\n");
 				return print_usage(argv[0]);				
 			}
+			if (size % 512 != 0)
+			{
+				fprintf(stderr, "WARNING: Sizes which are not multiple of 512 bytes might have problems being loaded with a 64drive\n\n");
+			}				
  
 			declared_size = size;
 			continue;
@@ -313,6 +312,11 @@ int main(int argc, char *argv[])
 				/* Invalid offset */
 				fprintf(stderr, "ERROR: Invalid offset argument\n\n");
 				return print_usage(argv[0]);
+			}
+			if((offset % 4) != 0)
+			{
+				fprintf(stderr, "ERROR: Invalid offset argument (must be multiple of 4)\n\n");
+				return print_usage(argv[0]);				
 			}
 
 			/* Write out needed number of zeros */
@@ -421,6 +425,17 @@ int main(int argc, char *argv[])
 		if(output_zeros(write_file, num_zeros))
 		{
 			fprintf(stderr, "ERROR: Couldn't pad %zu bytes to %zu bytes.\n", total_bytes_written, declared_size);
+			return print_usage(argv[0]);
+		}
+	}
+	else if((total_bytes_written % 512) != 0)
+	{
+		/* Pad size to 512 bytes to avoid possible incompatibilities with 64drive
+		   due to a firmware bug (present as of firmware rev 2.05) */
+		ssize_t num_zeros = 512 - (total_bytes_written % 512);
+		if(output_zeros(write_file, num_zeros))
+		{
+			fprintf(stderr, "ERROR: Couldn't pad %zu bytes to %zu bytes.\n", total_bytes_written, total_bytes_written+num_zeros);
 			return print_usage(argv[0]);
 		}
 	}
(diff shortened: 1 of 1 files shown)


rasky:
Found a bug: we want to pad to MIN_SIZE if the size is smaller, but at the same give an error if the user declared a size and we went bigger. I'll fix this.

rasky:
done, ready for review

tools/n64tool.c
```diff
@@ -245,6 +253,9 @@ int main(int argc, char *argv[])
 				fprintf(stderr, "WARNING: The output should have a '.z64' file extension\n");
 			}
 
+			tmp_output = malloc(output_len + 4 + 1);
```
meeq: ~What's the rationale for this? If these numbers are meaningful, it might be better for readability to declare them as constants (or at least throw in a comment explaining what's happening here).~

meeq: Oh, I get it: this is for `.tmp` -- should have read a bit further into it. 😄

meeq: Suggestion: Add `*.tmp`  to `.gitignore` under `## Build files`

tools/n64tool.c
```diff
@@ -264,7 +275,13 @@ int main(int argc, char *argv[])
 				fprintf(stderr, "ERROR: Invalid size argument; must be at least %d bytes\n\n", MIN_SIZE);
 				return print_usage(argv[0]);
 			}
-
+			if (size % 4 != 0)
```
meeq, line 271: Did you want to also throw in a warning for 64drive wanting size to be a multiple of 512?

tools/n64tool.c
```diff
@@ -384,22 +404,37 @@ int main(int argc, char *argv[])
...
 	/* Set title in header */
 	fseek(write_file, TITLE_OFFSET, SEEK_SET);
 	fwrite(title, 1, TITLE_SIZE, write_file);
 
 	/* Sync and close the output file */
 	fclose(write_file);
 
+	/* Rename to the final name */
+	if(rename(tmp_output, output) != 0)
+		perror("cannot rename output file");
```
meeq: This seems to be a "not-OK" scenario; should this `return STATUS_ERROR;`?

meeq: This error message makes sense to you, but would leave an end-user clueless. I suggest: `"Couldn't rename temporary output file '%s' to '%s'", tmp_output, output`

tools/n64tool.c
meeq: Also, add `*.tmp` to the Makefile `clean` targets for all the examples.

Alternatively, consider using ISO C [`tmpfile` or `mkstemp` to manage the temporary file](https://yakking.branchable.com/posts/temporary-file-handling/) and copy/rename it to output at the end.

meeq: If you don't decide to use `mkstemp`: consider changing this line to `asprintf(tmp_output, "%s.tmp", output)` for better readability

rasky: > Suggestion: Add *.tmp to .gitignore under ## Build files

It's already in there, albeit in a different section

> Also, add *.tmp to the Makefile clean targets for all the examples.

Do we really need it, given that I also registered deletion in `atexit`? I would say that it should be a very rare occurrence (eg: a segfault). Also, they're ignored anyhow, it's not that they confuse the Makefile by creating a false dependency.

> If you don't decide to use mkstemp: consider changing this line to asprintf(tmp_output, "%s.tmp", output) for better readability

Done.

tools/n64tool.c
rasky: We could do it, but I'm on the fence because it would be spam for users of any other dev cartridge, and for users of 64drive via both g64drive and UNFLoader. Opinions?

rasky (commented):
Thanks @meeq I've implement the review and left a few comments

tools/n64tool.c
meeq: I think we should be encouraging developers to create ROMs that are as widely-compatible as possible. A warning would be helpful for people who don’t know the quirks, and could just be ignored by people who don’t care.

rasky: OK done. But given your concern, I did something more: I made sure n64tool never generates a ROM which is not multiple of 512 bytes, even when the user has not declared explicitly a size (since it's not mandatory anymore). See the followup commit.

rasky:
@meeq I've also added one more commit that make the size default to bytes, rather than force the user to explicitly use the suffix `B`. I think it's a reasonable default and less surprising than erroring out if "B" is not specified.

meeq:
@anacierdem This looks alright to me. 👍🏻

Issue #76, opened 2019-11-03
Entrypoint only supports CIC 6102


anacierdem:
It always uses `0x80000318`. Also see `system.c` which uses both addresses `0xA0000318` and `0xA00003F0`.

rasky:
This is not true anymore in the unstable branch, as libdragon only supported the new open source IPL3, whose API is not CIC-dependent.

rasky:
IPL3 was merged to trunk too, so we now support all CIC types transparently (with the same IPL3).

PR #163, opened 2021-08-24
Cleanup and minor fixes

### LINKED ISSUE #87: We don't need version numbers anymore

### LINKED ISSUE #139: Disable interrupts on cache invalidation test - and maybe others?
When the interrupts are enabled, e.g the VI, unrelated code gets executed potentially invalidating the cache and breaking the test.
I can swear I have seen a case where there was garbage data in the asserted memory region but I was not able to reproduce it, so take that with a grain of salt.

Additionally, we may consider disabling interrups for other tests as well, especially the ones with the benchmarks, as they consistently fail for me on an everdrive whereas they pass on @rasky's 64drive.

anacierdem:
- Remove version numbers, closes DragonMinded#87
- Streamline testrom Makefile
- Disable interrupts when testing the cache, fixes DragonMinded#139
- Evict cache before each test
  - Add `data_cache_writeback_invalidate_all` and `inst_cache_invalidate_all` functions.
- Add missing 9th register to exception test
- Do cleanup via DEFER in test_exception
- Reduce testrom size to 2M
- Add new linker symbol for libdragon text location including the offset


tests/Makefile
```diff
@@ -1,55 +1,55 @@
+PROG_NAME = testrom
```
rasky, line 1: I'm going to submit the new build system this week (it's the first PR of the series of the audio library, given that the new build system contains RSP compiling rules). 

It doesn't matter whether this change is committed or not before that, but I'm skipping reviewing it as it would be overhauled anyway

tests/test_cache.c
```diff
@@ -1,5 +1,8 @@
 
 void test_cache_invalidate(TestContext *ctx) {
+	disable_interrupts();
+	DEFER(enable_interrupts());
```
rasky, line 6: Maybe leave a comment here on why we need this. Does this fix the failures with everdrive BTW?

tests/testrom.c
```diff
@@ -243,6 +246,10 @@ int main() {
 		fflush(stdout);
 		debugf("**** Starting test: %s\n", tests[i].name);
 
+		// Do a complete cache flush before running each test
+		data_cache_hit_writeback_invalidate(MEM_START, MEM_LENGTH);
+		inst_cache_hit_invalidate(MEM_START, MEM_LENGTH);
```
rasky: I think it would be better to abstract this in proper functions: data_cache_writeback_invalidate_all() and inst_cache_invalidate_all().

The best implementation would be to use the "cache by index" family of cache opcodes that manipulates the cache using the internal index rather than the memory address. Then you iterate the index on all the available slots (need to check how many cache slots are in VR4300) so that you fish everything with maybe 64 or 128 iterations rather than several thousands like in this implementation.

If you don't want to research this, it's OK to add the new API for now with this implementation and leave a TODO for later.

tests/testrom.c
```diff
@@ -8,6 +8,9 @@
 #define IN_EMULATOR  0
 #endif
 
+#define MEM_START ((void *)0x80000400)
+#define MEM_LENGTH (1024 * 1024 * 4)
```
rasky: I think entrypoint.S saves the correct size in a global variable you could access, so that this is correct even with the memory pak

rasky (commented):
Thanks for the cleanups!

tests/test_cache.c
anacierdem: I don't have a failing test at the moment unfortunately so I cannot verify if it totally fixes them. I'll keep an eye though.

tests/testrom.c
rasky: Actually, it is system.c that does this:
https://github.com/DragonMinded/libdragon/blob/452aa631a5f93a5994faaa6553797d8f99a37bac/src/system.c#L1009-L1013

and uses `get_memory_size` which is here:
https://github.com/DragonMinded/libdragon/blob/1c522f2da5928bb5bb79ac76436c2186091fef75/src/n64sys.c#L172-L175

This takes into account also the CIC being used (while entrypoint.S doesn't... we have an open issue of that)

tests/testrom.c
anacierdem: Done without index ops for now.

tests/testrom.c
anacierdem: I have made it an absolute linker symbol and also fixed the other bug in the process. Do you have any better ideas for keeping the memory start location?

anacierdem:
I think you can have another look now.

n64.ld
```diff
@@ -22,22 +22,24 @@ OUTPUT_ARCH (mips)
 EXTERN (_start)
 ENTRY (_start)
 
+__mem_start = 0x80000000;
```
rasky: I'm not 100% won on this new symbol ... I think in a way we hide in clarity because if I see `mem_start` in C code I might be not sure what it means (it's not a standard symbol, so it's something libdragon specific I have too lookup, and it's even defined in .ld so it's also non-trivial to find). On the other hand, `0x80000000` is very clear to whoever should be able to touch and read this code.

n64.ld
```diff
@@ -22,22 +22,24 @@ OUTPUT_ARCH (mips)
...
+
+/* Start address of code is 1K up in uncached, unmapped RAM.  We have
+* to be at least this far up in order to not interfere with the cart
+* boot code which is copying it down from the cart
+*/
+__libdragon_code_start = __mem_start + 0x400;
```
rasky: This instead makes sense to have as a linker-defined symbol after all. Maybe `__libdragon_text_start` is a little bit more clear, given that we're speaking of a linker symbol?

src/entrypoint.S
```diff
@@ -9,9 +9,9 @@
 	.section .boot
 	.global _start
 _start:
-	lw t0, 0x80000318			/* memory size */
-	li t1, 0x7FFFFFF0
-	addu sp,t0,t1				/* init stack */
+	jal get_memory_size
```
rasky: We can't call C code from here, we have not yet initialized SP and GP, and both of them are necessary to be able to call arbitrary C code.

Also, I'm not sure what it is the benefit. The idea of `get_memory_size` is that someone would call `sys_set_boot_cic` before calling it, and surely nobody can call that before `_start`. 

I don't like very much the `sys_set_boot_cic` thing, specifically because it doesn't solve stack placement (which is #76). I think it's been a mistake to introduce a runtime API. It should have been a compile-time option. Or if we want to do a runtime thing, it must be an autodetection in `entrypoint.S`. For instance, the code could checksum the IPL3 boot code (0x1000_0040 - 0x1000_1000) [like emulators do](https://github.com/rasky/r64emu/blob/4902c2b942ad1be36fd7bb662c93381fe10d5566/src/cartridge.rs#L64-L72) to try to guess the correct CIC model.

So, unless you want to start a new adventure, I'd remove this commit.

src/n64sys.c
```diff
@@ -26,6 +28,8 @@
  * @{
  */
 
+extern volatile void __mem_start;
```
rasky: If you decide to keep it, please add a comment here that says that it's defined in `n64.ld` as `0x80000000`, so at least it's easier for readers.

Also I don't think it has to be `volatile`.

rasky (changes requested):
Thanks. I think the `entrypoint.S` modification is wrong (see comment), everything else looks OK.

n64.ld
anacierdem: There is also a `__text_start` that was the reason for the name but that is fine as well. I'll update.

src/entrypoint.S
anacierdem: Haha you are right, I bodged this in the last minute 😅 sp is easy to initialize once more before calling it as we know the minimum memory amount and we can also prepare gp but I didn't know `get_memory_size` depended on user-set info either so this does not make sense. I will remove this nonsense 😊
Even the invalidate all functions are rather useless 🤷🏻‍♂️

n64.ld
anacierdem: I totally agree and still I couldn't find a better way. Let me try and think about hardcoded alternative once more...

src/n64sys.c
anacierdem: Sure I will do. Better to cast it to volatile later (cache invalidate funcs expect a volatile param)

n64.ld
anacierdem: I kept the linker symbol but only for linker script use. I defined a new `KSEG0_START_ADDR` for C code.

rasky (approved):
Looks fine to me! I've also checked the test ROM after the dfs_read timing changes and it works for me too. I hope that flushing cache is enough to definitely fix timing flakiness.

PR #164, opened 2021-08-25
Introduce new build system and use it for libdragon itself.

rasky:
This PR introduces a new libdragon build system "n64.mk". It is designed
to be used to build final ROMs in an easy way, but it can be used
for libdragon itself.

The main features of the build system:

 * Installed with libdragon, so can be used with "include n64.mk"
 * Minimal Makefile to be written for ROMs
 * Support building C files, CPU assembly files, RSP ucodes, DragonFS
 * Automatic C dependency generation: touch a .h -> recompile automatically affected files
 * Support optional out-of-tree builds
 * Support verbose mode (make V=1)
 * Support generation of debugging symbols (make D=1)
 * Clean minimal output

In addition of converting libdragon itself to it (which is sort of
a special case, because no ROM is built), the spritemap example is
also converted to it, to show how a typical ROM Makefile would look
like.


rasky:
One opinionated choice which is hidden here: since the build system needs to build both MIPS assembly code (like `entrypoint.S`) and RSP microcode (coming soon...), we need a way to differentiate between the two. Rather than coming up with a made up extension for RSP microcode, I preferred to keep everything under `.S`, and use the filename prefix instead: if a file is called `rsp*.S`, it is built as RSP microcode. 

This is maybe a little unconventional, but works for me and it's easy to explain and use.

anacierdem:
This looks good to me. I will give this a try to see how it works in practice. It might also be a good idea to migrate the ucode example.

rasky:
> It might also be a good idea to migrate the ucode example.

I did this, and it was a deep hole to dig :) Found out that the test is broken and has been broken since we added asserts for catching wrong usages of `data_cache_invalidate`; `read_data` / `read_ucode` was using that function incorrectly and tripped on the assert.

But that was hard to find out because we had left a non-implemented feature: asserts were not visible when interrupts were disabled. So basically the ucodetest would freeze with no indication of the problem.

I've now pushed a fix to show asserts with interrupts disabled, and to rsp.c to use the correct DMA cache ops. I've pushed conversion of ucodetest to the new build system as well.

Moreover, I've added two followups to the build system that I had around:

 * Show stats of the ELF size by default (code / data size are very important for console developers)
 * Make an explicit dependency on the installed version of libdragon, so that ROMs get rebuilt automatically when the installed version changes, which is very useful for development also of tests/examples (change something in libdragon, recompile the example, everything gets recompiled).

src/console.c
```diff
@@ -263,8 +263,7 @@ static void __console_render()
 
             if(t_buf == 0)
             {
-                display_show(dc);
-                return;
+                goto end;
```
anacierdem, line 266: Why can't we just do the conditional force here?

n64.mk
```diff
@@ -0,0 +1,108 @@
...
+		echo "    [AS] $<"; \
+		$(CC) -c $(ASFLAGS) -o $@ $<; \
+	fi
+
+$(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c 
+	@mkdir -p $(dir $@)
+	@echo "    [CC] $<"
+	$(CC) -c $(CFLAGS) -o $@ $<
+
+%.elf: $(N64_ROOTDIR)/mips64-elf/lib/libdragon.a $(N64_ROOTDIR)/mips64-elf/lib/libdragonsys.a
```
anacierdem: What about adding the linker script to the dependencies as well? This was already like that in my setup to be able to quickly test layout changes as well. I was able to do it like this;
```
%.elf: $(N64_ROOTDIR)/mips64-elf/lib/libdragon.a $(N64_ROOTDIR)/mips64-elf/lib/libdragonsys.a $(N64_ROOTDIR)/mips64-elf/lib/n64.ld
	@mkdir -p $(BUILD_DIR)
	@echo "    [LD] $@"
	$(LD) -o $@ $(filter-out $(N64_ROOTDIR)/mips64-elf/lib/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$@.map
	@echo "    [STATS] $@"
	$(N64_SIZE) -G $@
```
There may be a better way though...

examples/spritemap/Makefile
```diff
@@ -1,39 +1,18 @@
...
-	$(N64TOOL) $(N64TOOLFLAGS) -o $@ $(PROG_NAME).bin -s 1M $(PROG_NAME).dfs
-	$(CHKSUM64) $@
-
-$(PROG_NAME).bin: $(PROG_NAME).elf
-	$(OBJCOPY) $< $@ -O binary
-
-$(PROG_NAME).elf: $(PROG_NAME).o
-	$(LD) -o $@ $^ $(LDFLAGS)
+clean:
+	rm -f $(BUILD_DIR)/* spritemap.dfs spritemap.elf spritemap.z64
```
anacierdem: What do you think of putting all `elf`, `dfs` and `z64` in the `BUILD_DIR`?

anacierdem (commented):
This is awesome! Thank you @rasky!

I have a few comments that might be useful.

src/console.c
rasky: I didn't want to duplicate the conditional logic and the comment both here and at the end of the function (we need it in both places).

n64.mk
rasky: That's great, I'll add it

examples/spritemap/Makefile
rasky: We can do that. In general I like that the main output (z64) is in the source dir rather than hidden somewhere (I hate when I build something and then I need to dig around to find the binary), and at that point I kept the other mentioned files there for simplicity. I'm not strong on this though.

anacierdem: Maybe we can leave the rom file at root but hide others. Nothing I will insist on either, just an idea

rasky: OK done.

anacierdem:
An afterthought: we should be able to move `exception_halt` function's contents into the default handler as it does not need interrupts anymore. If it simply works like that it'd be nice to do in this pr as well. If it's getting complicated, I'm ok as is.

rasky:
> An afterthought: we should be able to move exception_halt function's contents into the default handler as it does not need interrupts anymore. If it simply works like that it'd be nice to do in this pr as well. If it's getting complicated, I'm ok as is.

Done as well. I had to change the "interrupt disabled" check in console_render because I need to also check for the exception status (on MIPS you can technically have interrupts enabled but not triggering because you are in an exception).

PR #165, opened 2021-08-27
Sync usb.c and add timeout for everdrive busy wait

anacierdem:
Also opened [this PR](https://github.com/buu342/N64-UNFLoader/pull/65) for UNFLoader.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ INSTALLDIR = $(N64_INST)
 libdragon: CC=$(N64_CC)
 libdragon: AS=$(N64_AS)
 libdragon: LD=$(N64_LD)
-libdragon: CFLAGS+=$(N64_CFLAGS) -I$(CURDIR)/include -I$(ROOTDIR)/mips64-elf/include
-libdragon: ASFLAGS+=$(N64_ASFLAGS) -I$(CURDIR)/include -I$(ROOTDIR)/mips64-elf/include
+libdragon: CFLAGS+=$(N64_CFLAGS) -I$(CURDIR)/include
+libdragon: ASFLAGS+=$(N64_ASFLAGS) -I$(CURDIR)/include
 libdragon: LDFLAGS+=$(N64_LDFLAGS)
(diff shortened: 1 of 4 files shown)


rasky (approved):
OK, but I'd wait for upstream to review and approve first.

anacierdem:
I re-synced but there is the upstream alignment change now. Should we wait for a proper fix or go ahead? We might de-vendor that file and fix as we need as well (possibly also removing libultra parts)

Makefile
```diff
@@ -87,7 +90,7 @@ install: libdragon.a libdragonsys.a
 	install -Cv -m 0644 include/libdragon.h $(INSTALLDIR)/mips64-elf/include/libdragon.h
 	install -Cv -m 0644 include/ucode.S $(INSTALLDIR)/mips64-elf/include/ucode.S
 
-clean:
+clean: install-mk
```
rasky: This one isn't really required, is it?

rasky (commented):
Upstream has reverted the problematic commit. I think you can sync with their trunk if you wish so.

Makefile
anacierdem: It is not required for the clean recipe but for `examples-clean` as the child makefiles does not have the file before anything is installed.

anacierdem:
I have also opened [this](https://github.com/buu342/N64-UNFLoader/pull/69).

Regarding Makefile changes:
- The first problem was that the root `install` recipe was only
depending on `libdragon.a` and `libdragonsys.a`, not the complete chain.
This is fixed by making install depend on `libdragon` instead.
- Second problem; example clean recipes using the include are unable to
find it when the parent makefile (`clobber` recipe) is executed. The
root Makefile has access to the build system but children does not until
it is installed. This was also causing `build.sh` to fail.
- I have also changed the sub make commands into `$(MAKE)` such that
make can know they are actually make commands. Also see [this](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable)

PR #166, opened 2021-08-31
Refactoring RSP support in libdragon and add first RSP code library

rasky:
This PR is merged of 4 commits that can be reviewed independently:

 * Rename ucode.S to rsp.inc
 * Refactoring of rsp.h / rsp.c
 * Add vector shifts pseudo-op and other macros to rsp.inc to improve RSP ucode writing
 * Add RSP DMA ucode library


include/rsp.h
```diff
@@ -10,12 +10,206 @@
...
+ * This function starts running the RSP, and wait until the ucode is finished.
+ */
+void rsp_run(void);
+
+/** @brief Run RSP async.
+ * 
+ * This function starts running the RSP in background. Use rsp_wait() to
+ * synchronize later.
+ */
+void rsp_run_async(void);
```
danbolt, line 127: I love having the option between the two here. :+1:

danbolt:
Similar to the discussion about the kernel implementation, I do really like that these interfaces are much simpler than libultra's. A starting homebrew developer could get their bearings quickly.

rasky:
@danbolt thanks!

include/rsp.h
```diff
@@ -10,12 +10,206 @@
...
+ * from RDRAM to IMEM. Prefer using rsp_load instead.
+ * 
+ * @param[in]     code          Pointer to buffer in RDRAM containing code. 
+ *                              Must be aligned to 8 bytes.
+ * @param[in]     size          Size of the code to load. This is rounded up
+ *                              to 8 bytes.
+ * @param[in]     imem_offset   Byte offset in IMEM where to load the code.
+ *                              Must be a multiple of 8.
+ */
+void rsp_load_code(void* code, unsigned long size, unsigned int imem_offset);
```
danbolt: If a programmer was looking to do something like a microcode overlay, would using `imem_offset` be how they do it?

src/rsp.c
```diff
@@ -56,93 +21,115 @@ static void __SP_DMA_wait(void)
...
+    assert(((uint32_t)start % 8) == 0);
+    assert((dmem_offset % 8) == 0);
+
     disable_interrupts();
     __SP_DMA_wait();
 
     SP_regs->DRAM_addr = start;
     MEMORY_BARRIER();
-    SP_regs->RSP_addr = (void*)SP_DMA_DMEM;
+    SP_regs->RSP_addr = SP_DMEM + dmem_offset/4;
```
danbolt, line 68: Since `dmem_offset` is unsigned, would there be any rounding problems using `>> 2` instead of integer division here?

This might come across as a nitpick, but my rationale is that the extra cycles are given back to the person doing their homebrew project. Thoughts?

danbolt: Same goes for `imem_offset` in `rsp_load_code` too.

include/rsp.h
rasky: Possibly, yes. In general I'm adding the offset because I feel like being limited to 0 is gratuitous, especially for loading/saving data to DMEM.

src/rsp.c
rasky: Actually the division does not translate to a division opcode because it's added to a `uint32_t*` pointer, so the value in `dmem_offset` is directly added to the constant behind `SP_DMEM` (as you'd expect, since `dmem_offset` is a byte offset anyway).

Even if that wasn't the case, GCC is able to optimize `/4` into `>>2` on unsigned integers, just like it can optimize `%2` into `&1`.

include/rsp.h
```diff
@@ -10,12 +10,206 @@
...
+
+/** @brief Run RSP async.
+ * 
+ * This function starts running the RSP in background. Use rsp_wait() to
+ * synchronize later.
+ */
+void rsp_run_async(void);
+
+/** @brief Wait until RSP is idle. */
+void rsp_wait(void);
```
anacierdem, line 130: nit: `rsp_wait_halt` might be a better name b/c halt is not the only way to synchronize with RSP and one may choose to do so with a breakpoint and this might be confusing.

include/rsp.h
```diff
@@ -10,12 +10,206 @@
...
 #endif
 
-void rsp_init();
-void load_ucode(void * start, unsigned long size);
-void read_ucode(void* start, unsigned long size);
-void load_data(void * start, unsigned long size);
-void read_data(void* start, unsigned long size);
-void run_ucode();
+/** @brief RSP DMEM: 4K of data memory */
+#define SP_DMEM       ((volatile uint32_t*)0xA4000000)
```
anacierdem, line 14: Do you think we should just include KSEG1 directly in the addresses we use or should we use macros to convert physiscal addresses in general?

include/rsp.h
```diff
@@ -10,12 +10,206 @@
...
+void rsp_wait(void);
+
+/** @brief Do a DMA transfer to load a piece of code into RSP IMEM.
+ * 
+ * This is a lower-level function that actually executes a DMA transfer
+ * from RDRAM to IMEM. Prefer using rsp_load instead.
+ * 
+ * @param[in]     code          Pointer to buffer in RDRAM containing code. 
+ *                              Must be aligned to 8 bytes.
+ * @param[in]     size          Size of the code to load. This is rounded up
```
anacierdem: It is not rounding but asserting it might be nice to update the comment as well

include/rsp_dma.inc
```diff
@@ -0,0 +1,183 @@
...
+    li t2, DMA_IN
+    # fallthrough!
+    .endfunc
+
+    .set noat
+    .func DMAExec
+DMAExec:
+    # Wait for the DMA engine to be ready. We do this
+    # every time because with async transfers it's easy to get
+    # confused and forget to wait.
```
anacierdem, line 97: Might be nice to have an option to disable the check for very thight things in the future. Not much of a concern ATM.

include/rsp_dma.inc
```diff
@@ -0,0 +1,183 @@
...
+#
+# At the hardware level, both the DMEM and the RDRAM addresses are
+# treated as 8-byte aligned (that is, the lowest 3 bits are ignored).
+# If you issue a transfer from RDRAM with an unaligned address (in s0),
+# the DMEM address (s4) will be updated to reflect where the byte that
+# was actually requested ended up. For instance, if s0 = 0x80041203, s4=0x100,
+# after the function s4 will be 0x103, as that is the address where the
+# byte at 0x80041203 ended up being written to.
+#
+# INPUT:
```
anacierdem, line 46: Should we follow MIPS conventions - e.g a0-a4 for arguments etc. o/w each library might end up defining its own convention and we will need to look its description up

include/rsp_dma.inc
```diff
@@ -0,0 +1,183 @@
...
+    .func DMAExec
+DMAExec:
+    # Wait for the DMA engine to be ready. We do this
+    # every time because with async transfers it's easy to get
+    # confused and forget to wait.
+    mfc0 $1, COP0_DMA_FULL
+    bnez $1, DMAExec
+    # Merge pitch into the width/height word, converting it
+    # into "skip" (= pitch-width)
+    andi $1, t0, 0xFFFF
```
anacierdem: Wasnt the [width](https://github.com/DragonMinded/libdragon/pull/166/files#diff-0b1db597974243386a372fc3028af955f77db23942d78955b06a2bbeea8a5d6dR665) 12bits?

include/rsp.inc
```diff
@@ -0,0 +1,680 @@
...
+
+# Compose the RSP DMA size register with immediate constants
+# "width", "height", and "pitch".
+# It is similar to DMA_SIZE, but it can be used when the pitch is a compile-time
+# constant.
+#define DMA_SIZE3(width, height, pitch)  (((width)-1) | (((height)-1)<<12) | (((pitch)-(width))<<20))
+
+# DMAExec operation macros. Pass one of these in t2 to DMAExec
+#define DMA_IN_ASYNC     0x00000000
+#define DMA_OUT_ASYNC    0xFFFF8000
```
anacierdem, line 712: Was the `8000` part an arbitrary choice?

include/rsp.inc
```diff
@@ -0,0 +1,680 @@
...
+    .long (0x12 << 26 | 0x0 << 21 | \vreg << 11 | hex.\reg << 16 | \element << 7)
+.endm
+
+
+##################################################
+# Vector shift pseudo-opcodes
+##################################################
+# Helper macros for vector shifts.
+# How to use:
+#   1) Call macro "vsll_data" and/or "vsll8_data" in the data segment to define
```
anacierdem, line 465: Can't we structure this to choose between the two such that `vsll <VDST>, <VSRC>, <QTY < 8>` will do `vsll_data` + `setup_vsll` + `vsll` and `vsll <VDST>, <VSRC>, <QTY >= 8>` do `vsll8_data` + `setup_vsll8` + `vsll8`?
Usage could be simplified to `vsll <VDST>, <VSRC>, <QTY>, <VTEMP>`

anacierdem (commented):
Looks great as always! A few comments here and there.

include/rsp.h
rasky: what's the difference? `break` is the way you halt the processor from within the RSP, and the only thing it makes sense to wait is for the processor to be halted. Maybe the comment is wrong?

I think what we're going to have, at some point, is more granularity like "enqueue one task and wait for that task to be done", but since there's no API surface about this now, I think it's premature.

anacierdem: Breaking the rsp triggers the SP interrupt, so we can also wait for that when we have the kernel for example, or with a callback etc. In that case I think we will just change the implementation then? I also think its early to decide and that was the reason I wanted to reserve the name by making it more descriptive. Nothing I will insist at all though, if you think we can change the implementation without breaking assumptions of usages won't be a problem.

rasky: Yes if we implement a interrupt-driven queue, `rsp_wait()` could be changed to follow that. I think the semantic is clear. Maybe the comment should say "until the RSP has finished processing" which is more "general".

For kernelizing this function (and other similar), my plan is as follows. Current implementation:

```C
while (!(*SP_STATUS & SP_STATUS_HALTED)) {} 
```

Future implementation:

```C
while (!(*SP_STATUS & SP_STATUS_HALTED)) {   
    kthread_wait_event(&KEVENT_IRQ_SP);
} 
```

with `kthread_wait_event()` being a no-op when the kernel is not initialized.

include/rsp.h
rasky: I think in this specific case the only way to access DMEM without halting the CPU is through KSEG1. Am I wrong?

include/rsp.h
rasky: Done

include/rsp_dma.inc
rasky: It can be added as an #ifdef of course, if required. 

In my experience trying to squeeze times, it's an important safety net. Code-wise, consider that if you remove this, you might be forced to add many calls to `DMAWaitReady` instead, which takes at least 1 instruction of space in IMEM per each site.

include/rsp_dma.inc
rasky: Right, I forgot to discuss about this.

In general, all my microcode does not follow the C MIPS ABI. That's because the RSP context is very different. First of all, all code is assembly; second, there is often no stack, so that also makes it impossible to both fully follow the ABI, and to do something as simple as saving/restore registers; third, the MIPS ABI forces you to shuffle registers a lot (saving them on the stack) because it has fixed assignments.

The MIPS ABI (or any ABI, for that matter) make sense in the context of allowing foreign code to interoperate, as it is normally the case for C which is compiled in different translation units so you don't know what the other code is doing. If you were to compile the 100% of a binary from a single C file (including standard libraries, etc.) the compiler could in theory forego the ABI and find an optimal allocation for the specific problem at hand -- no compiler implements this because it is such a specialistic use case that nobody needs.

And in fact, I think we also don't need it for RSP. RSP ucode can be written using some global register allocation where you have a few "local, scratch registers" that you can use freely and can be freely clobbered, and most registers are simply "globally allocated" for large sections of the code, or even the whole microcode. If you look at SGI RSP code snippets, they don't follow the standard MIPS ABI as well.

So what I should do, instead, is to document the convention I came up with and that all my microcode follows; given that I'm going to submit that code and that any convention is good, I claim the right of first arrival here so that we can use mine without me having to rename registers in all my code (which also would be troublesome to do without introducing bugs).

My convention is:

 * microcode always starts at 0x0000 in IMEM
 * `set noreorder` by default, to squeeze every last ounce of performance
 * `set at` by default, but you can switch to `set noat` to have one more scratch register when needed (eg: very useful in macro and helpers).
 * 2 levels of nested function calls are supported; use `sp` as `ra2` to implement the two levels. (Ab)use tail calls to help with this. If you need more, try to refactor, or in the worse case, you need to allocate another register as `ra3`.
 * `t0-t3`: scratch value registers, often used also as arguments to functions. These are considered clobbered by defaults in function calls.
 * `s0-s4`: scratch pointer registers, often used also as arguments to functions. These are often clobbered by functions calls.
 * `a0-a3`: display list input to be processed (eg: the 64bit primitive being processed can go in `a0/a1`). This applies if you process some sort of display list input of course.
 * anything else: use it for keeping values around in larger sections of the code (or even the whole microcode). Rename each register with #define to begin using, and then #undef after you don't need it. Wish gas had proper register renaming, but it doesn't, so we'll use preprocessor for this.

include/rsp_dma.inc
rasky: Good catch!

It looks like I introduced this bug in Jun20 in my codebase, when I introduced async functions, with this diff:

```diff
-    andi $1, t0, 0xFFF
+    .func DMAInOut
+DMAInOut:
+    mfc0 $1, COP0_SP_STATUS
+    andi $1, SP_STATUS_DMA_FULL
+    bnez $1, DMAInOut
+    andi $1, t0, 0xFFFF
```

I have never realized it because it happens not to matter... that value is used to compute the "skip" component which is only 12-bit wide, so those last 4 bits do not affect the result and they are shifted away later with `sll $1, 20`. Nonetheless, I fixed it. Thanks!

include/rsp.inc
rasky: Semi-arbitrary. We need a 16-bit negative value that sign-extends to a 32-bit negative value, so that you can load it with non-macro `li` (which does a 16-to-32 sign extension). This basically allows to load all the `DMA_` operation macros into `t2` with a single `li` opcode.

`0x8000` would work just as `0xCCCC`, but it leaves 15 free bits to encode other things. In fact, we do encode the wait flags in the lowest bits (we reuse the same bits of the status register to simplify the checking loop).

include/rsp.inc
rasky: We cannot do that because we need the setup part to be done once (possibly, once per microcode), allocating a global vector register. Of course there might be cases where you just need to shift once and in that case you could probably do it with a do-everything macro, but in the general case it's better to separate the setup from the actual shift instruction.

include/rsp.h
anacierdem: I think that wasn't clear, this is not something technical only cosmetic.
What I mean is we can also write;
```
#define SP_DMEM_PHYS      0x04000000
#define SP_DMEM                ((volatile uint32_t*) TO_UNCACHED_KSEG1(SP_DMEM_PHYS))
```
This is a general question, how should we handle such things in libdragon? Might be nice to have them explicit. I almost always forget the memory layout and check docs.

include/rsp_dma.inc
anacierdem: I didn't mean strictly full ABI but overall conventions, this is all OK for me. Would be nice to write this somewhere though.

include/rsp.inc
anacierdem: Hmm I wonder if there is a way do something only once, it sounds possible. Regardless we can have a wrapper around it if we need that. Thx for the clarification.

include/rsp.inc
anacierdem: I have seen the part where we used the lowest bits and its ingenious but I thought `0xFFFF0000` would also work?

anacierdem: Ah sorry we use tha last 16bits 👍🏼

anacierdem:
This is ok to merge for me, anytime you like!

include/rsp_dma.inc
rasky: Done.

rasky:
Ready for commit.

PR #167, opened 2021-09-01
Remove files.in, not used anymore with new build system

rasky:
I think this used to be part of the build system PR, but I might have lost it during its evolution.


rasky:
resolved conflict, @anacierdem

PR #168, opened 2021-09-13
Lock line endings for shell scripts

anacierdem:
Because we use the shell scripts in multi-platform settings and execute them in *nix a container, they should be natively checked out. CRLF would never be useful in a Windows environemnt for a bash script anyways.
- Add gitattributes
- Rename the toolchain build script
- Remove travis config


rasky (approved):
We also have a top-level `build.sh`. Maybe since we're at it, we should rename it `tools/build-toolchain.sh`?

PR #172, opened 2021-09-19
PI dma: add support for misaligned transfers using undocumented hw features

rasky:
Nominally, PI DMA is able only to transfer an even number of bytes
from/to 8-byte aligned addresses. In reality, there's a half-baked
hardware implementation to transfer any number of bytes between 2-byte
aligned addresses that unfortunately does not fully work.

This commit modifies dma_read so that it handles almost all misaligned
addresses by relying on this half-baked support and supplementing it
with a few manual I/O read/writes to complete the transfer.

It also adds a dma_read_async function that only starts the transfers,
and a dma_wait function that waits for completion. dma_busy is
deprecated instead: first, we prefer busy waits to happen within
libdragon (so that they're easier to port to the multitasking kernel),
and second dma_busy is a misnomer and both waits for DMA and any other
kind of PI busy. So it is better to phase it out.

The previous low-level DMA primitive is now exposed as dma_read_raw_async,
just in case somebody absolutely need to just program the hardware, but
dma_read_async is pretty fast to falling back to it for aligned addresses,
so there should be in principle no reasons to call it anymore. This
is also why I decided to improve the dma_read API rather than adding
a dma_read_misaligned API.

The test does a bruteforce for all small sizes and all alignments, so
that we can be confident on the implementation.

### DIFF
--- a/src/dma.c
+++ b/src/dma.c
@@ -47,30 +47,43 @@ static volatile struct PI_regs_s * const PI_regs = (struct PI_regs_s *)0xa460000
  *
  * @return nonzero if the DMA controller is busy or 0 otherwise
  */
-volatile int dma_busy() 
+volatile int __dma_busy(void)
 {
     return PI_regs->status & (PI_STATUS_DMA_BUSY | PI_STATUS_IO_BUSY);
 }
 
+volatile int dma_busy(void)
+{
+    return __dma_busy();
+}
+
 /**
- * @brief Read from a peripheral
+ * @brief Start reading data from a peripheral through PI DMA (low-level)
  *
- * This function should be used when reading from the cartridge.
+ * This function should be used when reading from a cartridge peripheral (typically
+ * ROM). This function just begins executing a raw DMA transfer, which is
+ * well-defined only for RAM addresses which are multiple of 8, ROM addresses
+ * which are  multiple of 2, and lengths which are multiple of 2.
+ * 
+ * Use #dma_wait to wait for the end of the transfer.
+ * 
+ * See #dma_read_async for a higher level primitive which can perform almost
+ * arbitrary transfers.
  *
  * @param[out] ram_address
- *             Pointer to a buffer to place read data
+ *             Pointer to a buffer to place read data (must be 8-byte aligned)
  * @param[in]  pi_address
- *             Memory address of the peripheral to read from
+ *             Memory address of the peripheral to read from (must be 2-byte aligned)
  * @param[in]  len
- *             Length in bytes to read into ram_address
+ *             Length in bytes to read into ram_address (must be multiple of 2)
  */
-void dma_read(void * ram_address, unsigned long pi_address, unsigned long len) 
+void dma_read_raw_async(void * ram_address, unsigned long pi_address, unsigned long len) 
 {
     assert(len > 0);
 
     disable_interrupts();
 
-    while (dma_busy()) ;
+    while (__dma_busy()) ;
     MEMORY_BARRIER();
     PI_regs->ram_address = ram_address;
     MEMORY_BARRIER();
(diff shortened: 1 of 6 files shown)


include/cop0.h
```diff
@@ -14,7 +14,7 @@
 
 /** @brief Read the COP0 Count register (see also TICKS_READ). */
 #define C0_COUNT() ({ \
-    uint32_t x; \
+    volatile uint32_t x; \
```
anacierdem: I was wondering for some time: is this equivalent to wrapping inline assembly with memory barriers? Or we can add full clobbers to the inline assembly?

src/dma.c
```diff
@@ -47,42 +47,230 @@ static volatile struct PI_regs_s * const PI_regs = (struct PI_regs_s *)0xa460000
...
+
+/** @brief Low-level 32-bit unaligned PI ROM read.
+ * 
+ * @note This function must be called with interrupts disabled.
+ */
+static uint32_t __io_read32u(unsigned long pi_address) {
+    uint32_t val = 0;
+
+    // We need to manually issue assembly opcodes here because we must
+    // wait for #dma_busy inbetween. So we can't simply tell GCC to
```
anacierdem, line 116: Is this a hardware requirement?

src/dma.c
```diff
@@ -47,42 +47,230 @@ static volatile struct PI_regs_s * const PI_regs = (struct PI_regs_s *)0xa460000
...
+static uint32_t __io_read32u(unsigned long pi_address) {
+    uint32_t val = 0;
+
+    // We need to manually issue assembly opcodes here because we must
+    // wait for #dma_busy inbetween. So we can't simply tell GCC to 
+    // generate the LWL/LWR pair via __attribute__((aligned(1)), otherwise
+    // we would only be able to wait once before both of them.
+    while (__dma_busy()) {}
+    MEMORY_BARRIER();
+    __asm volatile("lwl %0,0(%1)" : "+r"(val) : "r"(pi_address));
```
anacierdem, line 121: Does this also read from the register? wouldnt be `=r` adequate? for the next one it seems ok

src/dma.c
```diff
@@ -47,42 +47,230 @@ static volatile struct PI_regs_s * const PI_regs = (struct PI_regs_s *)0xa460000
...
+    while (__dma_busy()) {}
+    MEMORY_BARRIER();
+    return *(volatile uint32_t*)pi_address;
+}
+
+/** @brief Low-level 32-bit unaligned PI ROM read.
+ * 
+ * @note This function must be called with interrupts disabled.
+ */
+static uint32_t __io_read32u(unsigned long pi_address) {
```
anacierdem, line 112: I don't know if this is a convention but wouldn't it be better if we provide some explanation stating non `u` postfixed versions should only be used for aligned adresses? Just looking at say `__io_read16`, it is hard to get it is meant for aligned reads. If this is something very common, it is not important.

tests/test_dma.c
```diff
@@ -0,0 +1,27 @@
...
+		memset(ram, 0xAA, sizeof(ram));
+		data_cache_hit_writeback_invalidate(ram, sizeof(ram));
+		dma_read(ram+ram_offset, rom+rom_offset, length);
+
+		ASSERT_EQUAL_MEM(ram+ram_offset-16, expAA, 16, "invalid prefix [%d/%d/%d]", ram_offset&7, rom_offset, length);
+		ASSERT_EQUAL_MEM(ram+ram_offset, rom_copy+rom_offset, length, "invalid data [%d/%d/%d]", ram_offset&7, rom_offset, length);
+		ASSERT_EQUAL_MEM(ram+ram_offset+length, expAA, 16, "invalid suffix [%d/%d/%d]", ram_offset&7, rom_offset, length);
+	}
+
+	for (int i=56; i<64; i++) {
```
anacierdem, line 22: Why start from 56?

anacierdem (commented):
Looks good to me. Mostly questions. Also I wonder if this will really make a difference in practice? For example it doesn't seem to affect dfs test timings, which indeed does small reads. It is nice to have it though.

rasky:
> Also I wonder if this will really make a difference in practice? For example it doesn't seem to affect dfs test timings, which indeed does small reads. It is nice to have it though.

I haven't updated dfs_read to benefit from it. Currently it's still calling the new function on cases where the previous one already did work, so the new codepath is not excercised.

The new function makes a difference if you need to do a misaligned transfer. In that case, before it was impossible without an intermediate memory buffer and then a copy, while now it is possible.

include/cop0.h
rasky: Uhm I have many doubt about this. I've kept this in my tree for a long time, but I'm always unable to really reproduce the bug that is fixed by this change. Also, the change the way it is written pessimizes the code a little bit. So maybe I should back this out from this PR, and try first to reproduce the bug.

src/dma.c
rasky: Yes, the PI requires the busy flag to be ready before accessing it. LWL and LWR are two different memory accesses, so you need to wait for the busy bit in-between.

src/dma.c
rasky: Technically it does (both LWL and LWR only write a subset of bytes of the register). In this case, though, we know that LWR will change the bytes that LWL keeps intact. So we could indeed shortcut this and declare that LWL doesn't use the input value. But that would be false semantically speaking, and only true because there's a perfectly matching LWR later. So all in all, I'd rather keep the asm constraints semantically correct.

tests/test_dma.c
rasky: We need to check 8 different alignments. One common bug could be that the DMA transfers more bytes than requested, that the test verified by checking the suffix that must be intact (0xAA). Another possible mistake is that the DMA transfers bytes *before* the beginning of the destination buffer, and we want to be able to check that as well. So we can't start from 0. I just picked the range 56-64 out of the blue. It could be 64-72 or whatever, as long as there are some preceding bytes to check.

src/dma.c
rasky: I've added "aligned" to `__io_read16` comment. It was there already for `__io_read32`, so It made sense to do the same.

include/cop0.h
anacierdem: I used explicit dependencies for the exception test's inlines but it only works if the referenced symbol is used on the code path that we want to execute after/before it. GCC docs clearly states that it can change inline assembly's position depending on its side effects and the only order guarantee is for the instructions within the same inline block. I'm not sure clobbers will fix this either b/c it might still be placed differently if some instructions do not work on data though I'm not sure if there are such practical examples. I agree that we need to find a way to reproduce and then commit to a solution.

The timing issue between ed/64drive is eventually solved so I think the same will happen to this one as well.

anacierdem: Hmm without explicit dependencies it will always be possible to switch instructions wth syscalls/breaks for example

rasky:
Ready for commit

PR #175, opened 2021-09-20
Start building a docker image

anacierdem:
- Changed build.sh to support non-default installation locations. make searches
default locations (i.e `/usl/local/include`) for `n64.mk` but when it is
installed elsewhere, the script was failing

- Added a two step dockerfile. It builds the toolchain in the first stage and
then it installs the files to N64_INST on second stage to produce a minimal
image.

- Removed the unnecessary native GCC build.

- Apply build script improvements from https://github.com/rasky/libdragon/pull/8

- Add two workflows to verify everything is building fine (`ci.yml`)
and build and push the docker file (`docker-build.yml`).

The default branch (`trunk`) is where we build and push the docker image to
GitHub Packages. `build.sh` script will get executed whenever we push something
or open a pull request over the latest docker image. If `build-toolchain.sh`
file is changed, then it will be used to create a temporary "latest" image to
run the build. The file is always compared against the default branch as latest
image is pushed from there.

When something is merged to the default branch, the ci workflow is executed for
it after the docker-build workflow publishes the new image if necessary. We
should wait for it in order not to rebuild everything from scratch for default
branch ci check. For all other branches, we build the image in the ci workflow
when necessary.

There are a few caveats;
We use the cache implementation of `docker/build-push-action` and the cache key
is per-branch thus we will re-build the images on default branch again even if
we have already built one during the PR process. This can be remedied by
implementing a custom cache but I think the toolchain updates are rare enough
that this is not a problem in practice. Also the cache might grow with time and
I'm not sure if this will be an issue. The buildkit cache is experimental and we
can try to find workarounds or open issues if we face any problems.
The other problem is that if a new PR utilizing a newly building image its ci
will fail as the default branch already have the same toolchain script and ci
will skip building a new temporary image but this is an unlikely event. Even if
it happens its a matter of restarting that job to make it pass after the image
is pushed.

- Update the build badge to use the new workflow.

@DragonMinded It is important that you approve this PR as it is using your `GITHUB_TOKEN` It would be nice to require the ci check on merges to trunk as well but it is not a showstopper.

P.S I have opened the secondary PR to trigger the `pull_request` event as well.
It does not trigger without the branch here having the workflow. Maybe it was unnecessary on second thought, but wanted to make sure it triggers as well. In practice we wont be seing the push check.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: ${{ steps.toolchain_change.outputs.exitCode == 1 }}
+        uses: docker/build-push-action@v2
+        with:
+          context: ./
+          file: ./Dockerfile
+          push: false
+          tags: ghcr.io/${{ github.actor }}/libdragon:latest
+          load: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
```
rasky: Obviously I think it would be better if all this code wasn't duplicated in two actions. Maybe you can find a way to reuse an action using the `uses:` command?

build.sh
```diff
@@ -12,19 +12,19 @@ LIBMIKMOD_COMMIT=738b1e8b11b470360b1b919680d1d88429d9d174
 LIBMIKMOD_DIR=/tmp/libmikmod
 
 # Clean, build, and install libdragon + tools
-make -j${CPU_COUNT} clobber
-make -j${CPU_COUNT} install tools-install
+make -j${CPU_COUNT} -I${N64_INST}/include clobber
+make -j${CPU_COUNT} -I${N64_INST}/include install tools-install
```
rasky: maybe you can do `MAKE=make -j ${CPU_COUNT} -I${N64_INST}/include` and then use `${MAKE}` everywhere, rather than copying all the options in all instances.

rasky: since we're at it, maybe I'd upgrade these scripts to [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/)

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# syntax=docker/dockerfile:1
+
+# Stage 1 - Build the toolchain
+FROM ubuntu:18.04
```
rasky, line 4: Did you try switching to alpine? It can be done as followup of course.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# Stage 2 - Prepare minimal image
+FROM ubuntu:18.04
+ARG N64_INST=/n64_toolchain
+ENV N64_INST=${N64_INST}
+
+COPY --from=0 ${N64_INST} ${N64_INST}
+RUN apt-get update && \
+    apt-get install -yq gcc g++ make libpng-dev git  && \
+    apt-get clean && \
+    apt autoremove -yq
```
rasky, line 39: I think if you switch these two statements (first RUN then COPY), it should improve caching, because the COPY will possibly change everytime libdragon changes, but the way it is right now also RUN will be re-run every time.

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN strip ${N64_INST}/libexec/gcc/mips64-elf/10.2.0/cc1plus
+RUN strip ${N64_INST}/libexec/gcc/mips64-elf/10.2.0/cc1
+RUN strip ${N64_INST}/libexec/gcc/mips64-elf/10.2.0/install-tools/fixincl
+RUN strip ${N64_INST}/libexec/gcc/mips64-elf/10.2.0/lto1
+RUN rm -rf ${N64_INST}/share/locale/*
+
+# Stage 2 - Prepare minimal image
+FROM ubuntu:18.04
+ARG N64_INST=/n64_toolchain
+ENV N64_INST=${N64_INST}
```
rasky, line 33: TIL multi stage builds!

How big is the final image now?

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# install dependencies
+RUN apt-get update
+RUN apt-get install -yq wget bzip2 gcc g++ make file libmpfr-dev libmpc-dev zlib1g-dev texinfo git gcc-multilib
+
+# Build
+COPY ./tools/build-toolchain.sh /tmp/tools/build-toolchain.sh
+WORKDIR /tmp/tools
+RUN ./build-toolchain.sh
```
rasky, line 15: I have one question here. You're copying the shell file into `/tmp` which I think it defeats Docker caching. Since the script is in tmpfs, these steps will always be executed. Thus, every time you run `docker build` on this Dockerfile, I believe it will rebuild the toolchain from scatch without caching.

If you instead copy the `build-toolchain.sh` script to the filesystem, I *think* it should then use it as part of the caching. That is, I would expect that if the script has not changed, the build will be skipped as cached.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,56 @@
+# Build and push a docker image whenever toolchain script is updated. It is
+# important that this workflow always runs to be able to trigger the default
+# branch ci check.
+name: Docker Build
+
+on:
+  push:
+    branches:
+      - trunk
```
rasky, line 9: Can't you use `on.push.paths` here to simplify the checking?

rasky (commented):
Thanks, great work!

.github/workflows/ci.yml
anacierdem: It can probably be done with a common action but the tags are also different so the value would be minimal. I have tried many different things to be able to share the image build but it ended up like this being most practical. Edit: I think you also mean the diff checks etc. I think we can improve this later.

build.sh
anacierdem: will do

Dockerfile
anacierdem: I did try, it was not that simple though needs trial and error on my side, so I skipped.

Dockerfile
anacierdem: You are right but I don't think it will matter too much b/c if we have already built the toolchain it already takes a lot of time. This might be negligable :) Also arguably it does also update packages to the latest if distro lists are updated?

Dockerfile
anacierdem: Not very different, a little smaller: 619MB
But the practical implications of this is much important as we don't have to run the whole build if we write a failing strip command now. The previous steps will be cached.

Dockerfile
anacierdem: Docker does not perform file system comparisons it only looks at the copied file's hashes so this gets properly cached. For the `RUN`s it does only look at the command itself. Also see [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) and the latest check's build time, it used the cache 👍

.github/workflows/docker-build.yml
anacierdem: It is a little complicated than that :) On the default branch we also want to execute `ci` to display the build badge. Currently we run `ci` check after the image is built and pushed so that we don't have to build it both for the `docker-build` and `ci` when something is merged.

The `ci` workflow is triggered on PRs and pushes except for the default branch and also when the `ci`flow is completed for `trunk`. Because of this `docker-build` should always run to trigger `ci` on `trunk`. This also guarantees `trunk`'s `ci` is actually run on the live image just in case.

I know this flow is a little counterintuitive but upon much trial and error I decided that this is the way to go.

Dockerfile
anacierdem: I was previously debugging the image build on my pc by first converting a big `... && ...` into docker commands to enable caching and back to a huge bash command after I'm done so that final image is small.

There was even someone e-mailing me that said he converted every step of the toolchain build into docker `RUN`s to be able to tweak it :)

.github/workflows/docker-build.yml
anacierdem: Hmm I think I have found a better solution now. I will push that instead. Let me verify that first.

anacierdem: Unfortunately it is not possible to prevent triggering a workflow if a path match occurs. i.e;
```
on:
  push:
    paths:
      - "tools/build-toolchain.sh"
      - "Dockerfile"
```
will trigger if relevant files change but there is no "else" to it that we can start `ci` without building an image.

anacierdem:
I improved a few things. Unfortunately it is not possible to further simplify this using path matches. It was something I also wanted and I worked hard to make that happen but it didn't work 🤷‍♂️

.github/workflows/ci.yml
anacierdem: Done to the extent possible. composite actions does not support conditionals.

.github/actions/path-diff/action.yml
```diff
@@ -0,0 +1,34 @@
+name: "Example"
```
rasky: Maybe a different name?

rasky (approved):
Great work!

rasky:
Given that GitHub is giving unlimited CI minutes and unlimited space for Github packages, I think we could assume that @DragonMinded is fine with this, given that we are not impacting any quote of the account. I'd suggest to go with this because we have many more things to build on CI and this is the first step to unblock that work

anacierdem:
I will also make the dockerfike set the new PREFIX and merge this.

PR #176, opened 2021-09-21
[DEMO: DO NOT MERGE] ugfx: Microcode powered 3D&2D graphics

snacchus:
This PR mainly introduces a new system called "ugfx", which is short for "Microcode graphics".

As the name suggests, this system is capable of rendering graphics utilizing the RSP. The API and functionality are heavily inspired by libultra, but the implementation is written from scratch. I originally based this on the rudimentary graphics microcode that is part of n64chain, but over time it has evolved into something completely different.

These are the changes made:
- Add `include/ugfx.h` and `src/ugfx.c`, which contain the entire API.
- Add `src/rsp_ugfx.S`, which contains the microcode. For better readability, it includes quite a few additional source files from the `src/ugfx` directory.
- Add a new example ROM `ugfxtest`, which showcases a lot of ugfx's functionality. The source code also contains some explanatory comments on the API.
- Add some new functions to `display.h`, which proved very useful in the example ROM.
- Add a new macro `PhysicalAddr`, which is needed in some places for specifying DMA addresses for the RSP.
- Modify some RSP assembly macros in `rsp.inc`. Since most vector Load/store operations shift the offset by the operand size, this is now reflected in the generated instructions. This means that the offset is always specified in bytes, which resembles the original RSP language by nintendo more closely. This will most likely break existing microcode.
- Add some more useful defines in `rsp.inc`.
- Fix an issue in `n64.mk`, where RSP assembler invocations were missing some command line flags.

Some notes on this PR:
- I am aware of the recently introduced RSP microcode library and coding guidelines, but since I have started writing the ucode quite a while ago, I have not been able to adapt the code accordingly or make use of the new library. This can be addressed in the future of course.
- As mentioned above, the design of this system is "heavily inspired by libultra". This means a lot of the API is quite similar, but some changes and even a couple of improvements have been made. For some of the trickier parts of the microcode, I resorted to inspecting a disassembled version of the official microcode (which I did not disassemble myself), which acted as a valuable reference implementation for me. I never copied&pasted entire code segments from it, though.
- I have not added any doxygen comments to the API yet. Partly due to lack of time, but also because the API may be subject to change in response to feedback on this PR.
- And finally: The API & implementation are far from perfect and tons of things still need to be improved and optimized. But I just want to get this code out there and gather some feedback since I feel like its current state is fairly presentable 😄


snacchus:
I just realized that I didn't test the final version of the example ROM on hardware before submitting the PR 🤦 
Anyway, I just tested it and I'm getting a floating point exception in `ugfx_matrix_from_column_major`, probably in the macro expansion of `float_to_fixed`. ~~Apparently the multiplication is generating an "inexact operation" exception. I looked this up and apparently most floating point operations will generate this exception, and it can be safely ignored.
Maybe some flags need to be set in the FPU control register to ignore this exception? I will investigate this further when I have the time.~~
EDIT: Looks like I got the specific exception type wrong. It's actually an "Unimplemented Operation Exception" which in this case is triggered by one of the operands being a denormalized number.

The question is: How should this be handled? The VR4300 Manual says "How to use the unimplemented operation exception is arbitrarily determined by the system", which to me sounds like "it doesn't matter". I looked at how n64chain handles this for reference. Their exception handler seems to ignore all floating point exceptions. Should Libdragon do the same? If so, I could implement an appriopriate check in `inthandler.S`.

examples/ugfxtest/Makefile
```diff
@@ -0,0 +1,19 @@
+BUILD_DIR=build
+include n64.mk
+
+all: ugfxtest.z64
+
+$(BUILD_DIR)/spritemap.dfs: \
```
rasky: You probably want to rename this to `ugfxtest.dfs`

include/rsp.inc
```diff
@@ -263,44 +263,44 @@ makeNotImplemented tnei
     .endm
 .endm
 
-.macro makeLoadInstruction name, opcode
+.macro makeLoadInstruction name, opcode, offsetshift
```
rasky: I think we should also add a `.error` in case `\offset` is not a multiple of `\offsetshift`.

include/rsp.inc
```diff
@@ -480,27 +497,27 @@ makeStoreInstruction swv, 0b00111
 
     .macro setup_vsll vshiftreg
         .set noat
-        la $1,%lo(V_SHIFT)
-        lqv \vshiftreg,0,  0,$1
+        la $1, %lo(V_SHIFT)
+        lqv \vshiftreg, 0, 0, $1
```
rasky: The spacing here was intentional, so that visually the second zero refers to the last argument :)

include/rsp.inc
```diff
@@ -456,6 +456,23 @@ makeStoreInstruction swv, 0b00111
...
+#define VE_2h 6
+#define VE_3h 7
+#define VE_0  8
+#define VE_1  9
+#define VE_2  10
+#define VE_3  11
+#define VE_4  12
+#define VE_5  13
+#define VE_6  14
+#define VE_7  15
```
rasky: I like these. Not 100% sure about the naming though, but I don't have better suggestions to offer.
I had previously played with the syntax `e(0h)`, which you can obtain on top of these defines with something like:

```
#define PPCAT2(n,x) n ## x
#define PPCAT(n,x) PPCAT2(n,x)
#define e(xx) PP_CAT(VE_, xx)
```

How do you like it?

```vmudl $v00, $v00, $v00,e(2h)```

File: n64.mk [hunk 5]
```diff
@@ -70,7 +70,7 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.S
 	@mkdir -p $(dir $@)
 	set -e; if case $(notdir $(basename $@)) in "rsp"*) true;; *) false;; esac; then \
 		echo "    [RSP] $<"; \
-		$(N64_CC) $(N64_ASFLAGS) -nostartfiles -MMD -Wl,-Ttext=0x1000 -Wl,-Tdata=0x0 -Wl,-e0x1000 -o $@ $<; \
+		$(N64_CC) $(ASFLAGS) -nostartfiles -MMD -Wl,-Ttext=0x1000 -Wl,-Tdata=0x0 -Wl,-e0x1000 -o $@ $<; \
```
rasky: Can you explain why you need this change? I think `N64_ASFLAGS` was intentional to separate RSP assembly flags from MIPS assembly flags.

rasky:
Hi @snacchus thank you very much for the submission! It looks great On first sight. I'll make a first round of review over the weekend.

include/ugfx.h
```diff
@@ -0,0 +1,601 @@
...
+#define UGFX_OP_LOAD_LIGHT             0x84ULL
+#define UGFX_OP_LINK_COMMANDS          0x85ULL
+#define UGFX_OP_SET_CULL_MODE          0x86ULL
+#define UGFX_OP_SET_GEOMETRY_MODE      0x87ULL
+#define UGFX_OP_SET_PERSP_NORM         0x88ULL
+#define UGFX_OP_SET_TEXTURE_SETTINGS   0x89ULL
+#define UGFX_OP_SET_NUM_LIGHTS         0x8AULL
+#define UGFX_OP_SET_CLIP_RATIO         0x8BULL
+#define UGFX_OP_SET_ADDRESS_SLOT       0x8CULL
+#define UGFX_OP_DRAW_TRIANGLE          0x8DULL
```
rasky, line 223: I think in general we should probably split off the RDP passthrough commands in a different library. I had implemented similar macros here:
https://github.com/rasky/mvs64/blob/main/lib/rdp_commands.h

which can be used also from RSP assembly to pack RDP commands (somehow). That headers also contains some mid-level macros that are groups of RDP commands that achieve something (eg: loading a sprite).

The architecture I had in mind for Libdragon ucode was to have a single display list parser (with stack), and then allow different libraries to plug in their own ucode "overlays" registering them to some range of opcodes. So for instance ugfx could take range 0x80-0x8F, maybe audio processing gets commands 0x90-0x9F, and then a game might want to do their own RSP stuff with other commands. Everything can be mixed-match in display lists, get sent to RSP, and RSP processes it switching different microcode overlays when required.

So in this picture RDP passthrough commands are just one of those ranges (0x20-0x3F), with their own specific little microcode that just pass them through. And they don't need to be part of ugfx as they are more general.

include/ugfx.h
```diff
@@ -0,0 +1,601 @@
...
+        } color;
+
+        struct {
+          int8_t x;
+          int8_t y;
+          int8_t z;
+          uint8_t a;
+        } normal;
+    } attr;
+} __attribute__((aligned(16))) ugfx_vertex_t;
```
rasky, line 525: Did you think about having a more flexible vertex description? For instance having either color or normal is a compromise that one might want not take. If we stick to a static vertex structure like this one, it becomes hard to add/remove steps of the T&L pipeline. 

Maybe we could come up with a more flexible representation where you have a vertex buffer descriptor (a bitmask of flags) to describe what's inside the vertex; the order of component would be fixed but all components could be either present or absent depending on the vertex buffer bitmask.

src/display.c
```diff
@@ -553,4 +553,24 @@ void display_show_force( display_context_t disp )
...
+bitdepth_t display_get_bitdepth()
+{
+    return __bitdepth == 2 ? DEPTH_16_BPP : DEPTH_32_BPP;
+}
+
+uint32_t display_get_num_buffers()
+{
+    return __buffers;
+}
+
```
rasky, line 598: unrelated, but let's also add a getter to the Nth buffer pointer.

src/ugfx.c
```diff
@@ -0,0 +1,180 @@
...
+
+    return ugfx_set_color_image(__safe_buffer[disp - 1], UGFX_FORMAT_RGBA, pixel_size, display_get_width() - 1);
+}
+
+void ugfx_viewport_init(ugfx_viewport_t *viewport, int16_t left, int16_t top, int16_t right, int16_t bottom)
+{
+    left = float_to_fixed(left, 2);
+    top = float_to_fixed(top, 2);
+    right = float_to_fixed(right, 2);
+    bottom = float_to_fixed(bottom, 2);
```
rasky: Maybe just cast and shift here? using `float_to_fixed` for non-floats seems dubious.

include/rsp.inc
```diff
@@ -397,52 +397,52 @@ makeOpInstruction vsubc, 0b010101
...
+makeStoreInstruction ssv, 0b00001, 1
 /** @brief Store Transpose from Vector Register */
-makeStoreInstruction stv, 0b01011
+makeStoreInstruction stv, 0b01011, 4
 /** @brief Store Unsigned Packed from Vector Register */
-makeStoreInstruction suv, 0b00111
+makeStoreInstruction suv, 0b00111, 3
 /** @brief Store Wrapped vector from Vector Register */
-makeStoreInstruction swv, 0b00111
+makeStoreInstruction swv, 0b00111, 4
```
rasky: I'm not opposed to this, but I have ~6000 lines of assembly code to update :) I'll have to come up with a script for sure if we decided to go with this. Why did you feel it was important?

include/ugfx.h
```diff
@@ -0,0 +1,601 @@
...
+
+typedef struct ugfx_matrix_t {
+    int16_t integer[4][4];
+    uint16_t fraction[4][4];
+} __attribute__((aligned(16))) ugfx_matrix_t;
+
+void ugfx_matrix_from_column_major(ugfx_matrix_t *dest, const float *source);
+void ugfx_matrix_from_row_major(ugfx_matrix_t *dest, const float *source);
```
rasky, line 533: I guess we're going to add helpers to create matrices for the user. Should these helpers work on float matrices and then convert them to `ugfx_matrix_t`? Or should we try to have functions that create and manipulate `ugfx_matrix_t` directly?

Also I'm not sure we want to support both style of float matrices. Once we will grow functions to create matrices, we should just pick one style and stick to it I would say.

src/ugfx.c
```diff
@@ -0,0 +1,180 @@
...
+static inline int32_t ugfx_pixel_size_from_bitdepth(bitdepth_t bitdepth)
+{
+    switch (bitdepth)
+    {
+        case DEPTH_16_BPP:
+            return UGFX_PIXEL_SIZE_16B;
+        case DEPTH_32_BPP:
+            return UGFX_PIXEL_SIZE_32B;
+        default:
+            return -1;
```
rasky: maybe just assert here?

src/ugfx.c
```diff
@@ -0,0 +1,180 @@
...
+    {
+        do 
+        {
+            // Round to next power of 2?
+            buffer->capacity = buffer->capacity << 1;
+        } while (new_size > buffer->capacity);
+
+        buffer->data = realloc(buffer->data, buffer->capacity * sizeof(ugfx_command_t));
+    }
+}
```
rasky, line 80: Generally speaking, I think it would be better to have multiple display list buffers and simply allocate a new one if the previous one is full, just dropping a "jump" instruction at the end that will be parsed to RSP to move to the next buffer. The dynamic approach seems like too taxing on the runtime allocator, and copying stuff around after reallocation can also get slow.

src/ugfx.c
```diff
@@ -0,0 +1,180 @@
...
+ugfx_buffer_t* ugfx_buffer_new(uint32_t capacity)
+{
+    ugfx_buffer_t *buffer = malloc(sizeof(ugfx_buffer_t));
+
+    buffer->data = malloc(capacity * sizeof(ugfx_command_t));
+    buffer->capacity = capacity;
+    buffer->length = 0;
+
+    return buffer;
+}
```
rasky, line 55: I did some experiments with display list buffers in my RDL library. https://github.com/rasky/mvs64/blob/main/lib/rdl.h

I found out that, at the end of the day, explicitly managing display lists is complicated to do correctly and is probably best left to the library. Libdragon RDP library gets this right IMO: there is a single global display list (circular buffer) and the APIs do not refer to it explicitly but rather implicitly enqueue commands to it. This is also how OpenGL 1/2 works, if you think of it.

What can be done better with RSP on top of RDP is that we can have multiple display list support (just like OpenGL). We can have functions to create display lists such as:

```
// start enqueing commands in this buffer rather than the global one.
void ugfx_displaylist_begin(ugfx_buffer_t *buffer);
void ugfx_displaylist_end();

// enqueue a command in the global list that tells RSP to call this buffer and then get back to the global one
void ugfx_displaylist_draw(ugfx_buffer_t *buffer);
```

So for instance, `ugfx_set_model_matrix` is now a function-like macro that returns a 64-bit opcode. Instead, it should become a (possibly inline) function that enqueue that opcode into the current display list (which can be either the global, implicit one, or the one we're currently compiling after a call to `ugfx_displaylist_begin`).

This also allows us to have ugfx functions that enqueue multiple 64-bit commands into the display list. Think of a higher-level functions to help the user. And moreover, it doesn't coerce us to 64-bit commands: in the global vision of multiple different ucodes sharing the same global display list, there might be ucodes that prefer to have fatter or smaller commands. For instance, my H264 decoder uses 48-bytes commands in the display list; but this is not exposed to the user: the user just calls `rsph264_interpolate_luma()`, which enqueues as many bytes as it prefers to the active display list.

src/ugfx/dma.S
```diff
@@ -0,0 +1,25 @@
...
+    mtc0  dma_mem_addr, COP0_DMA_SPADDR
+    bgtz  dma_is_write, dma_write
+    mtc0  dma_dram_addr, COP0_DMA_RAMADDR
+    j     dma_done
+    mtc0  tmp2, COP0_DMA_READ
+  dma_write:
+    mtc0  tmp2, COP0_DMA_WRITE
+  dma_done:
+    jr    ra
+    nop
```
rasky, line 23: This code IMO is too similar to official SGI code, I'd be concerned with copyright. Can you please remove this and use rsp_dma.inc?

snacchus:
I managed to make the example ROM work on hardware. I couldn't verify this in detail, but the floating point exception must have been caused by the perspective matrix not being initialized properly.

If you start the example ROM, you should see a blue rotating cube with some texture on it. I tested it with an Everdrive x7 on a PAL console. There seems to be a slight problem with screen tearing, though.

rasky (changes requested):
Thank you very much for this submission. Let me say I really appreciated how clean your submission is and how you strived to make it follow libdragon's approach even in corners that were only very lately merged, like the new RSP library or the build system. We really appreciate that because it helps us in the review and merge process a lot.

The first issue I want to raise upfront is the copyright status of this work. Given past experiences with Nintendo, we are trying very hard to build something that is completely free of copyright, we also have an ongoing plan to replace IPL3 with an open source version, for instance. If we are to merge this PR, we must be absolutely certain that this does not create a copyright problem.

For instance, as I commented below, I recognized the DMA code coming from SGI RSP documentation. This would not be possible to merge, so we would need to change that with different DMA routines, such as those already merged in libdragon (rsp_dma.inc).

I can't comment on the copyright status of the rest of the ucode because I have never seen the commercial microcode. I have asked a couple of people that know about this to have a look at the PR, and both agreed that there are strikingly similarities with the Fast3D ucode, to the point that it might be regarded as a forked version of it, rather than a full original work. I would like to know what you think on this before we move forward, so that we can have a plan together on how to solve this issue.

Because of this outstanding issue (plus lack of time), I've still not reviewed the ucode itself properly. I have mainly glanced the C code for now from an API / architecture standpoint. 

I suggest that we split the rsp.inc changes in a different PR. Those are good to go by themselves, and if we finally agree on changing the load/store opcodes offsets, I'll have to roll up my sleeves on my ucode anyway, so better land that separately and earlier.

As for the general architecture, I left a comment on the design that I was toying with and I had discussed already with other people. Mainly, I think it's more flexible to handle one implicit, global circular buffer for commands with OpenGL-style commands (implicit destination buffer), so that we also allow a pluggable design where other ucodes for different tasks (audio, video, whatever) can enqueue commands of different lengths. I'd love to hear your opinion on this.

Thanks again for your submission!

rasky:
> I managed to make the example ROM work on hardware. I couldn't verify this in detail, but the floating point exception must have been caused by the perspective matrix not being initialized properly.

Thanks, and agreed. Floating point exceptions have helped me find several hairy bugs in the code. It's true they were turned off in the original toolchain, but I think we're doing a much better job at making a easy-to-use toolchain here, including all asserts for invalid arguments that would simply crash the console otherwise. So I think we'd rather keep those exceptions on.

examples/ugfxtest/Makefile
snacchus: Copy paste strikes again! 🤦

include/rsp.inc
snacchus: I see, I didn't catch that. I rolled it back.

src/display.c
snacchus: Good idea! I added it.

src/ugfx.c
snacchus: Agreed. I don't know what I was thinking there.

src/ugfx.c
snacchus: Agreed. I'm still unsure what format to use here. How about `assert(!"<message>")` ? Or maybe assert for an actual condition at the start of the function?

src/ugfx/dma.S
snacchus: Yes, I fully agree. I think a lot of the microcode probably needs to be rewritten or redesigned, so I would l ike to do that together with all other changes.

include/rsp.inc
snacchus: I like that idea a lot! I will include it in the next commit.

n64.mk
snacchus: Strange. I had to make this change before because an include path was missing. I just tried rebuilding without the change and it worked again. Maybe a caching issue? Anyway, I'll roll it back.

include/rsp.inc
snacchus: I think I felt it was important because I originally wrote the microcode using armips, which tries to replicate the original syntax closely. Thinking about it, it does make more sense to specify the offset in operand-units instead of bytes because it's a lot less error prone. I will roll this change back and adjust my code (which by the sounds of it would be a lot less work than the alternative).

include/rsp.inc
snacchus: See my other comment, I will roll this change back again.

include/ugfx.h
snacchus: I'm a big fan of this. In libultra, the microcode needs to be re-uploaded for every single task, which seems wasteful. This idea of a modular microcode system sounds a lot better to me. It also sounds like a lot of design effort needs to go into this, so a big question is how we should approach this. I saw that you opened another PR with a new audio system (which looks great by the way) that also includes microcode. Would you like to merge that first, or shall we try and make the two systems compatible right away? In any case, I'd gladly participate in the effort.

include/ugfx.h
snacchus: I have played around with [cglm](https://github.com/recp/cglm) before, which works on float matrices. I'm not sure if using the FPU is worth the convenience vs the performance, though. Some profiling should be done to assess this, I think. Either way, we should probably offer our own set of helpers for creating matrices. I don't think we could just incorporate a library like cglm into libdragon...?

I agree that we should stick to one matrix format. My pick would be column major, because that's the format used by the microcode anyway. Matrix multiplications just work out to be much easier and more efficient that way. (Libultra uses row major by the way and I wonder what made them decide on that format)

include/rsp.inc
rasky: Actually, hold on on this a little bit. I'd like to poll other people.

I think the byte offset is a more natural syntax and also mimics how scalar load/store opcodes work. If I can find a way to automate conversion of my ucode, I think it's not going to be as much as a hassle as I initially thought.

src/ugfx.c
snacchus: Good idea! I think this is closely related to the idea of a modular microcode system (as mentioned in another comment). So instead of a display list that is managed by ugfx, it would be a more generic component.

I think by making the user manage the command buffers I wanted to avoid having a singleton-like API, but in hindsight that doesn't really make sense in the context of libdragon. Maybe just a force of habit 😄

src/ugfx.c
snacchus: Good point, I hadn't thought of this. This could also be how the globally managed display list is handled internally.

include/ugfx.h
snacchus: I generally like this idea. I was thinking about going even further and separating each vertex component into their own buffers (so one could choose the vertex format on the fly without having to rearrange the vertex buffer). But that would make efficient DMA transfers very difficult, if not impossible. It's only possible to specify stride and line width on the source, but not on the target, correct? This is just an idea, though. I'm not sure if this additional effort is even worth it, or would serve any valid use cases.

snacchus:
@rasky Thank you very much for the review! I already pushed some fixes for the minor issues, and replied to all of your other comments. I would say I generally agree with your suggestions.

Regarding the copyright concerns: I was afraid this would be an issue, and I'm not going to deny that the ucode I wrote is quite similar to Fast3D in some places. The DMA function for example is kind of an oversight of mine, but luckily it can just be replaced by the one in rsp_dma.inc.

I suspect the other areas of concern are mainly the `output_flush` function in `output.S` and various parts of `draw_triangle.S`. Admittedly, I think I was quite lazy while implementing `output_flush` and still don't fully understand what happens there, so that one's on me. We can probably find a better solution together.

As for the triangle drawing function: There are some intricate details about assembling the triangle commands that I just couldn't work out without looking at Fast3D. I'm not sure how to implement something equivalent without just making it look very similar. I hope we can work this out together as well.

Please correct me if I'm missing other parts that were of particular concern. I think generally it would be a good idea to discuss the ucode in closer detail, which maybe GitHub isn't even the best place for. (Maybe now is the time for me to finally join the Discord).

As I commented before, I really like your idea of a more generic, modular microcode system. This got me thinking, and perhaps some of the copyright concerns would disappear if we came up with a different rendering pipeline altogether, more akin to OpenGL for example? I guess I limited myself to replicating what libultra does while implementing this system, which was a mistake in hindsight. I gained a lot of practice in writing ucode though and would be willing to reimplement it as a consequence of big design changes. What do you think?

include/ugfx.h
rasky: I think it would be better to review and land that PR and this one independently, and then at some point work towards the unified command list. It could happen even after we merge both -- at the end of the day, if you think of it, it's all behind the public API (once you hide the explicit display list management).

I think your PR is a better staging area for this because the graphics microcode already needs to parse a command list (the audio microcode that I submitted would be a single command, I think -- but I have more to submit that would require different commands).

include/ugfx.h
rasky: That could be an option as well, yes, though you would need multiple DMA transfers to fetch a vertex, which might get a bit in the way. I would be happy to have a single vertex buffer with a descriptor, similar to how OpenGL handles the vertex buffer objects. It'd be a more static version of course, for performance, so I don't think we need to allow any kind of arrangement of data. Just having a bitmask to turn on/off attributes (like optional fields of a single structure) I think it would be enough.

include/ugfx.h
rasky: Nice I didn't know that library. I wouldn't incorporate it verbatim though, it's too much stuff that we don't need for 3d graphics, plus we don't need the SSE versions, etc. But we can borrow some nice ideas I saw on the APIs for "modern C".

Column major is fine by me.

src/ugfx.c
rasky: `assertf(0, "unsupported bit depth: %d", depth);`

src/ugfx.c
rasky: Yes, I was thinking of having N display list buffers that are used in circular fashion. We can either use a jump instruction, or alternatively the RSP knows the addresses of all of them beforehand (could be a single big array that we split in N "sections") and then we need a way to signal the RSP that a section is ready to be processed. I have some working code for something very similar already, but I need to extract it.

rasky:
> As for the triangle drawing function: There are some intricate details about assembling the triangle commands that I just couldn't work out without looking at Fast3D. I'm not sure how to implement something equivalent without just making it look very similar. I hope we can work this out together as well.

In these cases, I think we can work together. I haven't seen original microcode, so I think I can come up with a different solution.

> Please correct me if I'm missing other parts that were of particular concern.

I will ask. I think in general also adapting it to the libdragon register style (see also my ucode) would make it look different.

> I think generally it would be a good idea to discuss the ucode in closer detail, which maybe GitHub isn't even the best place for. (Maybe now is the time for me to finally join the Discord).

If you join the n64brew discord, we can discuss there, sure.

> As I commented before, I really like your idea of a more generic, modular microcode system. This got me thinking, and perhaps some of the copyright concerns would disappear if we came up with a different rendering pipeline altogether, more akin to OpenGL for example? I guess I limited myself to replicating what libultra does while implementing this system, which was a mistake in hindsight. I gained a lot of practice in writing ucode though and would be willing to reimplement it as a consequence of big design changes. What do you think?

Yes. I don't want to sound conceited, but I think we can do much better than the original microcode. We have more years of experience in what works on the console and what doesn't, more tools at disposal (eg: I use my own RSP debugger which is very handy), and no commercial deadlines to meet. So I would start from a fresh approach and see where we get with it.

include/rsp.inc
anacierdem: Another idea might be having something to switch between the two.

rasky:
@snacchus I think I've screwed up the rebase. Would you please rebase it on master? I'd love to keep this up to date every once in awhile

snacchus:
@rasky Done!

rasky:
We can close it now. The [unstable branch](https://github.com/DragonMinded/libdragon/wiki/Unstable-branch) features a more complete 3D solution based on the OpenGL API.

PR #177, opened 2021-09-26
New audio library

rasky:
This PR introduces a new audio library, on top of audio.c.

This first step is a new RSP-based mixer. The mixer mixes and plays up to 32 independent audio streams called "waveforms". It also supports resampling: each waveform can play at a different playback frequency, which in turn can be different from the final output frequency. The resampling and mixing is performed by a very efficient RSP microcode (see rsp_mixer.S). Some figures on RSP mixer performances are on top of rsp_mixer.S, but for instance you can playback 32 channels at 44100Hz using ~20% of total RSP time (resampling is what takes time -- mixing itself would take less than 2% RSP).

The waveform API is the pluggable entry point to implement audio formats. For now, the only implemented waveform is "WAV64", a simple raw sample format. The implementation is trivial: samples are pulled from disk using `dma_read` when needed (well, I must say, it's trivial now that `dma_read` has been powered up to allow unaligned transfers...).

The whole mixer API and implementation has been carefully designed to be zero-copy: the CPU never ever touches the samples, no buffer is copied around. All samples are always directly generated, processed and mixed on the RSP, with the minimal amount of shuffling, and then directly send to AI via DMA (using the new `audio_write_begin` zero-copy API). There are only a few exceptions where we absolutely need to touch a few samples with CPUs every once in a while, but it's so rare and affects so few samples that performance is not impacted.

A new tool is introduced called "audioconv64". It converts trees of files into a destination folder, matching all known audio formats and converting them for the audio library. For now, the only supported conversion is `.wav` -> `.wav64`.

More PRs will follow to introduce other audio formats (no spoilers...).


include/mixer.h
```diff
@@ -0,0 +1,461 @@
...
+
+    // Number of interleaved audio channels in this waveforms. Currently, this
+    // must always be 1.
+    uint8_t channels;
+
+    // Desired playback frequency (in samples per second, aka Hz).
+    float frequency;
+
+    // Length of the waveform, in number of samples. If the length is not
+    // known, this value can be initialized to a very large number (eg: 2**31-1)
```
Ryzee119: As discussed on discord, 2^31-1 does not work (the sample will not play), currently the max len appears to be 0x40000 - 1.

Ryzee119:
I can't comment on the subtle details on the N64 implementation, however I have incorporated this lib into my game port https://github.com/Ryzee119/cosmo64 with really good results.
Good work on this!
I tested the n64conv tool separately and the waveform conversion worked as expected.
I mainly tested generating PCM samples and feeding them in the mixer with a `waveformt_t` struct. Mixing music and sfxs upto 4 or 5 channels

include/mixer.h
rasky: I've pushed a change that allows for longer waveforms to be played back, switching to 64-bit positions in C code while keeping RSP 32-bit (RSP just processes a short window at a time anyway).

examples/mixertest/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+src = mixertest.c
+assets = $(wildcard assets/*.wav)
+assets_conv = $(addprefix filesystem/,$(notdir $(assets:%.wav=%.wav64)))
+
+all: mixertest.z64
+
+# Run audioconv on all WAV files under assets/
+# We do this file by file, but we could even do it just once for the whole
+# directory, becuase audioconv64 supports directory walking.
```
meeq: Typo: `because`

include/mixer.h
Ryzee119: Thanks! Appears to work.
I changed my code to WAVEFORM_UNKNOWN_LEN with loop_len = 0 and had it running for the past 10 mins or so with no issues in the music stream.

examples/mixertest/Makefile
rasky: fixed thanks

rasky:
I've added a commit that adds support for stereo waveforms (waveforms with interleaved left/right samples). Playing these waveforms use two mixer channels at the same time, as you need to mixing slots, but it's the best way in case your input source is already interleaved.

To avoid performance hits, this require a RSP ucode modification to de-interleave the left/right samples while resampling, in both 8-bit and 16-bit mode, but it's actually pretty easy otherwise and falls well within the overall design of the ucode.

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+ * One of the main design goals for the mixer is to provide an efficient way to
+ * compose different audio sources. To achieve full flexibility without impacting
+ * speed, the mixer tries to honor the "CPU zero copy" design principle: samples are
+ * (almost) never touched or moved around with CPU. The whole API has been
+ * carefully designed so that audio players can produce samples directly into
+ * a mixer-provided buffer, so that the mixer can then resample and mix them on
+ * the RSP and store the final output stream directly in the final buffer which
+ * is sent via DMA to the N64 AI hardware DAC. In general, the N64 CPU is very
+ * bandwidth-limited on RDRAM accesses, so "touching" 44100 samples per second
+ * with the PCU would already consume too many resources. Audio players are
```
snacchus: PCU -> CPU

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+ * other waveform that might have been reproduced on this channel.
+ * 
+ * Waveform settings are applied to the mixer channel; for instance, the
+ * frequency of the channel is modified to adapt to the frequency requested
+ * for correct playback of the waveform.
+ * 
+ * If the waveform is marked as stereo (channels == 2), the mixer will need
+ * two channels to play it back. "ch" will be used for the left samples,
+ * while "ch+1" will be used for the right samples. After this, it is
+ * forbidden
```
snacchus: Is there some text missing here?

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+ *********************************************************************/
+
+/**
+ * @brief Waveform callback function invoked by the mixer to read/generate samples.
+ * 
+ * WaveformRead is a callback function that will be invoked by the mixer
+ * whenever a new unavailable portion of the waveform is requested.
+ *
+ * "wpos" indicates the absolute position in the waveform from which
+ * starts reading, and "wlen" indicates how many samples to read.
```
snacchus: "...from which _it_ starts reading,..."?

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+ * that there was a jump in position that should be taken into account.
+ *
+ * Normally, a seeking is required in the following situations:
+ *
+ *   * At start (#mixer_ch_play): the first read issued by the mixer will
+ *     be at position 0 with seeking == true (unless the waveform was
+ *     already partially cached in the sample buffer's channel).
+ *   * At user request (#mixer_ch_set_pos).
+ *   * At loop: if a loop is specified in the waveform (loop_len != 0), the
+ *     mixer will seek when requires to execute the loop.
```
snacchus: requires -> required

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+    const char *name;
+
+    // Width of a sample of this waveform, in bits. Supported values are 8 or 16.
+    // Notice that samples must always be signed.
+    uint8_t bits;
+
+    // Number of interleaved audio channels in this waveforms. Supported values
+    // are 1 and 2 (mono and stereo waveforms). Notice that a stereo waveform
+    // will use two consecutive mixer channels to be played back.
+    uint8_t channels;
```
snacchus: Maybe use enums or defines for supported values? Like `WAVEFORM_SAMPLE_WIDTH_16` or `WAVEFORM_CHANNELS_MONO`

include/wav64.h
```diff
@@ -0,0 +1,59 @@
...
+/** 
+ * @brief WAV64 structure
+ * 
+ * This structure is initialized by #wav64_open to refer to an opened WAV64
+ * file. It is meant to be played back through the audio mixer, implementing
+ * the #waveform_t interface. As such, samples are not preloaded in memory
+ * but rather loaded on request when needed for playback, streaming directly
+ * from ROM. See #waveform_t for more details.
+ * 
+ * Use #wav64_play to playback. For more advances usage, call directly the
```
snacchus: advances -> advanced

include/mixer.h
```diff
@@ -0,0 +1,478 @@
...
+ * with #mixer_ch_play, they will use automatically two channels (the specified
+ * one and the following).
+ */
+typedef struct waveform_s {
+    // Name of the waveform (for debugging purposes)
+    const char *name;
+
+    // Width of a sample of this waveform, in bits. Supported values are 8 or 16.
+    // Notice that samples must always be signed.
+    uint8_t bits;
```
snacchus: In some places the sample width is defined in bytes and sometimes in bits. Is there a specific reason for this? If not, I would make it consistent, probably in favor of bytes.

src/audio/wav64.c
```diff
@@ -0,0 +1,79 @@
...
+	wav->wave.loop_len = head.loop_len; 
+	wav->rom_addr = dfs_rom_addr(fn) + head.start_offset;
+	dfs_close(fh);
+
+	wav->wave.read = waveform_read;
+	wav->wave.ctx = wav;
+}
+
+void wav64_set_loop(wav64_t *wav, bool loop) {
+	wav->wave.loop_len = wav->wave.len;
```
Ryzee119: never checks `bool loop`
Should this be something like `wav->wave.loop_len = (loop) ? wav->wave.len : 0;`

So you can do `wav64_set_loop(sfx, false)` to disable looping.

tools/audioconv64/audioconv64.c
```diff
@@ -0,0 +1,171 @@
...
+#include <dirent.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+
+bool flag_verbose = false;
+
+#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#define LE32_TO_HOST(i) (((uint32_t)((i) & 0xFF000000) >> 24) | ((uint32_t)((i) & 0x00FF0000) >>  8) | ((uint32_t)((i) & 0x0000FF00) <<  8) | ((uint32_t)((i) & 0x000000FF) << 24))
+#define BE32_TO_HOST(i) (i)
+#define HOST_TO_LE32(i) BE32_TO_HOST(i)
```
snacchus: Should probably be `LE32_TO_HOST(i)` right?

tools/audioconv64/audioconv64.c
```diff
@@ -0,0 +1,171 @@
...
+#else
+#define BE32_TO_HOST(i) (((uint32_t)((i) & 0xFF000000) >> 24) | ((uint32_t)((i) & 0x00FF0000) >>  8) | ((uint32_t)((i) & 0x0000FF00) <<  8) | ((uint32_t)((i) & 0x000000FF) << 24))
+#define LE32_TO_HOST(i) (i)
+#define HOST_TO_BE32(i) BE32_TO_HOST(i)
+#define HOST_TO_LE32(i) (i)
+#define BE16_TO_HOST(i) ((((i) & 0xFF00) >> 8)  | (((i) & 0xFF) << 8))
+#define LE16_TO_HOST(i) (i)
+#define HOST_TO_BE16(i) ((((i) & 0xFF00) >> 8)  | (((i) & 0xFF) << 8))
+#define HOST_TO_LE16(i) (i)
+#endif
```
snacchus: Some expressions are duplicated here. Maybe refactor them out?

src/audio/rsp_mixer.S
```diff
@@ -0,0 +1,1133 @@
...
+NUM_CHANNELS:             .half  0
+OUTPUT_RDRAM:             .long  0
+STATE_RDRAM:               .long  0
+
+	.align 4
+CHANNEL_VOLUMES_L:        .dcb.w MAX_CHANNELS
+CHANNEL_VOLUMES_R:        .dcb.w MAX_CHANNELS
+
+	.align 4
+	# FIXME: THIS COMMENT IS OUTDATED - PLEASE UPDATE IT
```
snacchus: Is it still outdated?

src/audio/rsp_mixer.S
```diff
@@ -0,0 +1,1133 @@
...
+XVOL_L:                   .dcb.w MAX_CHANNELS
+XVOL_R:                   .dcb.w MAX_CHANNELS
+
+
+	# Temporary cache of samples fetched by DMA. Notice that this must be
+	# less or equal than MIXER_LOOP_OVERREAD (mixer.c), because the
+	# RSP will over-read up to this amount of bytes after waveform's end.
+	.align 3
+	#define SAMPLE_CACHE_SIZE  64
+DMEM_SAMPLE_CACHE:		  .dcb.b SAMPLE_CACHE_SIZE
```
snacchus: These zero-filled data sections could be moved to .bss as well.

src/audio/rsp_mixer.S
```diff
@@ -0,0 +1,1133 @@
...
+MainLoop:
+	# Fetch output RDRAM pointer and set t1=1 if it's not 8-byte aligned.
+	# We expect the pointer to be 32-bit aligned (since it's a buffer of 16-bit
+	# stereo samples), but it might not be 64-bit aligned.
+	lw t1, %lo(OUTPUT_RDRAM)
+	andi t1, 7
+	slt t1, zero, t1
+
+	lh t4, %lo(NUM_SAMPLES)
+	beqz t4, End
```
snacchus: Maybe do this at the start of the loop?

src/audio/rsp_mixer.S
```diff
@@ -0,0 +1,1133 @@
...
+	beqz wv_addr, WaveLoopEpilog2
+	lw wv_pos,       0(s1)
+	lw wv_step,      4(s1)
+	lw wv_len,       8(s1)
+	lw wv_loop_len, 12(s1)
+
+	# Prepare output pointer
+	addi out_ptr, nchan, %lo(CHANNEL_BUFFER)
+	add out_ptr, nchan
+	move ticks, k1
```
snacchus: There is a comment at the start of the function about what `k1` is, but maybe add a define for it anyway, because it will make the code much more readable. This generally applies to function arguments.

src/audio/rsp_mixer.S
```diff
@@ -0,0 +1,1133 @@
...
+	# Load initial samples
+	lqv v_sample_0,0, 0,s0
+	lqv v_sample_1,0, 1,s0
+	lqv v_sample_2,0, 2,s0
+	lqv v_sample_3,0, 3,s0
+
+	# For optimal pipelining, output is stored at the beginning of the loop. To avoid
+	# corrupting memory, load the output register with whatever is there now.
+	lsv v_out_l,0, -2,s4
+	blt k0, 8, Mix8Start    # Optimized mixing loop for <= 8 channels
```
snacchus: The comment says `<= 8` but this actually checks for `< 8`

snacchus (commented):
I think this is some fantastic work! The API is very clean, well designed and thouroughly documented. I just left some comments on spelling mistakes or very small issues here and there. Since I don't know very much about audio processing, I can't really comment on the correctness of the implementation. But I trust that it is sound since it has already been tested.

The ucode seems good as well, apart from the few comments I left there is one thing maybe: I have found that ucode (and assembly in general of course) becomes much more readable and easy to follow if you give meaningful names to registers via defines and such. Of course if they are just scratch registers used over a couple lines that's fine, but as their scope becomes longer, it gets more complicated to follow the program's logic. The same applies to function arguments. I'm not sure how to handle passing arguments in from outside though. As in, I'm calling a function with `jal` and need to set the arguments, you still need to scroll all the way to the function definition to know the argument's meanings. Maybe we can come up with a scheme for this. Maybe this is just me who hasn't developed in asm for very long, but I think it will be better for newcomers as well.

The only other concern I have is the lack of an asynchronous API. Something like `mixer_poll_start()` and `mixer_poll_wait()`. Because having the CPU stall while the RSP is running seems like a waste. This becomes even more important once we merge the new kernel. Or was your intention to move `mixer_poll` into the audio thread anyway and handle asynchrony on the RSP-API level?

Once again, this is some great work, well done!

rasky:
> The ucode seems good as well, apart from the few comments I left there is one thing maybe: I have found that ucode (and assembly in general of course) becomes much more readable and easy to follow if you give meaningful names to registers via defines and such. Of course if they are just scratch registers used over a couple lines that's fine, but as their scope becomes longer, it gets more complicated to follow the program's logic.

Yes I agree. I think the resampling and mixing loops are pretty in line with what you describe, but the main loop at the start of the ucode doesn't use many defines. I'll change that.

> The same applies to function arguments. I'm not sure how to handle passing arguments in from outside though. As in, I'm calling a function with jal and need to set the arguments, you still need to scroll all the way to the function definition to know the argument's meanings. Maybe we can come up with a scheme for this.

I must say that it's not much different in C; if I see a function call point, I don't know the argument meanings without going to the function definition. Maybe the difference is that it's less clear which registers are used as function arguments?

> The only other concern I have is the lack of an asynchronous API. Something like mixer_poll_start() and mixer_poll_wait(). Because having the CPU stall while the RSP is running seems like a waste. This becomes even more important once we merge the new kernel. Or was your intention to move mixer_poll into the audio thread anyway and handle asynchrony on the RSP-API level?

Yeah the intention was to run mixer_poll into the audio thread, and let it yield while the RSP is running. An async API can be designed of course but it wouldn't be so easy to use. Consider that mixer_poll can call into RSP multiple times, so you would have to change the API so that each call to mixer_poll_wait returns the number of samples actually generated, so that the caller can keep on repeating calls until the required number of samples is reached. On top of that, consider that waveform_t reading functions might use RSP as well, for instance to decompress VADPCM or MPEG2. It's even less clear how we would handle that in an async API.

src/audio/wav64.c
rasky: Good catch, I'll fix that.

src/audio/rsp_mixer.S
rasky: Good catch, I'll fix this

src/audio/rsp_mixer.S
rasky: Done. But I can't do that for function arguments in general because they're function-relative, so I can't easily add defines to the calling sites.

I think maybe the DMA function are the one creating more issues here. I think those might have to be learnt by heart but in general s0/s4 are often input/output, and t0 is count. t2=pitch (not used in this microcode) is possibly the odd one. If you see Mixer, it also takes s4 as its output argument.

src/audio/rsp_mixer.S
rasky: Done

src/audio/rsp_mixer.S
rasky: Yes. I've fixed It now.

src/audio/rsp_mixer.S
rasky: Done, but not for WAVFORM_SETTINGS which is input data, and I don't want to move it to a different section where its address might change.

include/mixer.h
rasky: I'd rather keep them numeric because I use their actual numeric value in formulas (eg: `num_samples * channels`).

snacchus:
> I must say that it's not much different in C; if I see a function call point, I don't know the argument meanings without going to the function definition. Maybe the difference is that it's less clear which registers are used as function arguments?

Very true. In my ucode I experimented with adding global defines for function arguments with the function name as prefix. So for example `DmaDramAddr` or something. But obviously that only works if the function is defined before it is ever used, which is not always possible and it also clutters the global "namespace". I can't think of anything else so I think this is fine for now.

> Yeah the intention was to run mixer_poll into the audio thread, and let it yield while the RSP is running. An async API can be designed of course but it wouldn't be so easy to use. Consider that mixer_poll can call into RSP multiple times, so you would have to change the API so that each call to mixer_poll_wait returns the number of samples actually generated, so that the caller can keep on repeating calls until the required number of samples is reached. On top of that, consider that waveform_t reading functions might use RSP as well, for instance to decompress VADPCM or MPEG2. It's even less clear how we would handle that in an async API.

If `mixer_poll` is always put into the audio thread then that's fine by me. It should work pretty well that way. The issues you mentioned with an async API are valid, and I hadn't thought of them. It would make the API way too confusing so it's probably not a good idea.

src/audio/rsp_mixer.S
snacchus: That's fair. Also see my previous comment.

src/audio/mixer.c
```diff
@@ -0,0 +1,645 @@
...
+	}
+
+	assert(cur == Mixer.ch_buf_mem+totsize);
+}
+
+void mixer_set_vol(float vol) {
+	Mixer.vol = vol;
+}
+
+void mixer_close(void) {
```
meeq, line 188: `mixer_close` does not appear in `mixer.h` -- was this intentional?

meeq, line 186: `mixer_close` does not appear in `mixer.h` -- was this intentional?

meeq:
I'm trying to build the tools on this branch, and I'm getting this cryptic error:

```
make -C audioconv64
make[1]: Entering directory 'libdragon/tools/audioconv64'
cc -std=gnu11 -MMD -O2 -Wall -Wno-unused-result -Werror -I../../include  -o audioconv64 audioconv64.c
In file included from audioconv64.c:38:
conv_wav64.c: In function ‘wav_convert’:
conv_wav64.c:44:9: error: ‘strncpy’ output truncated before terminating nul copying 4 bytes from a string of the same length [-Werror=stringop-truncation]
   44 |         strncpy(head.id, "WV64", 4);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [Makefile:7: audioconv64] Error 1
```

I'm using GCC 11.1.0 on x64 Manjaro Linux

**EDIT**

I was able to work around this with the following patch:
```diff
diff --git a/tools/audioconv64/conv_wav64.c b/tools/audioconv64/conv_wav64.c
index 4fbaf2d..f4699be 100644
--- a/tools/audioconv64/conv_wav64.c
+++ b/tools/audioconv64/conv_wav64.c
@@ -38,18 +38,17 @@ int wav_convert(const char *infn, const char *outfn) {
 		loop_len -= 1;
 	}
 
-	wav64_header_t head;
-	memset(&head, 0, sizeof(wav64_header_t));
-
-	strncpy(head.id, "WV64", 4);
-	head.version = WAV64_FILE_VERSION;
-	head.format = 0;
-	head.channels = wav.channels;
-	head.nbits = nbits;
-	head.freq = HOST_TO_BE32(wav.sampleRate);
-	head.len = HOST_TO_BE32(cnt);
-	head.loop_len = HOST_TO_BE32(loop_len);
-	head.start_offset = HOST_TO_BE32(sizeof(wav64_header_t));
+	wav64_header_t head = {
+		.id = {'W', 'V', '6', '4'},
+		.version = WAV64_FILE_VERSION,
+		.format = 0,
+		.channels = wav.channels,
+		.nbits = nbits,
+		.freq = HOST_TO_BE32(wav.sampleRate),
+		.len = HOST_TO_BE32(cnt),
+		.loop_len = HOST_TO_BE32(loop_len),
+		.start_offset = HOST_TO_BE32(sizeof(wav64_header_t)),
+	};
 
 	if (flag_verbose)
 		fprintf(stderr, "Converting: %s => %s\n", infn, outfn);
```

Ryzee119:
> conv_wav64.c:44:9: error: ‘strncpy’ output truncated before terminating nul copying 4 bytes from a string of the same length [-Werror=stringop-truncation]
>    44 |         strncpy(head.id, "WV64", 4);
>       |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~

What if a memcpy is used instead of a strncpy?

This is a very pendandic error from -Wall and -Werror. Otherwise we need something like `-Wno-error=stringop-truncation`?

meeq:
I have [replaced the mixer in FlappyBird with the implementation from this branch](https://github.com/meeq/FlappyBird-N64/commit/4bd5978bcb258ca70d944aada52583c21d0e75a0), but audio isn't working for me in cen64 or m64p.

Compiled ROM: [FlappyBird.zip](https://github.com/DragonMinded/libdragon/files/7298156/FlappyBird.zip)


There isn't much surface area to screw up here, but I must be doing something wrong.

Ryzee119:
> I have [replaced the mixer in FlappyBird with the implementation from this branch](https://github.com/meeq/FlappyBird-N64/commit/4bd5978bcb258ca70d944aada52583c21d0e75a0), but audio isn't working for me in cen64 or m64p.
> 
> Compiled ROM: [FlappyBird.zip](https://github.com/DragonMinded/libdragon/files/7298156/FlappyBird.zip)
> 
> There isn't much surface area to screw up here, but I must be doing something wrong.

Tested your zip and have audio here (With Cen64) 🤔 Cool game btw!

rasky:
@meeq you probably need to update emulators to the latest version. The new dma_read() function which works with misaligned addresses requires at least a partial implementation of PI DMA undocumented behaviours, which has been done on cen64, m64p, and ares.

Notice that I'm not a fan of breaking emulators, I thought *a lot* of this. But for audio and in general, the advantage of unlocking this previously unknown potential of the PI DMA is very useful, and hopefully the emulators will catch up in a short time. Also I'm personally opening PRs for this to make up with having broken them :)

rasky:
> I'm trying to build the tools on this branch, and I'm getting this cryptic error:

thanks, just pushed the fix for this to this PR

src/audio/mixer.c
rasky: No it's an oversight. I'll add it, thanks.

include/mixer.h
rasky: The general idea is to use bits for the public API, because I think everybody is familiar with the concept of audio being either 8-bit or 16-bit (I think the sentence "let's play this 8-bit wav" is more clear than "let's play this 1-byte wav"). 

But then internally, bits are not very useful and I mainly need two different things: "bytes per sample" (that also includes the number of channels into account, so it is something like `bits / 8 * channels`) and "bitshifts per samples", which is the shift amount that you can use to convert sample indices to byte offsets viceversa. For instance, a 16-bit stereo audio requires 4 bytes per sample, so the shift amount to use is `<< 2`. Obviously you can't really map everything conceivable into a shift, but as long as we are limited to 8 or 16 bits and 1 or 2 channels, then you can indeed do everything with shifts, which makes the code faster.

The issue with this is that in mixer.c and samplebuffer.c unfortunately I'm using the "bps" acronym for both "bytes per sample" and "bitshift per sample", which can make the code a little confusing to read. I could fix this in the future to help readability.

But going back to your comment, I think that it's the correct choice to use bits in public API and bytes internally. It might be less uniform, but I think it's clearer for users.

tools/audioconv64/audioconv64.c
rasky: Done

rasky:
Ok I think I've applied all review comments. I think nobody has actually reviewed the C implementation yet, but I understand that not many people are familiar with low level digital audio processing.

On the other hand, it's been tested in three different games with good results, and I have done endless tests on it so I think the implementation is indeed correct (I've just fixed some small glitch with stereo waveforms -- but those were only recently introduced). I believe the implementation is mostly sound.

@anacierdem for me it would be good to merge. If you think you will have time to do an implementation review in the next few days we can hold it, otherwise I think I'd rather get it in now so that we can get on with XM player (which is hopefully a much simpler review).

include/mixer.h
snacchus: > I think that it's the correct choice to use bits in public API and bytes internally. It might be less uniform, but I think it's clearer for users.

That's fair enough. I was thinking about the API more as a programmer I guess 😄

rasky:
I realized I hadn't updated the example Makefile to latest bells & whistles, nor the CI failed because I forgot to add the Makefile to the examples/Makefile. I have fixed it now and verified the CI compiles audioconv64 and then the mixertest correctly, with asset conversion.

tools/Makefile
```diff
@@ -1,13 +1,14 @@
...
-install: chksum64 ed64romconfig n64tool
+install: chksum64 ed64romconfig n64tool audioconv64
 	install -m 0755 chksum64 ed64romconfig n64tool $(INSTALLDIR)/bin
 	$(MAKE) -C dumpdfs install
 	$(MAKE) -C mkdfs install
 	$(MAKE) -C mksprite install
+	$(MAKE) -C audioconv64 install
 
 .PHONY: clean
 clean:
```
meeq, line 14: `make clean` does not clean the `audioconv64` sub-folder

rasky: thanks, fixed

examples/mixertest/mixertest.c
```diff
@@ -0,0 +1,93 @@
...
+
+        if (ckeys.c[0].B) {
+     		mixer_ch_stop(CHANNEL_SFX2);
+        }
+
+        // Check whether one audio buffer is ready, otherwise wait for next
+        // frame to perform mixing.
+        if (audio_can_write()) {    	
+	        short *buf = audio_write_begin();
+			mixer_poll(buf, audio_get_buffer_length());
```
Ryzee119, line 89: Tabs/Spaces

src/audio/mixer.c
```diff
@@ -0,0 +1,646 @@
...
+	mixer_fx64_t len;
+	/* Length of the loop in the waveform (in bytes) */
+	mixer_fx64_t loop_len;
+	/* Pointer to the waveform */
+	void *ptr;
+	/* Misc flags */
+	uint32_t flags;
+} mixer_channel_t;
+
+typedef struct rsp_mixer_channel_s {
```
Ryzee119, line 78: There's an assert on the `rsp_mixer_channel_s` struct size. Do we want to use packed to make this explicit?
As they're all 4 byte data datatypes it shouldnt really be an issue, I wouldnt hold anything up with this.

src/audio/mixer.c
```diff
@@ -0,0 +1,646 @@
...
+		if (Mixer.limits[i].max_buf_sz && bufsize[i] > Mixer.limits[i].max_buf_sz)
+			bufsize[i] = Mixer.limits[i].max_buf_sz;
+
+		assert((bufsize[i] % 8) == 0);
+		totsize += bufsize[i];
+	}
+
+	// Do one large allocations for all sample buffers
+	assert(Mixer.ch_buf_mem == NULL);
+	Mixer.ch_buf_mem = malloc(totsize);
```
Ryzee119, line 171: assert after this malloc to confirm it was successful?

src/audio/mixer.c
```diff
@@ -0,0 +1,646 @@
...
+void mixer_exec(int32_t *out, int num_samples) {
+	if (!Mixer.ch_buf_mem) {
+		// If we have not yet allocated the memory for the sample buffers,
+		// this is a good moment to do so.
+		mixer_init_samplebuffers();
+	}
+
+	tracef("mixer_exec: 0x%x samples\n", num_samples);
+
+    uint32_t fake_loop = 0;
```
Ryzee119: tabs/spaces

Ryzee119 (commented):
Sorry for the late review but I flicked over the code, which i hadnt really done previously when I was just testing the functionality.

Looks really good to me. I only found a couple things, all minor and mostly nits.

examples/mixertest/mixertest.c
rasky: Done

src/audio/mixer.c
rasky: Done. I don't think it matters per-se, as long as the RSP definition is the same. That is, even if the non-packed structure had a uint8_t, it would be sufficiente for the RSP DMEM structure to follow it with its packing. OTOH, RSP allows misaligned memory accesses, so packed saves DMEM, doesn't slow down the RSP, and doesn't force the programmer to think of padding rules. So all in all I think you're right that it's better.

src/audio/mixer.c
rasky: Done

src/audio/mixer.c
rasky: Done

PR #179, opened 2021-09-26
test: in test_timer, be more lenient on timing for emulators

rasky:


tests/test_timer.c
```diff
@@ -231,7 +231,10 @@ void test_timer_ticks(TestContext *ctx) {
 		// Check that all ticks are monotonically increasing, that there are
 		// no large jumps (eg: high part incremented twice), and that the
 		// timer callback was called (if it was meant to).
-		ASSERT(t0 < t1 && t1 < t2 && t2 < t3 && t3 < t4 && t4 < t5 && (t5-t0)<1000 && cbcalled == use_timer, 
+		// We also check the ticks length of the test, but we're more lenient
+		// with emulators
+		const int explen = IN_EMULATOR ? 10000 : 1000;
```
anacierdem: 1000 is sometimes too little even without an emulator. I think 1200 will work fine.

rasky: changed to 2000, it doesn't need to be too strict. It's just a test to make sure there are no very huge values that would be wrong. Smallish values are ok.

PR #182, opened 2021-10-02
Implement debug filesystem for everdrive 64

anacierdem:
- Add dma_write_raw_async that only starts the transfer
- Add latency and pulse width to PI_regs_s
- Separate 64drive SD implementation to a file
- Implement initialization routine for the SD card
- Implement write and read routines
- Enable debug SD for everdrive
- Add a few new tests for longer reads/writes


src/debug.c
```diff
@@ -11,12 +11,17 @@
...
+#include "regsinternal.h"
 #include "system.h"
 #include "usb.h"
 #include "fatfs/ff.h"
 #include "fatfs/ffconf.h"
 #include "fatfs/diskio.h"
 
+// SD implementations
+#include "debug_sdfs_ed64.c"
+#include "debug_sdfs_d64.c"
```
rasky: I'd call this `debug_sdfs_64drive.c`. I know sometimes we're forced to use other shortcuts because "64drive" is not a valid identifier name, but in this case it's not a problem.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+
+	set_everdrive_sd_bitlen(8);
+	res = everdrive_sd_read_command();
+
+	// Effectively we are bitshifting the command buffer until we find a zero
+	// (start bit), followed by another zero (transmission bit)
+	// We should be able to find it in 8 bytes
+	set_everdrive_sd_bitlen(1);
+	while ((res & 0x80) != 0) {
+		if (!timeout--) return 0xFF;
```
rasky: Is `0xFF` a correct value in case of timeout, or is it supposed to be some sort of "out of band" value? In the second case, I don't see you checking for it in the callers.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+#define ED64_SD_MODE_BLOCK_READ             0x01
+#define ED64_SD_MODE_BLOCK_WRITE            0x02
+
+// Everdrive communication mode - see everdrive_sd_set_mode
+#define ED64_SD_MODE_CMD_READ           0x00
+#define ED64_SD_MODE_CMD_WRITE          0x04
+#define ED64_SD_MODE_DATA_READ          0x08
+#define ED64_SD_MODE_DATA_WRITE         0x0C
+
+static uint32_t everdrive_sd_active_mode = 0;
```
rasky: Since you're initializing this below in the init function, I'd say to remove the `= 0` which is confusing because `0` is actually a valid value (ED64_SD_MODE_CMD_READ)

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+
+void everdrive_sd_write_command(uint8_t val) {
+	everdrive_sd_set_mode(ED64_SD_MODE_CMD_WRITE);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_CMD_WRITE, val);
+	everdrive_sd_busy();
+}
+
+uint8_t everdrive_sd_read_command() {
+	everdrive_sd_set_mode(ED64_SD_MODE_CMD_READ);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_CMD_READ, 0xffff);
```
rasky, line 109: Isn't this `io_write` an exact copy of what `everdrive_sd_set_mode(ED64_SD_MODE_CMD_READ);` does?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+
+static uint32_t __attribute__((aligned(16))) everdrive_sd_config = 0;
+// Sets how many bits are read/written at a time, per lane. This is equal to the
+// times the clock line is toggled per read/write. The actual data manipulated
+// depends on the mode (see everdrive_sd_set_mode)
+static void set_everdrive_sd_bitlen(uint8_t val) {
+	everdrive_sd_config &= ~ED64_SD_CFG_BITLEN;
+	everdrive_sd_config |= (val & ED64_SD_CFG_BITLEN);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_STATUS, everdrive_sd_config);
+}
```
rasky, line 92: Here you don't avoid calling the function if the bitlen has not changed. Maybe you could?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+void everdrive_sd_write_command(uint8_t val) {
+	everdrive_sd_set_mode(ED64_SD_MODE_CMD_WRITE);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_CMD_WRITE, val);
+	everdrive_sd_busy();
+}
+
+uint8_t everdrive_sd_read_command() {
+	everdrive_sd_set_mode(ED64_SD_MODE_CMD_READ);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_CMD_READ, 0xffff);
+	everdrive_sd_busy();
```
rasky, line 110: It's not clear to me why here you need to check for the busy flag. When is this busy check required?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+}
+
+void everdrive_sd_write_data(uint8_t val) {
+	everdrive_sd_set_mode(ED64_SD_MODE_DATA_WRITE);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_DATA_WRITE,  0x00ff | (val << 8));
+}
+
+uint8_t everdrive_sd_read_data() {
+	everdrive_sd_set_mode(ED64_SD_MODE_DATA_READ);
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_DATA_READ,  0xffff);
```
rasky, line 123: Same doubt here about this `io_write` being a duplicate.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+// Wait for and read the first byte
+uint8_t everdrive_sd_read_first() {
+	uint8_t res;
+	uint32_t timeout = 64; // 8 bytes * 8 bits
+
+	set_everdrive_sd_bitlen(8);
+	res = everdrive_sd_read_command();
+
+	// Effectively we are bitshifting the command buffer until we find a zero
+	// (start bit), followed by another zero (transmission bit)
```
rasky, line 135: The comment doesn't seem to match the code. The loop below exits when the MSB (0x80) is zero, but it doesn't seem to check the bit after that.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	return io_read(ED64_BASE_ADDRESS + ED64_REGISTER_SD_BASE + ED64_SD_MODE_DATA_READ);
+}
+
+// Wait for and read the first byte
+uint8_t everdrive_sd_read_first() {
+	uint8_t res;
+	uint32_t timeout = 64; // 8 bytes * 8 bits
+
+	set_everdrive_sd_bitlen(8);
+	res = everdrive_sd_read_command();
```
rasky: Why here you're reading 8 bits and then checking also the MSB? It would seem to me that you should get into `bitlen(1)` immediately, or alternatively check all bits read here.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	while ((res & 0x80) != 0) {
+		if (!timeout--) return 0xFF;
+		res = everdrive_sd_read_command();
+	}
+	return res;
+}
+
+static uint8_t everdrive_sd_resp_buff[5] = { 0 };
+// Wait for and read an RX like response
+static uint8_t everdrive_sd_read_RX() {
```
rasky: It looks like `everdrive_sd_resp_buff` is actually used as value returned by `everdrive_sd_read_RX`. So I think it would be better to define this function as `everdrive_sd_read_response(uint8_t resp[5])`. It would make code in callers clearer.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	// read remaining bytes
+	set_everdrive_sd_bitlen(8);
+	everdrive_sd_resp_buff[1] = everdrive_sd_read_command();
+	everdrive_sd_resp_buff[2] = everdrive_sd_read_command();
+	everdrive_sd_resp_buff[3] = everdrive_sd_read_command();
+	everdrive_sd_resp_buff[4] = everdrive_sd_read_command();
+
+	// Make sure everything is consumed, we just don't use them
+	while(everdrive_sd_read_command() != 0xFF) {
+		if (!timeout--) return 0xFF;
```
rasky: What about this `0xFF`? Maybe you can put a boolean in the return value instead to signify success/failure and then let the caller handle it appropriately.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	set_everdrive_sd_bitlen(8);
+
+	everdrive_sd_write_command(0xff);
+	everdrive_sd_write_command(cmd);
+	everdrive_sd_write_command(arg >> 24);
+	everdrive_sd_write_command(arg >> 16);
+	everdrive_sd_write_command(arg >> 8);
+	everdrive_sd_write_command(arg);
+
+	// LS bit must always be 1
```
rasky: I'm not used to the notation `LS bit`. I read that like there is a bit called "LS" in the reference manual (line signal, or whatever). For least-significant bit, I usually see "LSB", or "bit 0".

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	everdrive_sd_resp_buff[4] = everdrive_sd_read_command();
+
+	// Make sure everything is consumed, we just don't use them
+	while(everdrive_sd_read_command() != 0xFF) {
+		if (!timeout--) return 0xFF;
+	};
+	return everdrive_sd_resp_buff[0];
+}
+
+static uint8_t everdrive_sd_execute_command(uint8_t cmd, uint32_t arg) {
```
rasky: There are lots of `uint8_t` around here, maybe we could use a enum for the commands?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	uint8_t result;
+	if (
+		(everdrive_sd_active_mode & ED64_SD_MODE_ACCESS) == mode &&
+		everdrive_sd_address == addr
+	) {
+		return 0;
+	}
+
+	result = everdrive_sd_stop_transmission();
+	if (result & ED64_SD_TIMEOUT) return result;
```
rasky: Why calling stop transmission here? What does that function do?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+
+	// Set lo speed for initialization
+	everdrive_sd_config &= ~ED64_SD_CFG_SPEED;
+	io_write(ED64_BASE_ADDRESS + ED64_REGISTER_SD_STATUS, everdrive_sd_config);
+
+	// Put in idle
+	everdrive_sd_execute_command(ED64_SD_CMD0, 0);
+
+	// IF cond
+	result = everdrive_sd_execute_command(ED64_SD_CMD8, 0x1AA);
```
rasky: What is 0x1AA? Maybe add some comment.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+	int num_retries = ED64_SD_ACMD41_TOUT_MS / ED64_SD_ACMD41_WAIT_MS;
+	while(1) {
+		if(!num_retries--) {
+			debugf("SD card did not respond\n");
+			return RES_ERROR;
+			break;
+		}
+
+		// Query with HCS and 3.2-3.4V
+		result = everdrive_sd_send_app_command(ED64_SD_CMD41, 0, 0x40300000);
```
rasky: For instante, here you're not testing result for timeout, so in the next line you might be testing the response buffer from a previous command.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+#define ED64_SD_CMD2                    0x42 // All send CID
+#define ED64_SD_CMD3                    0x43 // Send relative addr
+#define ED64_SD_CMD6                    0x46
+#define ED64_SD_CMD7                    0x47 // Select/deselect card
+#define ED64_SD_CMD8                    0x48 // Send interface condition
+#define ED64_SD_CMD12                   0x4C // Stop transmission on multiple block read
+#define ED64_SD_CMD18                   0x52 // Read multiple block
+#define ED64_SD_CMD25                   0x59 // Write multiple block
+#define ED64_SD_CMD55                   0x77 // Application specific cmd
+#define ED64_SD_CMD41                   0x69
```
rasky, line 36: I think I saw that the commands are really called "CMDnn" but do you think it's better to keep them like this for consistency? Because the code is really unreadable with these names, you don't understand what a command does without coming here for the comment.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+		// start the transfer.
+		set_everdrive_sd_bitlen(1);
+		do {
+			if (!timeout--) {
+				PI_regs->dom1_pulse_width = old_pw;
+				debugf("Data token timeout\n");
+				return RES_ERROR;
+			}
+			result = everdrive_sd_read_data();
+		} while (result != 0xf0);
```
rasky, line 474: There are many of these loops in the code. Maybe you could factor them out into a function?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,590 @@
...
+		data_cache_hit_writeback_invalidate(buff, 512);
+		dma_read(buff,  ED64_BASE_ADDRESS + ED64_SD_IO_BUFFER, 512);
+
+		// TODO: actually check the CRC?
+		data_cache_hit_writeback_invalidate(crc, 8);
+		dma_read(crc, ED64_BASE_ADDRESS + ED64_SD_IO_BUFFER, 8);
+		buff += 512;
+	}
+
+	everdrive_sd_address = sector + count;
```
rasky, line 494: This is not updated on the failure paths. Maybe you should increment it everytime you read a sector instead, so that it's always correct even if the function aborts at some point.

src/dma.c
```diff
@@ -95,6 +95,44 @@ void dma_read_raw_async(void * ram_address, unsigned long pi_address, unsigned l
...
+ * See #dma_read_async for a higher level primitive which can perform almost
+ * arbitrary transfers.
+ *
+ * @param[out] ram_address
+ *             Pointer to a buffer to place read data (must be 8-byte aligned)
+ * @param[in]  pi_address
+ *             Memory address of the peripheral to read from (must be 2-byte aligned)
+ * @param[in]  len
+ *             Length in bytes to read into ram_address (must be multiple of 2)
+ */
```
rasky, line 115: Tne comment needs to be updated

rasky (commented):
Great work! The CRC swizzling is insane, I guess it took a while to figure it out :) I am so happy that the debugging library is finally getting ED64 support.

src/debug_sdfs_ed64.c
anacierdem: Yes it is. From what I can tell, the first write to that register switches the mode and then we need to write again to tell we are going to "read" a value, which probably does the actual clock toggling to read the value from SD. That is just my own interpretation though and I know this does not work without that second write.

Another thing I realized now is that the reference code does the mode switch with bitlen set to 0 (it also keeps the old cfg reg and restores it afterwards) which I don't do. I just tried doing that as well but it didn't work b/c of some timing issue with my code. Although writing a few high bits to the lanes is not a problem I will try to fix that as well.

src/debug_sdfs_ed64.c
anacierdem: I honestly do not know. This is part of the hardware design and I was not able to decipher this. Seems like it is only required for cmd read/write

src/debug_sdfs_ed64.c
anacierdem: This is how it works really 🤷‍♂️

src/debug_sdfs_ed64.c
anacierdem: Nice catch! This was a leftover from when I though this was SPI.

src/debug_sdfs_ed64.c
anacierdem: By setting the bit length to 8 we shift a byte immediately and then we start checking one by one. This saves a few cycles and is probably more readable.

 e.g
```
xxxx xxxx 1234 5678
\_______/ \_______/
  our      new response from card
buffer

After first 8 bit shift;

1234 5678 90ab cdef
\ we check these bits

After a single shift;

2345 6789 0abc defg
\ we check these bits
```
So this way we just dispose our initial 8 bits which , we don't care.

src/debug_sdfs_ed64.c
anacierdem: Once a block read/write command is issued, the SD card continuously sends data as the clock is toggled. To exit from this mode, we dispatch a `CMD12`. I will try to make this clearer.

src/debug_sdfs_ed64.c
anacierdem: Thanks! This was an oversight by me.

src/debug_sdfs_ed64.c
anacierdem: The issue is that they can mean different things when issued after a `CMD55` THen they become `ACMDnn` That's why I don't even have the comments here for `ED64_SD_CMD41` and `ED64_SD_CMD6` here. I think this is easier for when following them on the spec as well. I generally did the navigation in the spec :)

src/debug_sdfs_ed64.c
anacierdem: I decided to go with a cleanup routine instead.

src/dma.c
anacierdem: Hmm I did but reverted it at some point, will fix

src/debug_sdfs_ed64.c
anacierdem: That's good advice but there are slight nuances that makes it a little convoluted. So I'll leave them as is for now.

src/debug.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done

src/debug_sdfs_ed64.c
anacierdem: Done! This was something left from SPI days as well. bool makes better sense atm.

anacierdem:
Thank you for the in depth review! It really take some while to figure it out and your instant intuition on what's going on fixed my assumption that ed64 was accessing the card in SPI mode. Once that cleared out, it was just a few nightmarish days to implement 😄

src/debug_sdfs_ed64.c
rasky: OK maybe leave a comment that explains that, even this looks like an exact duplicate of the mode switch command, it's actually required to begin reading.

src/debug_sdfs_ed64.c
rasky: My doubt is that of the initial 8 bits, you actually do check the MSB only and ignore the other 7. When the loop start, `res` is the value read with bitlen(8).

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,598 @@
...
+	while(everdrive_sd_read_command() != 0xFF) {
+		if (!timeout--) {
+			return false;
+		}
+	};
+	return true;
+}
+
+static uint8_t everdrive_sd_resp_buff[5] = { 0 };
+static bool everdrive_sd_execute_command(uint8_t cmd, uint32_t arg) {
```
rasky: Just like before, can't the response just be an argument to everdrive_sd_execute_command? I think I don't see a value in having a global `everdrive_sd_resp_buff` that can become stale or being accessed out-of-order. The response of a command should be a local variable IMO.

src/debug_sdfs_ed64.c
anacierdem: No, we don't ignore them. When the loop start, `res` is a whole byte read but each 1 bit read shifts only a single bit into `res` from the right thus in each iteration we sequentially check each bit of that word. Am I missing something here?

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,598 @@
...
+	everdrive_sd_address = sector + count;
+
+cleanup:
+	io_write((uint32_t)&PI_regs->dom1_pulse_width, old_pw);
+	if (ret_val == RES_OK) return RES_OK;
+
+	// Do error cleanup
+	everdrive_sd_change_mode(ED64_SD_MODE_NONE, 0);
+
+	return ret_val;
```
rasky, line 506: In this cleanup, you don't change `everdrive_sd_address`. It looks to me that if you call this function to read 4 blocks, and after 2 blocks the reading goes into timeout, `everdrive_sd_address` will still point to the initial block, while in reality two blocks were read and the current address is probably the third block. Does this make any sense?

I'm worried in general of these global caches. I mean they're fine to minimize redundant hardware accesses but they must be very carefully updated on every code path.

src/debug_sdfs_ed64.c
```diff
@@ -0,0 +1,598 @@
...
+
+cleanup:
+	io_write((uint32_t)&PI_regs->dom1_pulse_width, old_pw);
+
+	if (ret_val == RES_OK) return RES_OK;
+
+	// Do error cleanup
+	everdrive_sd_change_mode(ED64_SD_MODE_NONE, 0);
+
+	return ret_val;
```
rasky, line 609: Same here

rasky (approved):
Thanks. I've left a comment to move the response buffer further down the chain into local stack buffers. That's all on my side. The hardware does seem to have weird corners but we have to live with it.

src/debug_sdfs_ed64.c
rasky: My understanding is that bitlen(8) + read means "extract 8 bits from the shift register". If this understanding is correct, then first read is reading 8 bits from the queue. The loop is checking the MSB. Then in the loop you call bitlen(1) + read which I think it means "extract 1 bit from the shift register", so it will extract the first bit that follows the 8 you have just extracted. This means that of the first 8 bits, you only checked 1 bit.

In your ASCII example in the comment above, it seems like you must completely discard and ignore the first 8 bit. In that case, you probably want to either reset res before the loop, or change the loop into a `do/while` so that the condition is at the end of the loop so that immediately uses the 1-bit read.

src/debug_sdfs_ed64.c
anacierdem: Ah! Couldn't think of that, sure I will do this!

src/debug_sdfs_ed64.c
anacierdem: Yes, that makes sense but `everdrive_sd_change_mode` already sets `everdrive_sd_address` to zero and changes the mode into `ED64_SD_MODE_NONE` which always require a sector restart anyways. `everdrive_sd_change_mode` will also stop the transmission in that case. Here I assume if something breaks, we have lost the current sector and start over.

In general I agree but especially for the multi block transmission, we have to keep the state as it is aimed at reducing successive SD commands. o/w we would use single block reads and update the sector before each read.

src/debug_sdfs_ed64.c
anacierdem: Added a comment to be a little more clearer.

rasky:
Another thing I noticed today is that most of your read/write reg functions are duplicates of similar functions in usb.c. It's all code that gets brought in only for debugging so maybe duplicate code doesn't matter much, but I think in general it would be better to avoid this. The 64drive SD code used functions from usb.c for this reason.

In your case it looks like your versions are actually better (eg PI accesses rather than DMA) so it would be better to bring your implantation to usb.c and submit it upstream.

As I said, not super important

PR #184, opened 2021-10-03
Update makefiles to work properly with custom $N64_INST

Hazematman:
These changes fix the following issues:

1. Update n64.mk to install using `sudo` so if you do have permission to write to $PREFIX you will be able to write there
2. Update the RSP build command to use $ASFLAGS instead of $N64_ASFLAGS. This fixes an issue I saw when trying to build @snacchus ugfx code with a custom $N64_INST path. $N64_ASFLAGS does not include the additional -I$(CURDIR)/include so rsp code that is in the libdragon library tree won't be able to find rsp.inc
3.  Another problem seen with ugfx, if you're building rsp code in the libdragon library tree, when it creates the section names it will use the whole $BUILD_DIR path in the section name. If the path name has periods in it, gcc will make those underscores instead. The make file needs to substitute periods with underscores like it does for slashes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ install-mk: n64.mk
 	# n64.mk must be installed into $PREFIX, otherwise make will not be
 	# able to find it later with "include n64.mk". This is the only file that
 	# must go outside of $INSTALLDIR.
-	install -Cv -m 0644 n64.mk $(PREFIX)/include/n64.mk
+	sudo install -Cv -m 0644 n64.mk $(PREFIX)/include/n64.mk
 
 install: install-mk libdragon
 	install -Cv -m 0644 libdragon.a $(INSTALLDIR)/mips64-elf/lib/libdragon.a
--- a/Makefile
+++ b/Makefile
@@ -59,9 +59,9 @@ tools-clean:
 
 install-mk: n64.mk
 	# n64.mk must be installed into $PREFIX, otherwise make will not be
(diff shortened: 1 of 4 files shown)


Hazematman:
For problem number 1, I've modified it to instead install the n64.mk file in `$(INSTALLDIR)/include/n64.mk` and updated the examples to use 
```
include $(N64_INST)/include/n64.mk
```

We could also updated it to install to `$(INSTALLDIR)/n64.mk` and update the examples to use
```
include $(N64_INST)/n64.mk
```

Although this makes the include line in the make file simpler, the default `$N64_INST` is `/usr/local` and this would pollute the root of `/usr/local` with the `n64.mk` file. Therefore I think we should go with the longer path that includes the "include" folder.

Makefile
```diff
@@ -59,9 +59,9 @@ tools-clean:
 
 install-mk: n64.mk
 	# n64.mk must be installed into $PREFIX, otherwise make will not be
-	# able to find it later with "include n64.mk". This is the only file that
+	# able to find it later with "include $(N64_INST)/include/n64.mk". This is the only file that
 	# must go outside of $INSTALLDIR.
```
anacierdem, line 63: The comment needs to change here

anacierdem (commented):
There is a minor incorrect comment, I will merge once that is fixed.

anacierdem:
I'm removing that comment in the docker PR.

PR #185, opened 2021-10-04
Fix tests Makefile & build tests in build.sh as well

anacierdem:


danbolt:
Thanks for the update; will give this a go soon!

PR #186, opened 2021-10-06
Improvements to n64.mk

meeq:
* Minor internal consistency fixes
* Add support for `N64_ED64ROMCONFIGFLAGS`
* Migrate eepromfstest example to use n64.mk and demonstrate `ed64romconfig` (@Ryzee119)
* Improve readability of existing n64.mk usage


examples/eepromfstest/Makefile
```diff
@@ -1,35 +1,26 @@
-PROG_NAME = eepromfstest
+BUILD_DIR = build
+include $(N64_INST)/include/n64.mk
 
-ROOTDIR = $(N64_INST)
-GCCN64PREFIX = $(ROOTDIR)/bin/mips64-elf-
+.DEFAULT_GOAL := all
```
rasky: I would say this is superfluous. It's also not very common (I for one didn't know that it existed), so I would left it out from examples that should show how to compactly achieve a build.

examples/eepromfstest/Makefile
```diff
@@ -1,35 +1,26 @@
...
-N64TOOLFLAGS = -l 1M -h $(ROOTDIR)/mips64-elf/lib/header -t "EEPROMFS Test"
+all: eepromfstest_4k.z64 eepromfstest_16k.z64
+.PHONY: all
 
-ifeq ($(N64_BYTE_SWAP),true)
-$(PROG_NAME).v64: $(PROG_NAME).z64
-	dd conv=swab if=$^ of=$@
-endif
+eepromfstest_4k.z64: N64_ROM_TITLE = "EEPROMFS Test 4k"
+eepromfstest_4k.z64: N64_ED64ROMCONFIGFLAGS = --savetype eeprom4k
```
rasky: I think the precedent of N64_ROM_TITLE was an attempt to hide the various step of ROM building to beginners (more about this later).

I would prefer to see a N64_SAVETYPE=eeprom4k here. You can still have a N64_ED64ROMCONFIGFLAGS behind the hood that by default uses N64_SAVETYPE, so that advanced users can change it, but for simple use cases I think just specifying the save type here would be easier

n64.mk
```diff
@@ -40,16 +41,21 @@ ASFLAGS+=-MMD    # automatic .d dependency generation
 %.z64: CFLAGS+=$(N64_CFLAGS)
 %.z64: ASFLAGS+=$(N64_ASFLAGS)
 %.z64: LDFLAGS+=$(N64_LDFLAGS)
-%.z64: $(BUILD_DIR)/%.elf
-	@echo "    [N64] $@"
-	$(N64_OBJCOPY) $< $<.bin -O binary
+%.z64: $(BUILD_DIR)/%.bin
+	@mkdir -p $(dir $@)
```
rasky: Well technically this rule depends on a path already in BUILD_DIR...

n64.mk
```diff
@@ -40,16 +41,21 @@ ASFLAGS+=-MMD    # automatic .d dependency generation
...
+	@echo "    [N64TOOL] $@"
 	@rm -f $@
-	DFS_FILE=$(filter %.dfs, $^); \
+	DFS_FILE="$(filter %.dfs, $^)"; \
```
rasky, line 60: Ah good catch!

n64.mk
```diff
@@ -40,16 +41,21 @@ ASFLAGS+=-MMD    # automatic .d dependency generation
 %.z64: CFLAGS+=$(N64_CFLAGS)
 %.z64: ASFLAGS+=$(N64_ASFLAGS)
 %.z64: LDFLAGS+=$(N64_LDFLAGS)
-%.z64: $(BUILD_DIR)/%.elf
-	@echo "    [N64] $@"
-	$(N64_OBJCOPY) $< $<.bin -O binary
+%.z64: $(BUILD_DIR)/%.bin
+	@mkdir -p $(dir $@)
+	@echo "    [N64TOOL] $@"
```
rasky: In general I think I'd prefer to hide the different tools both from beginner users and from the progress. They are just an internal implementation detail on how today libdragon does it with different tool. We could decide to merge everything into n64tool, or remove some parts (eg: checksum might not be needed in a OSS IPL3). I don't think it also gives a value to most builds having 3 different lines rather than one. For most users, this will just be a final step to convert the linker output to the ROM format, and that's probably enough. People can still look into this with V=1 if they want so we are not removing debugging power for instance. But in "default silent mode", I think one line is enough.

n64.mk
```diff
@@ -98,12 +104,17 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c
...
 	@mkdir -p $(BUILD_DIR)
 	@echo "    [LD] $@"
-	$(LD) -o $@ $(filter-out $(N64_ROOTDIR)/mips64-elf/lib/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
+	$(LD) -o $@ $(filter-out $(N64_LIBPATH)/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
 	$(N64_SIZE) -G $@
 
+%.bin: %.elf
+	@mkdir -p $(dir $@)
+	@echo "    [OBJCOPY] $@"
+	$(N64_OBJCOPY) $< $@ -O binary
```
rasky: I have a similar concern with this. I don't think the bin has a value per-se to warrant a differ Makefile target. Just like while building rSP
ucode, I'd keep it folded in the N64 step

rasky (commented):
Thanks!

anacierdem:
I have one ask: can you also remove `PREFIX ?= /usr/local`, which I forgot to do?

n64.mk
```diff
@@ -41,22 +45,23 @@ ASFLAGS+=-MMD    # automatic .d dependency generation
...
 	fi
 	$(N64_CHKSUMPATH) $@ >/dev/null
 
-# Support v64 ROMs via dd byteswap
-ifeq ($(N64_BYTE_SWAP),true)
 %.v64: %.z64
-	dd conv=swab if=$^ of=$@
-endif
+	@echo "    [V64] $@"
+	$(N64_OBJCOPY) -I binary -O binary --reverse-bytes=2 $< $@
```
meeq, line 73: This change makes the V64 byte-swap conversion more portable. Also, there's no need to gate it behind `N64_BYTE_SWAP` anymore. The V64 target will only get built if specifically requested.

anacierdem: That is not true for the automated builds right now. They dont have a way of doing it other than providing a variable. I dont think byteswap is necessary but this is something to keep in mind. e.g I would need to remove byteswap flag from the docker wrapper as well, though Im ok with removing them.

meeq: As of the move to `n64.mk` the `.v64` target is no longer the first target in the Makefile (even if `N64_BYTE_SWAP` is true) so the behavior of `N64_BYTE_SWAP` is already broken/changed.

anacierdem: Yes it creates both of them IINM but removing it alltogether is something else. Still Im OK to remove the condition but it will be changing the behaviour if someone is depending on the build creating a byteswapped rom.

anacierdem: Hmm I was mistaken, it is already broken. As no one cared up to now, this is totally ok :)

Ryzee119:
Thanks for doing this @meeq :)

Ive tested your changes by settings various types to N64_ROM_SAVETYPE. Results confirmed in hex editor on output rom files.
Working as expected!

PR #188, opened 2021-10-08
Add a high-performance XM module player

rasky:
This PR adds the XM64 player, a fully-featured player for the XM module format (originated back in the 90's with Fast Tracker II).

The implementation is based on the open source [libxm](https://github.com/Artefact2/libxm), a very good and neat XM player by @Artefact2, to which @bryc and I contributed also [several accuracy and compatibility fixes](https://github.com/Artefact2/libxm/commits?author=rasky) (thank you @Artefact2 for the quick reviews!). We tested it against ~700 XM modules and the current version basically provides perfect playback on most of those, with just a few very minor glitches on a few.

Once you sit on top of libxm, the implementation that binds it to the new mixer is remarkably easy (see `tick()` in xm64.c), as the mixer provides a very easy API to do sequencing. There is no custom RSP code to be written, since the mixer itself is providing enough features (mainly, loops and resampling). The XM player uses one mixer channel per XM channel.

An example ROM is provided with a few modules to test the behavior of the player. A conversion step must be done (XM -> XM64) to adapt the module to a few limits of the current RSP ucode; it's been integrated into the `audioconv64` tool.

Performance is extremely good. XM64 files up to 10-12 channels are usually played back using 2-5% of CPU, and about 10% of RSP time, which, together with module compactness, makes the format really a good choice for background music in games, without removing too many resources from the graphics. 

RAM usage is very little. Thanks again to the mixer API, samples are streamed from ROM via DMA at each tick, and the conversion tool computes the exact minimum samplebuffer size required for each module channel, which is then configured into the mixer. Even large modules (1Mb+) rarely need more than 100Kb in RDRAM.

I would like to thanks especially @bryc and MishtaLu for having provided lots of insights, help and tests during the development in the past months.

Here you can find the new example ROM that can be compiled with this PR, which includes a few demonstration XM modules.
[audioplayer.z64.zip](https://github.com/DragonMinded/libdragon/files/7312718/audioplayer.z64.zip)


include/xm64.h
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * @brief Open a XM64 module file and prepare for playback.
+ * 
+ * This function requires the mixer to have been already initialized
+ * (via mixer_init).
+ * 
+ * @param player Pointer to the xm64player_t player structure to use
+ * @param fn Filename of the XM64 in ROM.
+ */
+void xm64player_open(xm64player_t *player, const char *fn);
```
Ryzee119, line 63: When testing, putting "rom:/" (i.e `xm64player_open(&mx, "rom:/mysong.xm64");`) in the filename caught me off gaurd. Worth adding a comment here that its not required?

src/audio/xm64.c
```diff
@@ -0,0 +1,231 @@
...
+	uint32_t base_rom_addr = dfs_rom_addr(fn);
+
+	// Count samples
+	int ninst = xm_get_number_of_instruments(player->ctx);
+	int nwaves = 0;
+	for (int i=0;i<ninst;i++)
+		nwaves += xm_get_number_of_samples(player->ctx, i+1);
+
+	// Allocate waveforms (one per XM64's "samples" aka waveforms)
+	player->waves = malloc(sizeof(waveform_t) * nwaves);
```
Ryzee119, line 115: Assert on malloc?

examples/audioplayer/Makefile
```diff
@@ -0,0 +1,29 @@
+BUILD_DIR=build
+include $(N64_INST)/include/n64.mk
+
+src = audioplayer.c
+assets = $(wildcard assets/*.xm)
```
Ryzee119: Didnt pick up the xm files that had a capital .XM extension (tested on win10 with docker image).
Didnt investigate where is coming from (Makefile or audioconv tool)

Ryzee119 (commented):
very cool! works wonderfully on my gamejam proto under some other heavier loads

As mentioned on discord I hit this assert when playing a random XM file I found online. The converted file happened to be just over 1MB if that could cause it.

File here: https://modarchive.org/index.php?request=view_by_moduleid&query=60441
```
ASSERTION FAILED: buf->widx + wlen <= buf->size
file "/libdragon/libdragon/src/audio/samplebuffer.c", line 151, function: samplebuffer_append
samplebuffer_append: buffer too small
ridx:0 widx:0 wlen:1588 size:ac4
```

include/xm64.h
rasky: That's actually an interesting point I've been thinking of. The `rom:/` prefix is required to use stdio functions like `fopen`, but I wonder whether we should settle on a policy on whether libdragon APIs require it or not. Requiring them would make the APIs more generic (eg: they would work also on files on SD) but at the same time "forces" the usages of the newlib layer which has overhead that might not always be needed.

This specific case is quite a corner case: the xmplayer is based on a third party library that use `fopen()` so for compatibility I have to feed it `rom:/` paths, but at the same time the implementation of xmplayer (and wav64) is based on `dma_read` that can only work on ROMs. On the other hand, it's true that both wav64 and xmplayer could have a fallback to issue bare `fread()` calls to read samples for non-`rom:/` files, though I wonder whether that would be fast enough to do audio streaming.

I don't know, it's an open discussion point I think. @anacierdem @Hazematman @snacchus any idea?

snacchus: Always requiring the prefix would avoid some potential confusion I think. If you never have to ask yourself "Is the prefix required for this function or not?" it just makes life a bit easier. Maybe add a "shortcut" that circumvents the newlib layer if the prefix is `rom:/` ? I guess that would defeat the purpose of the generic file API in a way though.

rasky: We do need the bypass in case performance is critical. When doing I/O accesses at runtime within a game, you don't want to go through the newlib layer and do the memcpy if not required. I can measure the impact on the xm player if we want, but in general I'm a strong believer in "everything adds up" in embedded systems, so I'd rather not add gratuitous overhead. I'd also notice that 99% of fopen() accesses will always be to ROM anyway, and it sounds wrong to make them slower just because somebody somewhere might access a SD every once in a moonlight.

So the above is to justify the bypass. If we agree that the bypass is the right thing at least for code running at runtime (rather than load time), I would be OK with a policy that all APIs in libdragon that accepts a filename must accept it with a protocol specification, though I am still 50-50 on this.

examples/audioplayer/Makefile
rasky: Right, it's a little inconvenient because I have to duplicate the conversion rules as well, as for Makefile they are different files... not sure about this

Ryzee119: Its a minor thing anyway. I'd be happy with just a comment on the example rom or something. 

Where I got my xm files (modarchive.org) is a mix bag of .xm or .XM so I likely wouldn't be the last person to encounter this.

rasky:
> As mentioned on discord I hit this assert when playing a random XM file I found online. The converted file happened to be just over 1MB if that could cause it.

Thanks, I had a bug in the way the XM player conferred the optimal sample buffer size in the mixer. See fc3d2f7 for the fix, and b2a0d8f for a followup where I'm starting to enforce frequency limits on channels.

include/xm64.h
snacchus: Which APIs would actually be affected by this? So far I could only find `xm64player_open` and `wav64_open`. And I guess the same would apply to any future audio formats. We could also add a `sprite_open` function maybe? So it seems that only functions which are responsible for opening specialised file formats would be affected, correct? In that case I would be fine with this.

anacierdem: What if we require the `rom:/` prefix for such cases but not put it through newlib? I mean we can do that just for compatibility and assert if it is not in `rom:/`. Another idea: we allow any prefix, skip newlib if it is on rom thus we can have a crude version where it might be possible to debug things from sd etc. In the end we can just remove all the checks and only allow rom in an `NDEBUG` build?

Ryzee119:
> Thanks, I had a bug in the way the XM player conferred the optimal sample buffer size in the mixer. See [fc3d2f7](https://github.com/DragonMinded/libdragon/commit/fc3d2f75a2a19e53ce52e810561e0b00a467ba3f) for the fix, and [b2a0d8f](https://github.com/DragonMinded/libdragon/commit/b2a0d8f861edc181276927ab59faf749da5a5862) for a followup where I'm starting to enforce frequency limits on channels.

Can confirm that fixed the issue!

include/xm64internal.h
```diff
@@ -0,0 +1,9 @@
+#ifndef __LIBDRAGON_XM64_INTERNAL_H
+#define __LIBDRAGON_XM64_INTERNAL_H
+
+// This header file is not installed, and we want to keep the original libxm
+// header files separate for vendoring purposes. So just include them here.
+#include "../src/audio/libxm/xm.h"
+#include "../src/audio/libxm/xm_internal.h"
+
+#endif
```
snacchus: Do we need this file? As far as I can tell, it's only included from `conv_xm64.c`, which includes some source files from libxm that in turn include the files listed here anyway.

Also, as a more general note: We have a couple of header files like this (generally like `<something>_internal.h`), that aren't supposed to be part of the public API but are still placed in `include`. We don't need to do this from this PR, but wouldn't it make more sense to place them in `src`, because they are only ever used from source files anyway? Or do we want to uphold this strict requirement that header files go in `include`, and source files in `src`?

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
+#include <libdragon.h>
+#include "libxm/xm.h"
+#include "libxm/xm_internal.h"
+#include <stdbool.h>
+
+static void wave_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wlen, bool seeking) {
+	xm_sample_t *samp = (xm_sample_t*)ctx;
+
+	extern void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wlen, int bps);
```
snacchus: Maybe expose this in `wav64internal.h` instead?

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+		if (ch->sample) {
+			waveform_t *w = ch->sample->wave;
+
+			// Check if this sample is muted. This is an user-level muting
+			// control exposed via the xm.h API that we respect in case the
+			// user wants to mute some channels (usually for debugging).
+			bool muted = ch->muted || ch->instrument->muted;
+
+			// Play the waveform. Notice that the waveform might already
+			// being played in this channel, in which case the play
```
snacchus: being played -> be playing

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+	player->waves = malloc(sizeof(waveform_t) * nwaves);
+	assert(player->waves);
+	int nw = 0;
+	for (int i=0;i<ninst;i++) {
+		xm_instrument_t *inst = &player->ctx->module.instruments[i];
+		for (int j=0;j<inst->num_samples;j++) {
+			xm_sample_t *samp = &inst->samples[j];
+
+			// Convert offset of samples from relative to absolute
+			samp->data8_offset += base_rom_addr;
```
snacchus, line 124: What does this do exactly? If I understand correctly, then this means that the XM player only supports files in `rom:/` at the moment. This is fine for now I think, but out of interest: do you plan to support other filesystems as well?

This also relates to the other discussion: If we decide to always require the filesystem prefix in APIs, that doesn't necessarily mean that all APIs need to support all filesystems, right? What happens atm if someone calls `xm64player_open` with a path that doesn't point to `rom:/`?

snacchus (commented):
Looks good to me! We mainly need to decide on whether to require the filesystem prefix, and whether the XM player should support files that are not in `rom:/` (see my review comments). I commented on some other minor things as well.

include/xm64.h
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * @brief Start playing the XM64 module.
+ * 
+ * Notice that the player needs to use one mixer channel per each XM64 channel.
+ *
+ * @param player 	XM64 player
+ * @param fist_ch 	Index of the first mixer channel to use for playback.
+ */
+void xm64player_play(xm64player_t *player, int first_ch);
```
anacierdem, line 80: Would it make sense to do channel mapping instead of always using them in order?
Also is it possible to use an already used channel if you accept managing clipping by setting correct volumes? (I do not know how the mixer works ATM)

include/xm64.h
rasky: > What if we require the rom:/ prefix for such cases but not put it through newlib?

That's what I call "bypass" above. The problem with this is that you end up with two different codepaths that you need to implement (with and without newlib). But by doing so, you get maximum performance in the most obvious case (ROM).

src/audio/xm64.c
rasky: The quoted code converts the waveform offset from being file-relative to absolute ROM address. Currently `xm64player` only supports ROM, so if we switch to prefixed paths, I would have to assert on paths which are not `rom:/`. In the future it might be possible to add support to other filesystems like SDFS, but they need to be fast enough for audio not to crack. I think we will have to try and see if they work.

include/xm64.h
rasky: I'll note here that for `xm64player` to really work on different filesystem, ti's not only a matter of writing a small code path where `fread` is used in place of `dma_read`. It might be that the filesystem is too slow, for instance. We will have to try and test it.

include/xm64internal.h
rasky: > Do we need this file? 

Good catch, it's a leftover, I'll remove it.

> do we want to uphold this strict requirement that header files go in include, and source files in src?

The main issue with moving internal files to src is that they cannot be included anymore by games that want to bypass the libdragon public API. On the other hand, we're now advising to always vendor libdragon locally (which is also what the CLI tool does), so maybe that's not an issue anymore. We can discuss this later in a followup PR.

src/audio/xm64.c
snacchus: Alright. Asserting on non-ROM paths is definitely fine for now.

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+			// the current position, but even if it did, xm_tick() might
+			// have changed it since last tick, because there is a XM effect
+			// to force the position in the sample. So it's better to
+			// set it every time with mixer_ch_set_pos.
+			mixer_ch_play(first_ch+i, w);
+			mixer_ch_set_pos(first_ch+i, ch->sample_position);
+
+			// Configure also frequency and volume that might have changed
+			// since last tick.
+			// debugf("vol %f %f\n", ch->actual_volume, ch->actual_panning);
```
anacierdem: nit: can remove

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+				muted ? 0 : gvol * ch->actual_volume[0],
+				muted ? 0 : gvol * ch->actual_volume[1]);
+		} else {
+			// No sample in this channel: the channel is mute. Just stop it.
+			mixer_ch_stop(first_ch+i);
+		}
+	}
+
+	// Schedule next tick according to the number of samples in this tick.
+	int delay = ceilf(ctx->remaining_samples_in_tick);
```
anacierdem, line 76: Hmm it's interesting that we can have non-integer sample, so I assume the xm library can do some sample scaling as well

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+
+	char romfn[strlen(fn)+5+1];
+	snprintf(romfn, sizeof(romfn), "rom:/%s", fn);
+	player->fh = fopen(romfn, "rb");
+	assertf(player->fh, "Cannot open file: %s", fn);
+
+	// Load the XM context
+	int sample_rate = audio_get_frequency();
+	assertf(sample_rate >= 0, "audio_init() and mixer_init() must be called before xm64player_open()");
+	if (xm_context_load(&player->ctx, player->fh, sample_rate) != 0) {
```
anacierdem, line 93: Yes we seem to provide the rate here, nice!

src/audio/xm64.c
anacierdem: It is good to know that assert will do an alloc and can also fail IIRC

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+			// coming from the XM "sample".
+			samp->wave = &player->waves[nw++];
+			memset(samp->wave, 0, sizeof(waveform_t));
+			samp->wave->name = strdup(fn); // FIXME: maybe better use a proper name here
+			samp->wave->bits = samp->bits;
+			samp->wave->channels = 1;
+			samp->wave->frequency = sample_rate; // fake, will be changed at every key-on
+			samp->wave->len = samp->length;
+			samp->wave->loop_len = samp->loop_type == XM_NO_LOOP ? 0 : samp->loop_length;
+			if (samp->wave->loop_len&1)
```
anacierdem: Why this is necessary?

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+
+void xm64player_stop(xm64player_t *player) {
+	// Let the mixer callback stop playing
+	player->playing = false;
+}
+
+void xm64player_tell(xm64player_t *player, int *patidx, int *row, float *secs) {
+	// Disable interrupts to try to avoid race conditions with the player
+	// running in a different thread. Notice that this is not sufficient
+	// (you would need some kind of mutex), but let's say good enough,
```
anacierdem, line 183: How can this cause an issue while the interrupts are disabled?

src/audio/xm64.c
```diff
@@ -0,0 +1,240 @@
...
+	// FIXME: we need to stop playing without racing with the audio thread.
+	// This is not correct and may crash.
+	disable_interrupts();
+	if (player->playing) {
+		mixer_remove_event(tick, player);
+		player->playing = false;
+	}
+	for (int i=0;i<player->ctx->module.num_channels;i++) {
+		mixer_ch_stop(player->first_ch+i);
+		mixer_ch_set_limits(player->first_ch, 0, 0, 0);
```
anacierdem, line 221: Why do we still need this after a `mixer_ch_stop`?

tools/audioconv64/conv_xm64.c
```diff
@@ -0,0 +1,208 @@
...
+	fseek(xm, 0, SEEK_SET);
+
+	char *xmdata = malloc(fsize);
+	fread(xmdata, 1, fsize, xm);
+
+	size_t mem_ctx, mem_pat, mem_sam;
+	xm_get_memory_needed_for_context(xmdata, fsize, &mem_ctx, &mem_pat, &mem_sam);
+
+	xm_context_t* ctx;
+	xm_create_context_safe(&ctx, xmdata, fsize, 44100);
```
anacierdem: why 44.1?

tools/audioconv64/conv_xm64.c
```diff
@@ -0,0 +1,208 @@
...
+				break;
+			case XM_PING_PONG_LOOP:
+				length = loop_end + loop_length;
+				sout = malloc(length + MIXER_LOOP_OVERREAD);
+				out = sout;
+
+				memcpy(out, s->data8, loop_end);
+				out += loop_end;
+				for (int x=0;x<loop_length;x++)
+					*out++ = s->data8[(loop_end-x-1) ^ (bps>>1)];
```
anacierdem, line 93: Why we need to switch those bits here?

examples/mixertest/mixertest.c
```diff
@@ -70,7 +70,7 @@ int main(void) {
 			music_frequency /= 1.1;
 			mixer_ch_set_freq(CHANNEL_MUSIC, music_frequency);
 		}
-		if (music && music_frequency <= 128000 && ckeys.c[0].R) {
+		if (music && music_frequency*1.1 <= 128000 && ckeys.c[0].R) {
```
anacierdem, line 73: What was this for?

anacierdem (approved):
The heavy lifting seems to be already done by the mixer at this point, great to see it does pretty much what we may need and the API looks good as well! I think this is good to go once we assert on non-rom profixed paths. Mostly questions otherwise.

include/xm64.h
rasky: > Would it make sense to do channel mapping instead of always using them in order?

It would complicate the API also for the caller, and I'm not sure how useful that is. Channel allocation on the mixer should be a static decision made once, and can be changed easily if you simply use macros (see mixertest.c).

> Also is it possible to use an already used channel if you accept managing clipping by setting correct volumes? (I do not know how the mixer works ATM)

No: a channel is able to play only one waveform at a time. You cannot play two waveforms at the same time on the same channel -- that requires mixing. If you try to play a wav64 using a channel that is being used by the xmplayer, the result will be audio truncation: for instance, the instant you start playing the wav64, an instrument played by the xmplayer will be abruptly stopped; but then, after a while (depending not he XM song), the XM player will force play another instrument on the same channel, stopping the wav64.

src/audio/xm64.c
rasky: We can have non-integer playback frequency. The mixer uses `float frequency` in the API and the equivalent fixed point internally when talking to the RSP. That allows for very accurate resampling because many times frequencies are not perfect multiples.

src/audio/xm64.c
rasky: Right, there's no comment. This is explained in `wav64_set_loop`. This is the comment there:

```
	// Odd loop lengths are not supported for 8-bit waveforms because they would
	// change the 2-byte phase between ROM and RDRAM addresses during loop unrolling.
	// We shorten the loop by 1 sample which shouldn't matter.
	// Notice that audioconv64 does the same during conversion.
```

The only real constraint for `dma_read` now is that the RDRAM and ROM addresses must have the same 2-byte phase (that is, they must be both odd or both even). The reason is that physical DMA is actually 16-bit aligned so you can only run it on even addresses, but if you have two odd addresses, dma_read will just move one byte manually before starting the transfer.

In case of audio, in general you need to fetch samples from ROM into the RDRAM samplebuffer provided by the mixer. Normally they start aligned (both ROM waveforms and RDRAM samplesbufers starts as 8-byte aligned), and you always keep this alignment as you go. When the samplebuffer wraps around, the code in samplebuffer_discard (used as part of looping around in the buffer) will make sure not to lose the 2-byte phase. So in general the phase is always kept. The only exception would be a 8-bit waveform that declares a loop of odd size. In that case, the last byte of the waveform will be correctly phases, but the following byte (beginning of the loop) would be out of phase. We don't have a solution for this, we currently do not support this scenario, and we simply shorten the loop by 1 sample.

The code you've highlighted is technically wrong because it does the shortening also for 16-bit waveforms where it is not required. I'll fix it a document it, thanks.

src/audio/xm64.c
rasky: technically, the audio thread could have been preempted at any point, eg: while modifying the XM data structures in non-atomic way. Disabling interrupts make sure the current code is not preempted, but it gives you no guaranteed on where exactly the other threads are paused.

In this case, it should work by luck because audio thread is always going to be higher priority so in general it would preempt only on specific points (eg: while waiting for DMA or RSP) so outside the actual XM business logic. Obviously if you call xm64player_tell from an even higher priority thread, all hell break loose.

The only solution here would be to wrap all calls to libxm in a disable/enable preemption, so that all modifications to the XM data structures are kept atomic. I think this is what I will try to do when I get at it.

src/audio/xm64.c
rasky: `mixer_ch_stop` just stops the channel (stops the audio). 

`mixer_ch_set_limits` is a permanent configuration of the channel that gives the mixer information on how big to allocate the samplebuffer. By default, the mixer has to be conservative and allocate buffers for each channel that are big enough for the worst case (eg: 16-bit 44.1Khz playback). The xmplayer uses this API to minimize memory consumption (the converter calculates the *exact* size required for playback on each channel).

The limits are not set for each waveform being played because it would require too many deallocations/allocations. Instead, we calculate limits which are good for the whole song and stick with them. Experimentally, this is a good enough solution, as most XM songs require < 100Kb of RDRAM total (samplebuffers plus XM data structures) so I would say it's good enough.

Here we're shutting down the player, so, to be a good citizen, the player removes any limit that have been configured in the mixer, to leave it in pristine state. Otherwise, somebody else might try to play something else on one of the channels that were previous used by the xmplayer, and fail to playback because it would find a too small buffer.

tools/audioconv64/conv_xm64.c
rasky: Uhm good question. This is used as part of the sample buffer size estimation described above. I think this is technically wrong: since we don't know what the audio playback rate would be at runtime, we should probably be conservative here and use 48000 which is the maximum playback rate supported by the AI. I will do the change. This will increase by ~10% the memory used by the XM player, but I will try to compensate by reducing a bit the 5% margin I'm sticking in to compensate for small bugs in estimation (I think I've fixed most of those already anyway).

tools/audioconv64/conv_xm64.c
rasky: Even if the array is called `data8`, it's actually a union with `data16`, that is the waveform begin processed can be either 8 bit or 16 bit. Here we're unrolling a ping-pong loop, so assuming the waveform is:
 
 ```
    ABCDEFGHI
```

we need to unroll it like this:

``` 
   ABCDEFGHIHGFEDCBA
```

The problem comes with 16-bit samples. In this case you start from this (one letter is one byte):

 ```
    AaBbCcDdEeFfGgHhIi
```

and you want to obtain this:

 ```
    AaBbCcDdEeFfGgHhIiHhGgFfEeDdCcBbAa
```

so you're going back with samples, but you are keeping the same byte order *within* each 16-bit sample. So as you go backward instead of copying bytes at indices `10, 9, 8, 7, 6, 5, ..`  you need to copy `9, 10, 7, 8, 5, 6, ...`. If `idx` is a variable looping through the first sequence, `idx ^ 1` produces the second sequence. So `idx ^ (bps>>1)` produces the second sequence only if `bps` is 2 (16-bit) but not if `bps` is 1 (8-bit) (these are the only two possibile values).

Obviously the alternative would be to have two loops: one copying backward 8-bit samples via `uint8_t`, and one copying backward 16-bit samples via `uint16_t` (with several casts and multiplications, or by using different pointers of uint16_t type and then having to cast them back to the original pointers).

examples/mixertest/mixertest.c
rasky: See the commit message: I found this small bug in mixertest. The test is configuring the mixer for max 128Khz playback on that channel, but then the code was wrong and was playing a frequency higher than that. This started failing with assert with this PR that contains a commit where the frequency limit is enforced via assert, rather than crossing fingers and hoping the samplebuffer is big enough anyway.

src/audio/xm64.c
rasky: Notice that we have the same problem with the mixer data structures. If the audio threads is within `mixer_poll()`, and the game thread calls any mixer function such as `mixer_ch_play`, what happens is undefined. This is something that we want to support, but I don't know how to do it. 

A mutex in this case is too big a hammer because you don't want the `mixer_ch_play` call in the game logic to block because the mixer is currently producing samples in the audio thread. The best solution would be the let `mixer_ch_play` write somewhere that a waveform must begin playing, but only doing that in the context of the audio thread. That would be something like an internal command queue, but it complicates the code quite a bit. I'm thinking of it.

tools/audioconv64/conv_xm64.c
anacierdem: Also would be nice to get it from the command arguments and default to 48k, WDYT?

examples/mixertest/mixertest.c
anacierdem: Might be nice to squash those two commits if that's ok

include/xm64.h
rasky: OK i've changed the API to always accept a filesystem-qualified name, and for now the function will assert if the prefix is not "rom:/".

src/audio/xm64.c
rasky: Done!

src/audio/xm64.c
rasky: Done

src/audio/xm64.c
rasky: Done

src/audio/xm64.c
rasky: Done

examples/mixertest/mixertest.c
rasky: Done

tools/audioconv64/conv_xm64.c
rasky: I rethought about this and I was confused. The sample buffer size does not depend on the output rate because resampling happens within the mixer. The sample buffer contains a waveform at its native playback rate. For instance, if the XM logic decides to play an instrument at 8000 Hz and the speed of XM is 25 ticks per second, for each tick I will need 8000 / 25 = 320 samples in the sample buffer, and the mixer will upscale those to 44100Hz or whatever output frequency is requested.

The number specified in the code does affect the internal calculations the XM player, but it doesn't change this end result. In fact, even changing this manually, there is no change in the calculated buffer size.

So there is no need for the command line option. I would just change this to 48000 just because it's a little less arbitrary than 44100Hz, and leave a comment that it doesn't affect the calculations anyway.

rasky:
OK I should have resolved all comments. I think it's ready to go!

examples/audioplayer/Makefile
rasky: Done!

PR #190, opened 2021-10-08
C++ support for n64.mk

Hazematman:
This pull request introduces c++ support to the n64.mk build system. I'm looking for feedback on how to handle linking c++ code, as linking c++ code with LD can be pain, so I opted to hijack the link command and replace it with CXX to use the C++ compiler for linking. I'm not sure if this is the best/cleanest approach, so please provide thoughts if you can.

### DIFF
--- a/examples/cpptest/Makefile
+++ b/examples/cpptest/Makefile
@@ -1,37 +1,21 @@
-PROG_NAME = cpptest
+BUILD_DIR=build
+SOURCE_DIR=.
+include $(N64_INST)/include/n64.mk
 
-ROOTDIR = $(N64_INST)
-GCCN64PREFIX = $(ROOTDIR)/bin/mips64-elf-
+cpptest.z64: MAPCMD=-Wl,-Map
+cpptest.z64: LD=$(N64_CXX)
+cpptest.z64: LDFLAGS=-L$(N64_ROOTDIR)/mips64-elf/lib -ldragon -lc -lm -ldragonsys -Wl,-Tn64.ld -Wl,--gc-sections
 
-CC = $(GCCN64PREFIX)gcc
-CXX = $(GCCN64PREFIX)g++
-AS = $(GCCN64PREFIX)as
-LD = $(GCCN64PREFIX)ld
-OBJCOPY = $(GCCN64PREFIX)objcopy
-N64TOOL = $(ROOTDIR)/bin/n64tool
-CHKSUM64 = $(ROOTDIR)/bin/chksum64
+all: cpptest.z64
 
-CFLAGS = -std=gnu99 -march=vr4300 -mtune=vr4300 -O2 -Wall -Werror -I$(ROOTDIR)/mips64-elf/include
-CXXFLAGS = -std=c++11 -march=vr4300 -mtune=vr4300 -O2 -Wall -Werror -I$(ROOTDIR)/mips64-elf/include
-LDFLAGS = -L$(ROOTDIR)/mips64-elf/lib -ldragon -lc -lm -lstdc++ -ldragonsys -Tn64.ld -Wl,--gc-sections
-N64TOOLFLAGS = -l 1M -h $(ROOTDIR)/mips64-elf/lib/header -t "C++ Test"
+$(BUILD_DIR)/cpptest.elf: \
+	$(BUILD_DIR)/cpptest.o
 
-ifeq ($(N64_BYTE_SWAP),true)
-$(PROG_NAME).v64: $(PROG_NAME).z64
-	dd conv=swab if=$^ of=$@
-endif
+cpptest.z64: N64_ROM_TITLE="cpptest"
 
-$(PROG_NAME).z64: $(PROG_NAME).bin
-	$(N64TOOL) $(N64TOOLFLAGS) -o $@ $^
-	$(CHKSUM64) $@
-
-$(PROG_NAME).bin: $(PROG_NAME).elf
-	$(OBJCOPY) $< $@ -O binary
-
-# Use g++ to link the compiled C++ object into an ELF executable
-$(PROG_NAME).elf: $(PROG_NAME).o
-	$(CXX) -o $@ $^ $(LDFLAGS)
-
-.PHONY: clean
 clean:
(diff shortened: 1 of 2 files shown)


n64.mk
```diff
@@ -99,10 +106,16 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c
...
+$(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.cpp 
+	@mkdir -p $(dir $@)
+	@echo "    [CXX] $<"
+	$(CXX) -c $(CXXFLAGS) -o $@ $<
+
 %.elf: $(N64_ROOTDIR)/mips64-elf/lib/libdragon.a $(N64_ROOTDIR)/mips64-elf/lib/libdragonsys.a $(N64_ROOTDIR)/mips64-elf/lib/n64.ld
 	@mkdir -p $(BUILD_DIR)
 	@echo "    [LD] $@"
-	$(LD) -o $@ $(filter-out $(N64_ROOTDIR)/mips64-elf/lib/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
+	@echo "    [LD] ${LD}"
```
meeq: This line seems like it was changed for testing purposes, but should probably be reverted

Hazematman: Good catch, yeah I had that in there to debug which LD program was being called

examples/cpptest/Makefile
```diff
@@ -1,37 +1,21 @@
-PROG_NAME = cpptest
+BUILD_DIR=build
+SOURCE_DIR=.
+include $(N64_INST)/include/n64.mk
 
-ROOTDIR = $(N64_INST)
-GCCN64PREFIX = $(ROOTDIR)/bin/mips64-elf-
+cpptest.z64: MAPCMD=-Wl,-Map
```
meeq, line 5: I don't love that this will need to be part of the Makefile pattern for any C++ project that uses n64.mk. It seems like a leaky implementation detail, but I don't have a better solution off the top of my head.

Hazematman: I agree, I'm not sure what the best way to handle this without requiring the make file to leak details. Perhaps we could have a define like CXX_PROJECT (or something like that) that is set in the Makefile, and in the LD build step of n64.mk if CXX_PROJECT we use CXX for linking instead of LD.

With this approach you wouldn't have as leaky of an implementation, handling on the n64.mk side might be a bit ugly though

n64.mk
```diff
@@ -99,10 +106,15 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c
...
+$(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.cpp 
+	@mkdir -p $(dir $@)
+	@echo "    [CXX] $<"
+	$(CXX) -c $(CXXFLAGS) -o $@ $<
+
 %.elf: $(N64_ROOTDIR)/mips64-elf/lib/libdragon.a $(N64_ROOTDIR)/mips64-elf/lib/libdragonsys.a $(N64_ROOTDIR)/mips64-elf/lib/n64.ld
 	@mkdir -p $(BUILD_DIR)
 	@echo "    [LD] $@"
-	$(LD) -o $@ $(filter-out $(N64_ROOTDIR)/mips64-elf/lib/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
+	$(LD) -o $@ $(filter-out $(N64_ROOTDIR)/mips64-elf/lib/n64.ld,$^) $(LDFLAGS) $(MAPCMD)=$(BUILD_DIR)/$(notdir $(basename $@)).map
```
anacierdem, line 117: We already accept `LDFLAGS` which we can introduce extra flags, we shouldn't need something extra here. What I mean is we can just do `LDFLAGS+=..` in the consumer side.

anacierdem: Let me try this sometime, probably Im missing something just looking at the PR

Hazematman: @anacierdem the issue here is that the argument of `-Map` is not inside the `LDFLAGS` variable and when you use `g++` as a linker you can't just pass `-Map` to g++. It expects the argument to be `-Wl,-Map` so that `g++` knows its not a `g++` argument and it can hand it down to the linker.

rasky: can't we just switch to always using g++ as linker? Does that create a problem for C programs?

rasky (changes requested):

n64.mk
Hazematman: That's a great question, I can test that out and see if all the example roms still work when using g++ as a linker as well as check if it adds an unwanted symbols into the binary.

rasky: I think the g++ driver should only add the C++ standard library to the default library list, and the linker should then remove all references to it (hopefully). 

This also begs the question whether using `-lstdc++` with ld is enough to fix C++ code.

anacierdem: I have tried wrapping all linker flags with `-Wl,` but that does not play well with `-lc` next step is trying `-lstdc++` as @rasky has suggested. If that does not also work for some reason I will try linking everything with `g++` providing `-lc` for non-C++ code.

anacierdem: Verified that `-lstdc++` is not enough even with `--start-group -ldragon -lstdc++ -lc  -lm -ldragonsys --end-group` but I think I have found an acceptable solution. Will try to push when I'm available.

rasky: FWIW I have been running with g++ as LD in the past days and I haven't noticed anything strange

anacierdem: I can try that as well if it is not known to cause any issues.

anacierdem: Now tested linking all examples with g++ and they all seem to work fine although theydo not produce the same exact binary. The only symbol that is extra in the g++ version (I did `mips64-elf-nm` for both files and compared only the symbol names) is `init_interrupts` e.g for example for the mixertest example. This is curious but does not seem to affect anything?

anacierdem:
Opened an alternative PR #197

anacierdem:
Resolved via #197

PR #195, opened 2021-10-10
Make runtime call hardware init functions automatically

Hazematman:
Modify the c runtime to automatically call initialize functions
for interrupts and for cop1.

Addtionally add cop1 init routine flash denormalized floats to 0.

As a note this pull request also modifies the testrom to mark `test_dma_read_misalign` as `TEST_FLAGS_NO_BENCHMARK`. This was done as when running on hardware I found that this test would randomly fall because of the execution time.


n64.ld
```diff
@@ -20,6 +20,8 @@
 OUTPUT_FORMAT ("elf32-bigmips", "elf32-bigmips", "elf32-littlemips")
 OUTPUT_ARCH (mips)
 EXTERN (_start)
+EXTERN(__init_interrupts)
+EXTERN(__init_cop1)
```
rasky: I think the idea here is that the compiler will auto-remove constructors of modules that are not used by the application. This is actually a good feature for us, since we want in general to reduce code size.

In this specific case, I would say that the interrupt constructor doesn't need extern: if I manage to do a main that never calls disable/enable interrupts (!), then I would not need init_interrupts as well. The interrupts would stay disabled and that's it.

On the other, I think the issue is that you put the constructor in cop1.c which is a new object file and otherwise unreferenced. If you moved that to something like n64sys.c, there wouldn't be a need for this.

What do you think?

src/cop1.c
```diff
@@ -0,0 +1,25 @@
...
+ * @{
+ */
+
+__attribute__((constructor)) void __init_cop1()
+{
+    /* Read initalized value from cop1 control register */
+    uint32_t fcr31 = C1_FCR31();
+    
+    /* Set FS bit to allow flashing of denormalized floats */
+    fcr31 |= C1_FS_BIT;
```
rasky: I think you should elaborate a little bit on this. I personally don't know what means "flashing a denomal", so I'd explain here (or a in a function-level comment) why this is required and how it is good.

tests/test_denormalized.c
```diff
@@ -0,0 +1,11 @@
+#include <float.h>
+
+void __initalize_cop1();
+
+void test_denormalized(TestContext *ctx) {
+    /* Create a volatile float, so gcc does not optimize it out */
+    volatile float x = 1.0f;
+    x /= FLT_MAX;
+
+    ASSERT(x == 0.0f, "Denormalized float was not flushed to zero");
```
rasky: Also a comment here to explain that we're testing that an exception is not generated when a denormal is created.

(Technically, a test that could fail with an exception should trap it to avoid trashing the whole testsuite when it fails, but that's something we don't have helpers for, so don't worry. I'm sure it would be a new rabbit hole...)

tests/test_denormalized.c
```diff
@@ -0,0 +1,11 @@
+#include <float.h>
+
+void __initalize_cop1();
```
rasky: Leftover?

tests/test_denormalized.c
```diff
@@ -0,0 +1,11 @@
+#include <float.h>
```
rasky: Maybe I'd call the file test_cop1 or test_float. Typically a file might contain multiple related tests, so the name should be more generic.

tests/testrom.c
```diff
@@ -202,12 +203,11 @@ static const struct Testsuite
 	TEST_FUNC(test_eepromfs,                   0, TEST_FLAGS_IO),
 	TEST_FUNC(test_cache_invalidate,    	1763, TEST_FLAGS_NONE),
 	TEST_FUNC(test_debug_sdfs,             	   0, TEST_FLAGS_NO_BENCHMARK),
-	TEST_FUNC(test_dma_read_misalign,       7003, TEST_FLAGS_NONE),
+	TEST_FUNC(test_dma_read_misalign,       7003, TEST_FLAGS_NO_BENCHMARK),
+    TEST_FUNC(test_denormalized,               0, TEST_FLAGS_NO_EMULATOR),
```
rasky: Tab vs space here

tests/testrom.c
```diff
@@ -202,12 +203,11 @@ static const struct Testsuite
 	TEST_FUNC(test_eepromfs,                   0, TEST_FLAGS_IO),
 	TEST_FUNC(test_cache_invalidate,    	1763, TEST_FLAGS_NONE),
 	TEST_FUNC(test_debug_sdfs,             	   0, TEST_FLAGS_NO_BENCHMARK),
-	TEST_FUNC(test_dma_read_misalign,       7003, TEST_FLAGS_NONE),
+	TEST_FUNC(test_dma_read_misalign,       7003, TEST_FLAGS_NO_BENCHMARK),
```
rasky: Can you please leave this with benchmark? It doesn't fail on many consoles so there must be something else that we should look into. If you change this, we will simply forget about it.

src/interrupt.c
```diff
@@ -600,6 +600,14 @@ void init_interrupts()
     }
 }
 
+/**
+ * @brief Initialize the interrupt controller
+ */
+void init_interrupts()
+{
+    __init_interrupts();
+}
```
rasky: __init_interrupts is idempotent, but since it's been already called by the constructor there's really no need to call it again, I would say. In fact, you could even change this to function to a static inline function in the header file that is empty and thus removed even if compiled.

rasky (commented):
Great PR! We're doing a great step here. We finally have constructors working and we simplify the libdragon initialization for beginners.

n64.ld
Hazematman: I like the idea of not having a new file for this one routine. I'll move it into n64sys.c, I think the EXTERN will still be required to make sure the cop1 bits are set, as I don't think every program is guaranteed to use a symbol from n64sys.c

src/interrupt.c
Hazematman: Because of -Werror, if we define a function and then never use it an error will generated. Perhaps we just make it an macro instead that evaluates to nothing?

n64.ld
rasky: I think n64sys is always included because it contains get_memory_size which is used by sbrk. But I see the point of this being a little fragile in general, so the EXTERN is probably worth it anyway.

src/interrupt.c
rasky: I don't see this problem in rsp.h for deprecated functions, but anyway you can use `__attribute__((used))` to disable that warning.

Hazematman: I think I was missing the `inline` part, that seems to have fixed it. I pushed a new commit that should address all of your comments.

include/cop1.h
```diff
@@ -34,6 +34,8 @@
 #define C1_CAUSE_INVALID_OP         0x00010000
 #define C1_CAUSE_NOT_IMPLEMENTED    0x00020000
 
+#define C1_FS_BIT                   (1<<24)
```
rasky: I think we should name this `C1_FCR31_FS`. I think we usually put the register name first for which the mask must be used.

src/n64sys.c
```diff
@@ -251,4 +251,27 @@ void wait_ms( unsigned long wait_ms )
...
+    /* Read initalized value from cop1 control register */
+    uint32_t fcr31 = C1_FCR31();
+    
+    /* Set FS bit to allow flashing of denormalized floats
+       The FPU inside the N64 CPU does not implement denormalized floats
+       and will generate an unmaskable exception if a denormalized float
+       is generated by any floating point operations. In order to prevent
+       this exception we set the FS bit in the fcr31 control register to
+       instead "flash" and "flush" the denormalized number. To understand 
+       the flashing rules please read pg. 213 of the VR4300 programmers manaul */
```
rasky: Typo "manual"

rasky (approved):
Thanks! This is good to go IMO

src/n64sys.c
```diff
@@ -251,4 +251,27 @@ void wait_ms( unsigned long wait_ms )
     wait_ticks(TICKS_FROM_MS(wait_ms));
 }
 
+
+/**
+ * @brief Initalize COP1 with default settings that prevent undesirable exceptions.
```
sp1187: Typo "Initialize"

Hazematman:
Thanks for all the feedback everyone :) I've pushed a new commit that resolves the above mentioned issues

src/n64sys.c
```diff
@@ -251,4 +251,27 @@ void wait_ms( unsigned long wait_ms )
...
 }
 
+
+/**
+ * @brief Initialize COP1 with default settings that prevent undesirable exceptions.
+ *
+ */
+__attribute__((constructor)) void __init_cop1()
+{
+    /* Read initalized value from cop1 control register */
```
sp1187: Typo "initialized".

Hazematman: Thanks! It seems that although english is my first language my proof reading skills are very bad :sweat_smile: I've pushed a new commit with that typo fixed

Hazematman: @sp1187 could you review the final patch, and sign off if you think its good?

sp1187 (approved):

PR #196, opened 2021-10-13
rsp: improvements and additions to rsp.inc

snacchus:
- vector store/load operations now take byte offsets
- added vector element macro
- added more defines for hardware registers
- fixed SP_STATUS_BROKE definition
- adapted rsp_mixer.S to the changes made

### DIFF
--- a/include/rsp.inc
+++ b/include/rsp.inc
@@ -264,294 +264,54 @@ makeNotImplemented tnei
     .endm
 .endm
 
-.macro makeLoadInstructionQuad name, opcode
+.macro makeLsInstruction mode, name, opcode, size, rangemin, rangemax
     .macro \name vt, element, offset, base
         .ifgt (\element >> 4)
             .error "Invalid element"
             .exitm
         .endif
 
-        .ifne ((\offset) % 16)
-            .error "Invalid offset - must be multiple of 16"
+        .ifne ((\offset) % \size)
+            .error "Invalid offset - must be multiple of \size"
             .exitm
         .endif
 
         .ifge (\offset)
-            .ifgt ((\offset) - 1008)
-                .error "Invalid offset - valid range: [-1024, 1008]"
+            .ifgt ((\offset) - \rangemax)
+                .error "Invalid offset - valid range: [\rangemin, \rangemax]"
                 .exitm
             .endif
 
-            loadVector \opcode, \vt, \element, ((\offset) / 16), \base
+            \mode\()Vector \opcode, \vt, \element, ((\offset) / \size), \base
         .else
-            .iflt ((\offset) + 1024)
-                .error "Invalid offset - valid range: [-1024, 1008]"
+            .iflt ((\offset) - \rangemin)
+                .error "Invalid offset - valid range: [\rangemin, \rangemax]"
                 .exitm
             .endif
 
-            loadVector \opcode, \vt, \element, (128 + ((\offset) / 16)), \base
+            \mode\()Vector \opcode, \vt, \element, (128 + ((\offset) / \size)), \base
         .endif
     .endm
 .endm
 
-.macro makeLoadInstructionDouble name, opcode
-    .macro \name vt, element, offset, base
-        .ifgt (\element >> 4)
-            .error "Invalid element"
-            .exitm
-        .endif
-
-        .ifne ((\offset) % 8)
-            .error "Invalid offset - must be multiple of 8"
-            .exitm
-        .endif
-
-        .ifge (\offset)
-            .ifgt ((\offset) - 504)
-                .error "Invalid offset - valid range: [-512, 504]"
-                .exitm
-            .endif
-
-            loadVector \opcode, \vt, \element, ((\offset) / 8), \base
-        .else
-            .iflt ((\offset) + 512)
-                .error "Invalid offset - valid range: [-512, 504]"
-                .exitm
-            .endif
-
-            loadVector \opcode, \vt, \element, (128 + ((\offset) / 8)), \base
-        .endif
-    .endm
+.macro makeLsInstructionQuad mode, name, opcode
+    makeLsInstruction \mode, \name, \opcode, 16, -1024, 1008
 .endm
 
-.macro makeLoadInstructionLong name, opcode
-    .macro \name vt, element, offset, base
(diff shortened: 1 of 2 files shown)


include/rsp.inc
```diff
@@ -263,44 +263,54 @@ makeNotImplemented tnei
...
-                    loadVector \opcode, \vt, \element, \offset, \base
-                .endif
+            .ifne ((\offset) % (1 << \offsetshift))
+                .error "Invalid offset - must be multiple of operand size"
             .else
-                .iflt (\offset + 64)
-                    .error "Invalid offset - valid range: [-64, 63]"
+                .ifge (\offset)
+                    .ifgt ((\offset) - (63 << \offsetshift))
+                        .error "Invalid offset - valid range: [-64, 63]"
```
rasky: This ranges are not the correct ones anymore, so the error message needs to be updated. Can you use variables in the `.error` lines? If not, I think we should resort to duplicate these macros for all different shift values. I know it's verbose but it's code for a greater good: we write lots of duplicated code here to provide a very good assembly support for microcode writers.

include/rsp.inc
```diff
@@ -263,44 +263,54 @@ makeNotImplemented tnei
...
             .error "Invalid element"
         .else
-            .ifge (\offset)
-                .ifgt (\offset - 63)
-                    .error "Invalid offset - valid range: [-64, 63]"
-                .else
-                    loadVector \opcode, \vt, \element, \offset, \base
-                .endif
+            .ifne ((\offset) % (1 << \offsetshift))
+                .error "Invalid offset - must be multiple of operand size"
```
rasky: This should possibly say the equivalent of "Invalid offset - must be a multiple of %d"

include/rsp.inc
```diff
@@ -456,6 +466,28 @@ makeStoreInstruction swv, 0b00111
...
+#define VE_4  12
+#define VE_5  13
+#define VE_6  14
+#define VE_7  15
+
+# Instead of using the constants above, it is possible to use the syntax 'e(x)'
+# via the following macros.
+#define _PPCAT2(n,x) n ## x
+#define _PPCAT(n,x) _PPCAT2(n,x)
+#define e(xx) _PPCAT(VE_, xx)
```
rasky, line 488: Nice! The only issue here is that, looking at rsp_mixer, I don't specifically like `e(v)` much. I don't think it's very clear, it looks more like line noise to me. Maybe we can keep a `,0` in those cases? What do you think?

src/audio/rsp_mixer.S
```diff
@@ -1050,25 +1050,25 @@ Mix32Start:
...
 	  # 1 cycle stall here
-	vaddc v_out_l, v_out_l, v_out_l,6;                 lqv v_sample_2,0, 2,s0
-	vaddc v_out_r, v_out_r, v_out_r,6;                 lqv v_sample_3,0, 3,s0
+	vaddc v_out_l, v_out_l, v_out_l,e(2h);             lqv v_sample_2,0, 0x20,s0
+	vaddc v_out_r, v_out_r, v_out_r,e(2h);             lqv v_sample_3,0, 0x30,s0
 	  # 1 cycle stall here
-	vaddc v_out_l, v_out_l, v_out_l,12;                bnez t0, Mix32Loop
-	vaddc v_out_r, v_out_r, v_out_r,12;                
+	vaddc v_out_l, v_out_l, v_out_l,e(4);                bnez t0, Mix32Loop
+	vaddc v_out_r, v_out_r, v_out_r,e(4);
```
rasky, line 1071: Now that I see like this, I think this sequence (1q / 2h / 4) is a little arbitrary. I think it can be changed to be more symmetrical. Anyway, I'll look into it as a followup, this comment was just to highlight that this macro makes the code much easier to read.

rasky (changes requested):
thanks! looks good

include/rsp.inc
snacchus: If we specify 0 as a default value for `element` (see https://sourceware.org/binutils/docs/as/Macro.html) it's possible to just omit the argument, quite like in the original syntax.

rasky: Oh awesome! Let's do that then!

include/rsp.inc
snacchus: I didn't find a way to adjust the error messages unfortunately, so I just ended up duplicating the macro.

include/rsp.inc
snacchus: This one would actually work without duplicating the macro, if we pass in the element size directly instead of the shift amount. TIL that GAS expands macro arguments in strings as well!

snacchus:
@rasky I made some adjustments based on your comments. I also found a way to avoid the endless `.if, .else, .endif` nesting in GAS macros. You can exit a macro early by using `.exitm` https://sourceware.org/binutils/docs/as/Exitm.html

include/rsp.inc
rasky: Out of curiosity, did you try with one level of macro generation more?

```
.macro makeMakeLoadInstruction makeMacroName, shift
   .macro \makeMacroName name, opcode
       .macro \name vt, element, offset, base
          .ifne ((\offset) % (1<<shift))
             .error "Invalid offset -- must be multiple of \shift"
             .endm
          .endif
       
          [...]
      .endm
   .endm
.endm

makeMakeLoadInstruction makeLoadInstructionShort, 2
makeMakeLoadInstruction makeLoadInstructionLong, 4
[...]
```

rasky: or maybe even if it works, it becomes too complicated to edit and maintain...

snacchus: I think I found a pretty good solution now (just pushed it). What do you think?

rasky: Awesome! Your macro-foo is impressive :)

snacchus:
The changes are good to go from my side. I also tested the mixer example ROM and it still works.

rasky (approved):
OK good to go.

@anacierdem: this PR requires first approval for CI run, and then it can be merged. The sooner the better as I'm writing ucode and I need to migrate to the new format asap.

anacierdem:
This is `frankengas™ v2.0` Thank you all!

anacierdem:
I have also opened #198

PR #197, opened 2021-10-14
Link everything with g++

anacierdem:
- Start linking with `g++` by default.
- Revert the equality on `__do_global_ctors`. Linking with `g++` seem to
add an initail item to the constructor list thus breaking initialization.
This started to be an issue as we have started using the constructos for
regular initialization as well. Now that we link with `g++` by default everythng is
compatible. The only problem is we cannot link with ld anymore if ever needed.
On the other hand `g++` is working as fast and can create equivalent output.

### DIFF
--- a/n64.mk
+++ b/n64.mk
@@ -13,7 +13,10 @@ N64_LIBDIR = $(N64_ROOTDIR)/mips64-elf/lib
 N64_GCCPREFIX = $(N64_BINDIR)/mips64-elf-
 N64_HEADERPATH = $(N64_LIBDIR)/header
 
+COMMA:=,
+
 N64_CC = $(N64_GCCPREFIX)gcc
+N64_CXX = $(N64_GCCPREFIX)g++
 N64_AS = $(N64_GCCPREFIX)as
 N64_LD = $(N64_GCCPREFIX)ld
 N64_OBJCOPY = $(N64_GCCPREFIX)objcopy
@@ -29,8 +32,11 @@ N64_AUDIOCONV = $(N64_BINDIR)/audioconv64
 N64_CFLAGS =  -std=gnu99 -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
 N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
 N64_CFLAGS += -DN64 -O2 -Wall -Werror -fdiagnostics-color=always
+N64_CXXFLAGS =  -std=c++11 -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
+N64_CXXFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
+N64_CXXFLAGS += -DN64 -O2 -Wall -Werror -fdiagnostics-color=always
 N64_ASFLAGS = -mtune=vr4300 -march=vr4300 -Wa,--fatal-warnings
-N64_LDFLAGS = -L$(N64_LIBDIR) -ldragon -lc -lm -ldragonsys -Tn64.ld --gc-sections
+N64_LDFLAGS = -L$(N64_LIBDIR) -ldragon -lm -ldragonsys -Tn64.ld --gc-sections
 
 N64_TOOLFLAGS = --header $(N64_HEADERPATH) --title $(N64_ROM_TITLE)
 N64_ED64ROMCONFIGFLAGS =  $(if $(N64_ROM_SAVETYPE),--savetype $(N64_ROM_SAVETYPE))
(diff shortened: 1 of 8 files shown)


n64.mk
```diff
@@ -29,8 +32,11 @@ N64_AUDIOCONV = $(N64_BINDIR)/audioconv64
 N64_CFLAGS =  -std=gnu99 -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
 N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
 N64_CFLAGS += -DN64 -O2 -Wall -Werror -fdiagnostics-color=always
+N64_CXXFLAGS =  -std=c++11 -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
+N64_CXXFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
+N64_CXXFLAGS += -DN64 -O2 -Wall -Werror -fdiagnostics-color=always
```
rasky: Can we maybe do `N64_CXXFLAGS = $(N64_CFLAGS) -std=c++11`? It should work as the latest `-std` wins.

If you do it later in the file once N64_CFLAGS is finished, even better

n64.mk
```diff
@@ -117,10 +127,15 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.c
...
+$(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.cpp
+	@mkdir -p $(dir $@)
+	@echo "    [CXX] $<"
+	$(CXX) -c $(CXXFLAGS) -o $@ $<
+
 %.elf: $(N64_LIBDIR)/libdragon.a $(N64_LIBDIR)/libdragonsys.a $(N64_LIBDIR)/n64.ld
 	@mkdir -p $(dir $@)
 	@echo "    [LD] $@"
-	$(LD) -o $@ $(filter-out $(N64_LIBDIR)/n64.ld,$^) $(LDFLAGS) -Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
+	$(CXX) -o $@ $(filter-out $(N64_LIBDIR)/n64.ld,$^) -lc $(patsubst %,-Wl$(COMMA)%,$(LDFLAGS)) -Wl,-Map=$(BUILD_DIR)/$(notdir $(basename $@)).map
```
rasky, line 140: I'd rather define `LD=g++`, with a comment explaining that we're currently using `g++` for linking as that is required to also support C++ code.

tests/testrom.c
```diff
@@ -205,6 +206,8 @@ static const struct Testsuite
 	TEST_FUNC(test_debug_sdfs,             	   0, TEST_FLAGS_NO_BENCHMARK),
 	TEST_FUNC(test_dma_read_misalign,       7003, TEST_FLAGS_NONE),
 	TEST_FUNC(test_cop1_denormalized_float,    0, TEST_FLAGS_NO_EMULATOR),
+	TEST_FUNC(test_constructors,               0, TEST_FLAGS_NO_EMULATOR),
```
rasky: why no emulator?

src/do_ctors.c
```diff
@@ -27,7 +27,9 @@ void __do_global_ctors()
 {
 	func_ptr * ctor_addr = &__CTOR_END__ - 1;
 	func_ptr * ctor_sentinel = &__CTOR_LIST__;
-	while (ctor_addr >= ctor_sentinel) {
+	// This will break if you link using LD. You'll need to change this to be >=
+	// in that case. See __CTOR_LIST__ in n64.ld
+	while (ctor_addr > ctor_sentinel) {
```
rasky, line 43: I think we should either suggest to modify the linker script, or this code, not both.

My suggestion would be to always modify the linker script, because the linker script is more "under the user control" compared to a runtime file of libdragon. So I would leave a comment here saying that ctor_sentinel is a sentinel value put by the linker (either 0xFFFFFFF or a count) which is not a valid function pointer.

Then in the linker script, I'd simply the comment simply saying that if for any reason one want to use `LD` instead of `g++` for linking, add a sentinel here (and give a link to the GCC page).

n64.mk
anacierdem: Hmm maybe I was not clear enough on this one, we still use ld for ucode stuff.

tests/testrom.c
anacierdem: Oh an oversight, thanks!

n64.mk
anacierdem: Actually I intentionally kept them separate in case we need something different in one or the other but I was not sure, I will merge them 👍🏼

anacierdem: Also g++ does not like it;
```cc1plus: error: command-line option '-std=gnu99' is valid for C/ObjC but not for C++ [-Werror]```

rasky (approved):

PR #201, opened 2021-10-21
Fix breaking change

### LINKED ISSUE #200: Breakage: C projects can no longer link libdragon using LD
I updated FlappyBird to fix breakages in the latest libdragon trunk, was able to build after #199 but when I run the resulting ROM I get:

image

This means that all existing projects using C and libdragon who have not updated to n64.mk will start producing broken builds. If debug is on, there will at least be the assertion error, but there is no indication at compile-time that anything is awry.

anacierdem:
This fixes #200

We are using linker's `--wrap` option to provide an alternative `__do_global_ctors` function when linking with g++. It does not actually wrap the original but just provides an alternative (new version).

When that options is not provided, the function from before #195 is used to introduce minimum breakage as possible to our knowledge.

### DIFF
--- a/src/do_ctors.c
+++ b/src/do_ctors.c
@@ -24,18 +24,43 @@ extern func_ptr __CTOR_END__ __attribute__((section (".data")));
  * @brief Execute global constructors
  * "Constructors are called in reverse order of the list"
  * @see https://gcc.gnu.org/onlinedocs/gccint/Initialization.html
+ *
+ * This version of the function is kept for compatibility for projects not using
+ * the build system but linking directly with ld in a legacy setup. For the
+ * modern version see #__wrap___do_global_ctors which is activated by the new
+ * build system (n64.mk) via --wrap linker flag. Do not use that flag if you are
+ * using ld so that this function is used instead.
  */
 void __do_global_ctors()
 {
-	// Somehow __CTOR_END__ points to the next location after the list and the
-	// last item is a zero, so start from the actual end.
+	func_ptr * ctor_addr = &__CTOR_END__ - 1;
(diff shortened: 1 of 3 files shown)


src/do_ctors.c
```diff
@@ -27,8 +27,21 @@ extern func_ptr __CTOR_END__ __attribute__((section (".data")));
...
+	while (ctor_addr > ctor_sentinel) {
+		if (*ctor_addr) (*ctor_addr)();
+		ctor_addr--;
+	}
+}
+
+/**
+ * @brief Execute global constructors
+ * This version is used by the new build system via the --wrap linker
+ * flag. When that is provided, this version will be utilized instead.
```
rasky: I'll add explicitly: the previous version is kept for compatibility for people not using the build system but linking directly with ld.

src/do_ctors.c
```diff
@@ -27,8 +27,21 @@ extern func_ptr __CTOR_END__ __attribute__((section (".data")));
  */
 void __do_global_ctors()
```
rasky, line 34: I'll mention here that this version is used when linking with LD

rasky (approved):
OK nice solution I would say. It would be great if @meeq / @sp1187 could test it and confirm it it works with their projects that were broken  by the latest change.

rasky:
Also please change the commit message to something more descriptive like "Fix compilation with LD that was broken by previous change"

meeq:
This fix appears to work -- I was able to link using LD and did not hit the assertion:

image

:+1:

anacierdem:
Oops mobile UI just merged it without asking for the message @rasky 🤷🏻‍♂️

Issue #159, opened 2021-07-29
Video streaking/distortions on real hardware


meeq:
I noticed this on my [Flappy Bird](https://github.com/meeq/FlappyBird-N64) project -- it looks fine on emulators, but on real hardware the video signal has big smears and streaks.

Flappy Bird uses the following arguments to `display_init`: 

```
RESOLUTION_320x240, DEPTH_16_BPP,
BUFFERING_DOUBLE, GAMMA_NONE, ANTIALIAS_RESAMPLE
```

Terrible photo attached:
![image](https://user-images.githubusercontent.com/11916110/127412457-f61f3b29-a356-4ddf-9147-0ecc554815c4.jpeg)

Terrible video link:
https://media.discordapp.net/attachments/627380625766088715/861113817676513301/VID_20210704_001720.mp4

fpgashibe on N64brew Discord seemed to think it was because the VI was not being set up correctly by the Display subsystem: https://discord.com/channels/205520502922543113/627380625766088715/861758855331840032

Marshall was also able to reproduce zebra striping artifacts by launching the LibDragon `ctest` example ROM after a reset from Animal Forest: https://discord.com/channels/205520502922543113/763114214688555039/870091995132071966

parasyte:
This is caused by the "dither filter" bit set in 16-bit mode: https://github.com/DragonMinded/libdragon/blob/fd45d487f48885883e341a8facd9e8276fa410cf/src/display.c#L322-L326

Changing the value to `0x2` fixes the vertical bars.

And if you enable antialiasing, the vertical bars will not be present with dither filter also enabled. Unfortunately, the antialias filtering is pretty bad.

werkn:
@parasyte I'm also getting the vertical bars like @meeq mentions in my apps, are we just setting `control |= 0x20002` or `control |= 0x2`?  Thanks for any help (I'm using the exact same setup for `display_init`, so `RESOLUTION_320x240, DEPTH_16_BPP,
BUFFERING_DOUBLE, GAMMA_NONE, ANTIALIAS_RESAMPLE`)

parasyte:
@werkn The latter. This will fix the vertical bars, if you are _not_ using antialiasing:

```c
control |= 0x2;
```

Alternatively, especially if you do not want to patch `libdragon`, you can just enable antialiasing by passing `ANTIALIAS_RESAMPLE_FETCH_ALWAYS` to `display_init()`. Bear in mind the quality caveat mentioned earlier.

[...]

PR #205, opened 2021-10-21
Update VI origin every frame according to current field

sp1187:
This small change improves the look of the interlaced video modes significantly, this is done by alternating the starting line of the VI origin (framebuffer) register according to the current field.

Commercial games also modify V_VIDEO and V_BURST (for PAL), I did not see any significant improvement when trying those modifications and it would require a bit more of restructuring than this PR, though it could be looked at further in the future.

@Rasky also noted that the VI interrupt currently seems to be triggered mid-frame, the initialized VI interrupt register value was therefore changed so that the interrupts would appear in vblank.

Also changed the display initialization to only enable the dither filter used for 16-bit color if anti-aliasing is enabled as the filter will otherwise cause ugly vertical streaks, see #159.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -428,7 +428,7 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
 
     /* Set which line to call back on in order to flip screens */
     register_VI_handler( __display_callback );
-    set_VI_interrupt( 1, 0x200 );
+    set_VI_interrupt( 1, 0x2 );
 }
 
 /**
--- a/src/display.c
+++ b/src/display.c
@@ -240,15 +240,21 @@ static void __write_dram_register( void const * const dram_val )
  */
 static void __display_callback()
 {
+    uint32_t *reg_base = (uint32_t *)REGISTER_BASE;
+
+    /* Least significant bit of the current line register indicates
+       if the currently displayed field is odd or even. */
+    bool field = reg_base[4] & 1;
+
     /* Only swap frames if we have a new frame to swap, otherwise just
        leave up the current frame */
     if(show_next >= 0 && show_next != now_drawing)
     {
-        __write_dram_register( __safe_buffer[show_next] );
-
         now_showing = show_next;
         show_next = -1;
     }
+
+    __write_dram_register(__safe_buffer[now_showing] + (!field ? __width * __bitdepth : 0));
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/display.c
```diff
@@ -240,15 +240,18 @@ static void __write_dram_register( void const * const dram_val )
  */
 static void __display_callback()
 {
+    uint32_t *reg_base = (uint32_t *)REGISTER_BASE;
```
rasky: I'd declare this volatile. Shouldn't matter now but maybe for future PRs might.

rasky (approved):
Thanks! Looks ok to me and it does improve picture stability on my CRT in interlaced modes

src/display.c
```diff
@@ -240,15 +230,18 @@ static void __write_dram_register( void const * const dram_val )
  */
 static void __display_callback()
 {
+    volatile uint32_t *reg_base = (uint32_t *)REGISTER_BASE;
+    bool field = reg_base[4] & 1;
```
anacierdem, line 237: Would be nice to know what does this register hold, it is apparent the lsb is the field but more context is appreciated.

src/display.c
```diff
@@ -320,9 +313,7 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
     switch( bit )
     {
         case DEPTH_16_BPP:
-            /* We enable divot filter in 16bpp mode to give our gradients
-               a slightly smoother look */
-            control |= 0x10002;
+            control |= 0x2;
```
anacierdem, line 319: What was this for?

sp1187: There are comments related to this added further below in this code. Basically the dither filter should not be enabled when anti-aliasing is disabled as it will cause ugly vertical streaks, see #159.

src/display.c
sp1187: This register is just VI_V_CURRENT which stores the current scanline (and can be used to get the current field), perhaps I could add a comment mentioning that but no other VI register access currently does so in display.c

anacierdem: It would be great if you could add more information! I am ok to merge this as it is but improvements are welcome.

rasky:
I suggest we use the rebase strategy with this PR because they are unrelated changes in different commits; so if we hit a problem it's much easier to bisect the actual issue

rasky (approved):
Fine by me with the clarification comment just pushed. This fixes lots of outstanding annoyances with display.c, thank you!

PR #206, opened 2021-10-23
Three followups to the XM player

rasky:
* Allow to enable / disable looping playback
 * Add API to register a callback for unused XM effects. This can be used to put custom music cues in the score and synchronize graphics and music.
 * Remove an assert on invalid channel frequency, given that we can't enforce it correctly yet (see commit message)


meeq:
I have tested this branch with FlappyBird and confirmed that it resolved the assertion issue.

Issue #161, opened 2021-08-03
Windows support?


rasky:
I've received a few PRs by @networkfusion (opened in my fork) to compile libdragon tools with MSVC:
https://github.com/rasky/libdragon/pulls
Before reviewing / merging them, I'd like to agree on a general plan of how Windows support in libdragon would look like, so that it's easier to drive contributions in the right direction. Moreover, I'm resisting using POSIX/GNU specific extensions in tools

I'm not really up to date with Windows development nowadays, so forgive me in advance. Focusing on tools only for now, my understanding is that there are three main avenues:

1) Use WSL (Windows Subsystem for Linux). This is (I simplify) a Linux VM within Windows. My understanding is that libdragon should just work as-is on WSL, no changes required.
2) Use MinGW/MSYS (with GNU make as build system). This would require very little modifications to the source code of tools (possibly none?) and it would still allow to use things like `getopt` or `asprintf`. It might require some changes to Makefiles (different extensions, etc.). It might even be possible to cross-compile the tools for Windows target from Linux CI.
3) Use MSVC. This is where I'm more confused about. I'm sure this would require some modification to the source code of tools, it would prevent using POSIX-ism / GNU-ism. It would also require changing the build system as MSVC doesn't ship with a GNU make-compatible tool, so we would have to switch to CMake or similar I guess. 

As for compiling the actual game/ROM, I would say that we should stick to a cross-compiled GCC / newline / binutils anyway, so on this regard the effort would probably be setting up some script to be able to cross-compile a Windows version of the MIPS GCC toolchain, and then package everything for distribution.

Any comment on the above from some Windows developer?

networkfusion:
Hey @rasky .

1. Yes WSL will work fine for build, but will not work with deploy tools due to no support for USB (in WSL V2 which is basically a docker container).
2. I have never (personally) liked the dependency on MSYS or CYGWIN for windows, it adds a big barrier to new users and "should" not be required...

3. It could use Makefiles with careful support for NMake (https://docs.microsoft.com/en-us/cpp/build/reference/nmake-reference?view=msvc-160) to keep the existing MakeFiles (happy). Although there will still be some issues with regards to specific gcc extensions which could easily be handled by an `ifdef` e.g. 
![image](https://user-images.githubusercontent.com/11439699/128037408-658570b0-3237-4024-8be4-0cf41202f275.png)
[...]

PR #207, opened 2021-10-23
CI: build tools with MinGW.

rasky:
In https://github.com/DragonMinded/libdragon/issues/161, we agreed that a native Windows toolchain of libdragon should not be compiled with MSVC but with MinGW (GCC), to avoid too much overhead on contributors.

This PR makes the CI compile the tools with MinGW and also fixes a few bugs in them. I can't guarantee they work 100% now, but I did some quick runs and they seem to work at least in their basic usage.

Extracted from https://github.com/rasky/libdragon/pull/6, submitted by @networkfusion (thanks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
       - completed
 
 jobs:
-  ci:
+  Compile-Library-And-Examples:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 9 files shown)


rasky:
I found that audioconv64 didn't correctly convert XM64 files on 32-bit architectures (Windows 32bit is the first we're building tools on). I am not even sure we need Windows 32-bit builds. Anyway, the problem is related to an existing issue with the conversion code where we need to estimate the sizeof of some structures at runtime on N64, and that's not easy to do portably and properly. Since it's just an estimation that needs to be conservative, I just bumped it on Win32 and that's fine.

Also, I took the change in a separate commit to improve a bit error reporting on XM64 loading failures.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,43 @@
+name: Windows toolchain
+
+on: [push, pull_request]
+
+jobs:
+  Build-Tools:
+    runs-on: windows-latest
+    strategy:
+      fail-fast: false
```
anacierdem: Why disable this?

anacierdem (commented):
Looks good to me. I'm ok to have a single action as a build and verification at the same time

.github/workflows/windows.yml
rasky: I think that even the 64-bit build fails, it's still useful to see if the 32-bit one succeeds and viceversa, isn't it?

anacierdem: Thats a valid perspective...

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: Build
+name: Testing
```
anacierdem: This will change the badge text as well IINM. Does it also makes sense there?

anacierdem (commented):
Should we also include the win tests in the main ci / or create a new badge?

.github/workflows/ci.yml
rasky: Changed back to Build then.

rasky:
I've changed the PR so that the Windows build is now part of ci.yml and so part of the same badge. It is a build of course -- it's just that we only build tools for now.

PR #208, opened 2021-10-23
Add player for the YM module format

rasky:
This commit introduces a new music player, which is based on the .YM
module format, a format first popularized in the Atari ST emulator scene.

The format is based around the very popular AY-3-8910 sound chip, that
was powering a few successful 8-bit consoles such as Atari ST, ZX Spectrum
and Amstrad CPC. It is a 3-channel PSG with envelope support. It can produce
typical 8-bit "chiptune" music scores. Nowadays, it is possible to compose
soundtracks using the Arkos Tracker 2 tool, that exports in the YM format.

The YM format is a simple dump of the state of all registers of the AY
chip at a fixed time step. To playback, it is necessary to emulate the
AY PSG. The implementation has been carefully optimized for the N64 MIPS
CPU for high-performance, so that playback typically takes less than 5%
of CPU time, plus a few percents of RSP time for resampling and mixing
(done by the mixer).

The YM64 format is actually compatible with YM, it's just one of the many
partially-documented variations of YM to which we gave it a distinguished name.
audioconv64 can be used to convert most YM files to YM64.

In general, most homebrew productions will prefer the XM format for general
music tracking as it is capable of much higher quality and with many more
channels, but YM can attract some oldskool composers. Anyway, it doesn't
take ROM space if it's not used and it's an example of how to use the mixer
to playback generated digital samples.

This is the audio player example ROM as built with this PR, with YM files
provided, courtesy of MishtaLu.
[audioplayer.zip](https://github.com/DragonMinded/libdragon/files/7402588/audioplayer.zip)

*This PR is dedicated to the late Sir Clive Sinclair whose computer,
powered by the AY-3-8910, helped popularize what we now call
chiptune music.*


meeq:
You should add your dedication as a comment into `src/audio/ay8910.c` ❤️

src/audio/ym64.c
```diff
@@ -0,0 +1,266 @@
+#include "ym64.h"
+#include "ay8910.h"
+#include <assert.h>
+#include <string.h>
+#include <stdio.h>
+#include <libdragon.h>
+#include "lzh5.h"
+
+#define MIN(a,b) ((a)<(b)?(a):(b))
```
meeq: This macro is sub-optimal because it double-evaluates the arguments.

[From a Stack Overflow answer for "MIN and MAX in C":](https://stackoverflow.com/a/3437484)

> Everyone says "oh I know about double evaluation, it's no problem" and a few months down the road, you'll be debugging the silliest problems for hours on end.
> [...]
>You can fix the macro using [GCC's typeof](https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Typeof.html#Typeof) (you get typesafety bonus too) in a GCC statement expression:

```c
 #define MIN(a,b) \
   ({ __typeof__ (a) _a = (a); \
      __typeof__ (b) _b = (b); \
     _a < _b ? _a : _b; })
```

It may also be worth centralizing these generic utility macros into an internal include file.

tools/audioconv64/audioconv64.c
```diff
@@ -185,7 +194,7 @@ int main(int argc, char *argv[]) {
 				if (!strcmp(argv[i], "true") || !strcmp(argv[i], "1"))
 					flag_wav_looping = true;
 				else if (!strcmp(argv[i], "false") || !strcmp(argv[i], "0"))
-					flag_wav_looping = true;
+					flag_wav_looping = false;
```
meeq: Seems like this should be included in https://github.com/DragonMinded/libdragon/pull/206

tools/audioconv64/lzh5_compress.c
```diff
@@ -0,0 +1,1273 @@
+// Extracted from https://github.com/jca02266/lha
+// and simplified for our use case.
+// TODO(rasky): still not a proper library, it will call fatal_error + exit(1)
```
meeq, line 3: Is this even worth fixing?

src/audio/ym64.c
rasky: Right, I think I've got the correct implementation in mixer.c in fact. I guess it's time to add an internal include for these utilities. I'll do something like `cutils.h`.

tools/audioconv64/audioconv64.c
rasky: Right, it's out of place. It's actually an unrelated bug, I will separate in a different PR.

tools/audioconv64/lzh5_compress.c
rasky: Maybe not right now, no. I doubt anybody will hit any of those fatal errors in the context of audioconv64.

tools/audioconv64/audioconv64.c
rasky: #210

tools/audioconv64/conv_ym64.c
```diff
@@ -0,0 +1,292 @@
...
+}
+
+static void ymwrite(const void *buf, int sz) {
+    fwrite(buf, 1, sz, ym_f);
+}
+
+// Compress a file with LHA (algorithm -lh5-). This is done using
+// a stripped down version of https://github.com/jca02266/lha
+// stored as single file in lzh5_compress.c. The library works only
+// through FILE*, so the compression will happen from disk to disk.
```
anacierdem, line 70: Would `open_memstream` help?

tools/audioconv64/conv_ym64.c
```diff
@@ -0,0 +1,292 @@
...
+        csum += h[i+2];
+    for (int i=0;i<strlen(lha_fn);i++)
+        csum += (uint8_t)lha_fn[i];
+    csum += crc16 & 0xFF;
+    csum += crc16 >> 8;
+    head.checksum = csum;
+
+    // Write again the header and the crc16 of the file.
+    fseek(out, 0, SEEK_SET);
+    fwrite(&head, 1, sizeof(head), out);
```
anacierdem, line 119: Will this cause any issues with win32?

tools/audioconv64/conv_ym64.c
```diff
@@ -0,0 +1,292 @@
...
+            int f = i % nframes;
+            outdata[f*16+r] = data[i];
+        }
+
+        // Write a YM5 format (uncompressed) into temporary file.
+        const char *tmpfilename = ".song.tmp";
+        ym_f = fopen(tmpfilename, "wb");
+        if (!ym_f) fatal("cannot create: %s", tmpfilename);
+
+        ymwrite("YM5!LeOnArD!", 12);
```
anacierdem, line 195: Is this the actual header :)

src/audio/ym64.c
```diff
@@ -0,0 +1,265 @@
...
+
+static unsigned int lha_callback(void *buf, size_t buf_len, void *user_data) {
+	int32_t fh = (int32_t)user_data;
+	return dfs_read(buf, 1, buf_len, fh);
+}
+
+static void ym_wave_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wlen, bool seeking) {
+	ym64player_t *player = (ym64player_t*)ctx;
+
+	// Computer the number of samples per audioframe. Keep it as floating point
```
anacierdem: typo: computes

tools/audioconv64/conv_ym64.c
```diff
@@ -0,0 +1,292 @@
...
+
+static FILE *ym_f;
+static bool ym_compressed;
+static LHANewDecoder ym_decoder;
+
+static size_t lha_callback(void *buf, size_t buf_len, void *user_data) {
+    return fread(buf, 1, buf_len, ym_f);
+}
+
+static void ymread(void *buf, int sz) {
```
anacierdem, line 55: Wouldn't be better if we get the file handle from the arguments and pass it explicitly?

src/audio/ym64.c
```diff
@@ -0,0 +1,265 @@
...
+		assertf(0, "YM3 format cannot be played -- convert with audioconv64");
+	} else {
+		assertf(0, "invalid YM header: %s", head);
+	}
+
+	// Record the file offset at the beginning of audio frames. This will
+	// be useful for looping.
+	player->start_off = offset;
+
+	// Compute playback frequency. Use floating point for accurate representation
```
anacierdem, line 182: Isn't sample rate a better terminology here?

src/audio/ym64.c
```diff
@@ -0,0 +1,265 @@
...
+	// of what is requested by the module definition. The mixer supports fractional
+	// frequency so we don't want to waste precision.
+	float freq = (float)player->extfreq / 8 / AY8910_DECIMATE;
+
+	// Compute the waveform length and loop start position in sample. Again, use
+	// floating point here to try to be as accurate as possible.
+	int len = (float)player->nvbl * freq / player->playfreq;
```
anacierdem: So each frame has `player->playfreq` sample IINM?

Makefile
```diff
@@ -10,8 +10,8 @@ INSTALLDIR = $(N64_INST)
 libdragon: CC=$(N64_CC)
 libdragon: AS=$(N64_AS)
 libdragon: LD=$(N64_LD)
-libdragon: CFLAGS+=$(N64_CFLAGS) -I$(CURDIR)/include
-libdragon: ASFLAGS+=$(N64_ASFLAGS) -I$(CURDIR)/include
+libdragon: CFLAGS+=$(N64_CFLAGS) -I$(CURDIR)/src -I$(CURDIR)/include
```
anacierdem, line 13: Why we needed it?

include/ay8910.h
```diff
@@ -0,0 +1,125 @@
...
+ */
+#define AY8910_VOLUME_ATTENUATE   0.8
+
+/** @brief Generate silence as 0.
+ * 
+ * Normally, AY8910 generates output samples in which the silence is
+ * represented by -32768 (minimum volume). This is a little inconvenient
+ * if the caller wants to skip generation when the AY8910 is muted for
+ * performance reasons, because audio mixers normally assume that muted
+ * channels are made by samples with value 0, otherwise disabling the
```
anacierdem, line 45: Can't we convert the samples to match instead?

include/ay8910.h
```diff
@@ -0,0 +1,125 @@
...
+ * AY8910 would affect the volume of all other channels.
```
anacierdem, line 46: How so?

include/ay8910.h
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * By setting this macro to 1, the dynamic range will be halved in the range
+ * 0-32767, so the silence will be as expected, but the audio will
+ * be somehow "duller".
```
anacierdem, line 50: Doesn't this will also deteriorate SNR due to more aliasing?

anacierdem (commented):
Looks great as always!
I have a few questions + concerns on win32 compatbility, which is no very important anyways.
And the center volume solution didn't look like a good one although I don't have an exact alternative as I don't know YM.

tools/audioconv64/conv_ym64.c
rasky: Oh TIL. I'll check.

tools/audioconv64/conv_ym64.c
rasky: yes :)

src/audio/ym64.c
rasky: done

tools/audioconv64/conv_ym64.c
rasky: You would have to pass file handle, ym_decoder and compressed flag to every call. I know this is old style C programming with global variables, but C doesn't give us currying tools to help with this. I would have to create a structure with the state like YMReader and use that. I can do that if you think it's worth it.

tools/audioconv64/conv_ym64.c
rasky: those headers have static asserts on their size, and it passes the MinGW CI builds, so I don't foresee any problems.

include/ay8910.h
rasky: What you mean by converting the samples?

include/ay8910.h
rasky: Any fixed value produces a non audible waveform, so if you play a WAV whose samples are all 123, you hear nothing. 

In general a mixer assumes that you can mix samples of different channels by adding them. The identity of an addition is 0, so if you have 32 channels and are playing only on 1, the other 31 should produce "0" samples as mute, so that their presence doesn't affect the single channel that is playing. If other channels produce any other fixed value, you still don't hear anything else, but the *volume* of the playing channel is affected. This is how I understand it at least (even though volume is multiplicative, so it's not totally clear to me).

Anyway, if you turn on/off a channel that contains a fixed value, you do hear a change in the playing channels, unless that channel is 0. Unfortunately for us AY-3-8910 produces a 4-bit digital output were 0 is silence is 15 is maximum amplitude, so it's an unsigned value. If we convert that to signed, we get that silence is -32768, which is what we want to avoid in our scenario.

tools/audioconv64/conv_ym64.c
anacierdem: Probably not worth it but took me some while to understand what's going on :)

include/ay8910.h
rasky: Yeah it does have an effect on the audio. I said "duller", and is possibly deteriorating SNR though I'm not sure about it. That's unfortunate but I think it's the best compromise to mix the AY-3-8910 digital output to our mixer and allow other things to be played at the same time.

Actually I must say that the sound isn't too sharp anyway. If you use this very emulator of the chip, define AY8910_DECIMATE to 1, get the output samples and play them at their native rate with a PC player like sox (that will do a decent downsampling filter), the audio quality is **leaps and bounds** better. I think the corruption comes from either the downsampling made by window average (which is a very basic low-pass filter) or the subsequent upsampling without any interpolation (which is OK for low-quality waveforms, but possibly not OK for a signal that was already downsample a lot). This would be an area of investigation if we wanted to improve the audio quality of the YM player

include/ay8910.h
anacierdem: Scaling/transforming them such that we have 0 where we need it

include/ay8910.h
anacierdem: Actually having a constant value on one of the channels should not affect the volume. It would introduce a DC bias though, which would decrease the available headroom. i.e it would be easier to clip. How much it matters relates to the scaling of the output. Another thing to keep in mind, a channel with a non zero value can create a pop initially though the audibility would depend on the IR of the output stage. If we hear a change when there is a DC bias there maybe something wrong somewhere.

anacierdem: Also I'm not sure I follow that conversion how does a 4bit value of 0 becomes -2^15?

rasky: > having a constant value on one of the channels should not affect the volume

I can check that again. This code predates the mixer BTW.

> It would introduce a DC bias though, which would decrease the available headroom. i.e it would be easier to clip.

That for sure is a problem in itself. 

> How much it matters relates to the scaling of the output.

I've added a macro to lower the volume of the AY-3-8910 specifically for this. It's not that useful when you activate the center volume feature, in fact.

> Another thing to keep in mind, a channel with a non zero value can create a pop initially

Yes that happened pre-mixer, and I also had a comment here about that. It doesn't happen now because the mixer does a volume ramp in RSP specifically to avoid this (both when the volume suddenly changes during playback, and for start/stop).

> Also I'm not sure I follow that conversion how does a 4bit value of 0 becomes -2^15?

How do you convert an unsigned 4 bit amplitude into a 16 bit signed amplitude? The normal formula would be to map 0 to -2^15, and 15 to 2^15, with the other values in-between.

src/audio/ym64.c
rasky: Maybe? I'm not sure. In general, I've been using frequency as in "frequency of playback" almost everywhere.

include/ay8910.h
rasky: I think that's what `AY8910_CENTER_SILENCE` does?

Makefile
rasky: it's explained in that commit's message: I'd like to start putting private header files in src/ as discussed elsewhere. `utils.h` is the first one

src/audio/ym64.c
rasky: No, `playfreq` is the number of frames per second. BTW I've renamed `nvbl` into `nframes` and `extfreq` into `chipfreq` as I think they're more clear. The mental model is that you're emulating a console so there is a frame frequency (50Hz or 60Hz), a chip frequency (clock of the AY chip) and a total number of frames in the song.

tools/audioconv64/conv_ym64.c
rasky: It looks like it won't work. I have two codepaths: 1) no compression: write to final file directly. 2) compression: write to temporary file; then compress from temporary to final file. `open_memstream` doesn't allow to read from it, only write (duh!), so It can't replace the temporary file.

include/ay8910.h
anacierdem: What I thought initially was mapping 0-15 to 0-2^15 and I think this is what the define already does. From the descrption I thought it was causing a drop on representable sample resolution but thats not true as I now better understand it. I wonder where the original chip's output offset is. Then I suspect the "dullness" is coming from the fact that we are using a sort of "rectified" output which never crosses the outputs center point. Then the ideal solution would be adding a DC removal filter but that might be an overkill.

anacierdem:
I'm OK with this we may experiment with the audio quality later. You can merge whenever you are ready unless you change smth and want another review.

PR #209, opened 2021-10-23
Reset graphics FG/BG colors for assert and exception screens

sp1187:
Currently the console printed on crash screen (asserts or exceptions) will reuse whatever foreground/background graphics color settings the game used before the crash. This can cause readability issues if those colors happen to too similar. This PR resets the colors when entering those screens so that it will always be white text on black.

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -489,6 +489,7 @@ void debug_assert_func_f(const char *file, int line, const char *func, const cha
 	console_init();
 	console_set_debug(true);
 	console_set_render_mode(RENDER_MANUAL);
+	graphics_set_color(0xFFFFFFFF, 0x00000000);
 
 	fprintf(stdout,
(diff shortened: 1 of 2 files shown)


rasky:
I'm +1 on the idea but why not in console_init()? Console resets display altogether and set its own mode. It seems an oversight that it can inherit a previous settings of the graphics module.

sp1187:
I decided to be conservative here and keep the current semantics of console_init, in case there are programs relying on colors being kept by setting their custom colors before their console_init call. If others still think it would make more sense to modify console_init, that could of course be done instead.

anacierdem:
I dont think anyone would expect it an init command to keep the color set. If it was for draw text that might made sense but this is console "mode" all in itself. But who knows, maybe someone depends on that. Still moving to init seems a little better to me as well.

rasky:
Closing in favor of #213

PR #213, opened 2021-10-26
Reset graphics color in console_init

sp1187:
This PR resets the graphics foreground and background color when initializing any console, as an alternative approach to the issue described in #209.


rasky (approved):
Thanks!

PR #214, opened 2021-10-27
display: avoid configuring ANTIALIAS_OFF when it triggers a hw bug

### LINKED ISSUE #66: Disabling AA when setting up a video mode glitches up the display
Hello,
i had an issue with my homebrew games on the N64. (They were using libdragon for rendering of course)
When using something like :
``` display_init( RESOLUTION_320x240, DEPTH_16_BPP, 2, GAMMA_NONE, ANTIALIAS_OFF );```

It would glitch up the display and pretty much make things unplayable
vlcsnap-2019-08-21-04h45m23s128

This is how it's supposed to look like
vlcsnap-2019-08-21-04h46m35s986

If you replace ANTIALIAS_OFF with ANTIALIAS_RESAMPLE, then it works again. (but you have AA all over the screen, bummer)
Oddly enough, if you try to set up another video mode after that, it can glitch up like disabling AA.

I have no idea what causes this, i could get no AA mode to work just fine with n64chain for example.

rasky:
On NTSC consoles, disabling antialias/resample on low resolutions
triggers a hardware bug. The bug does not appear to be present on PAL
consoles though.

We assume that most users will want to build an application that works
on all consoles, so we assert with an explicit error message in case
ANTIALIAS_OFF is selected, suggesting to use ANTIALIAS_RESAMPLE which
should be acceptable in most use cases and works on all consoles.

Fixes #66

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -339,6 +339,21 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
     switch( aa )
     {
         case ANTIALIAS_OFF:
+            /* Disabling antialias hits a hardware bug on NTSC consoles on
+               low resolutions (see issue #66). We do not know the exact
+               horizontal scale minimum, but among libdragon's supported
+               resolutions the bug appears on 256x240x16 and 320x240x16. It would
+               work on PAL consoles, but we think users are better served by
+               prohibiting it altogether. 
+
+               For people that absolutely need this on PAL consoles, it can
+               be enabled with *(volatile uint32_t*)0xA4400000 |= 0x300 just
+               after the display_init call. */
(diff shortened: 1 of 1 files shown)


Ryzee119:
This change looks good and would have certainly prevented my issue and aid people developing on PAL consoles who are blind to this issue until tested on NTSC.

Is it worth capturing https://github.com/DragonMinded/libdragon/issues/159 somehow too? (In this PR or future PR)

rasky:
#159 was already fixed in a general way by https://github.com/DragonMinded/libdragon/commit/8676c7b284dd051ffd1b8c1ca4e25adef0bf9e2b

Ryzee119:
@rasky ah perfect. All good

anacierdem (approved):

PR #216, opened 2021-10-29
Shaded and flat triangles based on snacchus's guide

space222:
works on hardware (ntsc console). The values in rdp_enable_shading might not be perfect though (copied from a krom test and generated by online tool respectively).


src/rdp.c
```diff
@@ -802,6 +823,169 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+ * @param[in] v3B
+ *            Blue value for vertex 3
+ */
+void rdp_draw_shaded_triangle(float x1, float y1, float x2, float y2, float x3, float y3, float v1R, float v1G, float v1B, 
+                float v2R, float v2G, float v2B, float v3R, float v3G, float v3B)
+{
+    const float to_fixed_11_2 = 4.0f;
+    const float to_fixed_16_16 = 65536.0f;
+
+    if( y1 > y2 ) { SWAP(y1, y2) SWAP(x1, x2) SWAP(v1R, v2R) SWAP(v1G, v2G) SWAP(v1B, v2B) }
```
sp1187: I think these macro invocations looking like function calls without semicolons are a bit ugly.
You should probably change the SWAP macro to `#define SWAP(a, b) do { float t = a; a = b; b = t; } while (0)`, this would make the macro behave a bit more like normal statements.

src/rdp.c
```diff
@@ -423,6 +427,20 @@ void rdp_enable_blend_fill( void )
...
+ *
+ * This must be called before using #rdp_draw_shaded_triangle.
+ */
+void rdp_enable_shading( void )
+{
+    __rdp_ringbuffer_queue(0x2f002030);
+    __rdp_ringbuffer_queue(0);
+    __rdp_ringbuffer_queue(0x3C0F8E1F);
+    __rdp_ringbuffer_queue(0x001E0904);
+    __rdp_ringbuffer_send();
```
snacchus: This is `Set Other Modes` plus `Set Combine Mode` right? I know a lot of code in this file doesn't do this either, but it would be nice to add some documentation on what commands are being queued here with which parameters and what they are supposed to do.

Also I'm not sure what effect setting the `Other Modes` will have in the context of `rdp.c`. Do any other features depend on specific flags to be set? @rasky @anacierdem @Hazematman I'd like to hear your opinion on this as well.

src/rdp.c
```diff
@@ -802,6 +823,169 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+    
+    __rdp_ringbuffer_queue( (DrDe_fixed<<16) | (DgDe_fixed&0xffff) );
+    __rdp_ringbuffer_queue( (DbDe_fixed<<16) );
+    
+    __rdp_ringbuffer_queue( (DrDy_fixed<<16) | (DgDy_fixed&&0xffff) );
+    __rdp_ringbuffer_queue( (DbDy_fixed<<16) );
+    
+    __rdp_ringbuffer_send();
+    
+    return;
```
snacchus: `return` is not needed at the end of a `void` function.

src/rdp.c
```diff
@@ -826,43 +1010,56 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+    
+    __rdp_ringbuffer_queue((int)( xl * to_fixed_16_16 ) );
+    __rdp_ringbuffer_queue((int)( isl * to_fixed_16_16 ) );
+    __rdp_ringbuffer_queue((int)( xh * to_fixed_16_16 ) );
+    __rdp_ringbuffer_queue((int)( ish * to_fixed_16_16 ) );
+    __rdp_ringbuffer_queue((int)( xm * to_fixed_16_16 ) );
+    __rdp_ringbuffer_queue((int)( ism * to_fixed_16_16 ) );
     __rdp_ringbuffer_send();
+    
+    return;
```
snacchus: `return` is not needed at the end of a `void` function.

src/rdp.c
space222: Done, thank you.

src/rdp.c
```diff
@@ -802,6 +823,169 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+    const float Hx = x3 - x1;        
+    const float Hy = y3 - y1;        
+    const float Mx = x2 - x1;
+    const float My = y2 - y1;
+    const float Lx = x3 - x2;
+    const float Ly = y3 - y2;
+    const float nz = (Hx*My) - (Hy*Mx);
+    uint32_t lft = nz < 0;
+    
+    y1f = (y1f&0x1fff) | ((y1f&0x80000000)>>18);
```
snacchus: Maybe define this as a macro? Something like `#define TRUNCATE_S11_2(x) (((x)&0x1fff) | (((x)&0x80000000)>>18))`

src/rdp.c
```diff
@@ -802,6 +823,169 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+    if( y2 > y3 ) { SWAP(y2, y3); SWAP(x2, x3); SWAP(v2R, v3R); SWAP(v2G, v3G); SWAP(v2B, v3B); }
+    if( y1 > y2 ) { SWAP(y1, y2); SWAP(x1, x2); SWAP(v1R, v2R); SWAP(v1G, v2G); SWAP(v1B, v2B); }
+    
+    int y1f = y1*to_fixed_11_2;
+    int y2f = y2*to_fixed_11_2;
+    int y3f = y3*to_fixed_11_2;
+    
+    y1 = y1f/4.0f;
+    y2 = y2f/4.0f;
+    y3 = y3f/4.0f;
```
snacchus: Can you explain why this is necessary? Why get rid of the additional precision here?

src/rdp.c
```diff
@@ -826,43 +1010,56 @@ void rdp_draw_filled_rectangle( int tx, int ty, int bx, int by )
...
+    if( y2 > y3 ) { SWAP(y2, y3); SWAP(x2, x3); }
+    if( y1 > y2 ) { SWAP(y1, y2); SWAP(x1, x2); }
+
+    int y1f = y1*to_fixed_11_2;
+    int y2f = y2*to_fixed_11_2;
+    int y3f = y3*to_fixed_11_2;
+    
+    y1 = y1f/4.0f;
+    y2 = y2f/4.0f;
+    y3 = y3f/4.0f;
```
snacchus: See my other comment.

snacchus (changes requested):
Thank you for the submission and for putting my guide into code! I have a couple of questions/suggestions, but the implementation looks good so far. Also @anacierdem recently implemented triangle rendering for their jam entry and achieved pretty good results, so they could probably give some good feedback on the implementation details as well.

Another thing that isn't strictly necessary but would be very appreciated is if you could add an example project to the `examples/` folder that tests/demonstrates the features you have implemented. I think `mixertest` and `audioplayer` are the most recently added examples and should serve as a good reference regarding project structure and setup (some others don't use the new build system for example).

src/rdp.c
space222: Based on searching the discord the cause of gaps between triangles (which I was getting, and this appeared to solve) is calculating the slopes using full Y precision, but RDP only using 2bits for Y fraction. I used an implicit cast rather than a rounding function from a combination of laziness and not knowing what kind of rounding would be most correct or desired.

src/rdp.c
space222: Had a college instructor back in the day that was very insistent on it being good style and been doing it ever since. Should I take it that no explicit return from void is recommended style for libdragon? (I tried to stick to existing style, just didn't notice that element)

src/rdp.c
space222: For Other Modes, it'll take some looking up. And the others likely know better than I for what's most correct too. For Combine Mode I searched the discord and found I had the same striped triangle result as another and that the value in krom's test did not work on hardware (cen64 does not appear to care). What I ended up with  was simply an attempt to get the output to be just the shade color out of the (A-B)*C+D. IIRC, I tried to get everything else to be zero and put SHADE in D and it worked on hardware so I left it alone after that.

I do not know. For all I know, I've disabled texturing in the Other Modes command or something (which would naturally affect textured rects).
What is probably needed is all the constants/defines for the flags/options in both these commands and an actual function/system to set them, as well as deciding whether it should always be done by the callee or by the actual function checking a cached version and doing what's necessary to change to correct state for their operation. This would be even more necessary as the different triangle and color and blending options permutate (looks like there's a lot!).

space222: Also, thank you for the guide, it's really good!

src/rdp.c
snacchus: Personally I think the additional `return` doesn't really add anything, so I'm not sure what makes it good style. It is a matter of preference though. But yes, I think we should stick with a consistent style in libdragon, which is to omit the `return`.

src/rdp.c
snacchus: I just checked again and the other functions (`rdp_enable_primitive_fill`, `rdp_enable_texture_copy` etc.) all set the Other Modes as well. But instead of `0x2F` they use `0xEF` (I don't know why, I think the former is more correct) which is quite confusing. So if each of these functions represents a state which supports a certain mode of drawing (like "sprites" or "shaded triangles" etc.) it should be fine to set the Other Modes each time. Also AFAIK the Combine Mode has no effect in copy or fill mode, so that should be ok to set here as well.

As long as these values do the job they should be fine. But as I said, to avoid future confusion some documentation is needed. A simple comment should be enough for now.

A new system to generate and queue raw RDP commands is actually something that is planned. We can then just replace these manually assembled values with the new API when it's ready.

Hazematman: I think this setup of manually setting the other modes flag will get unwieldy as `rdp.c` grows. Right now functions like `rdp_enable_blend_fill` and `rdp_enable_primitive_fill` will also override the "set other modes" state. It would probably be better to shadow the set other modes bits in some static variable in `rdp.c` and have a separate function that generates the `set other mode` command based on on that static variable. Then in this function you can just set/unset the appropriate bits without effecting other state that people might want to keep such as the sample type, blend state, cvg coverage, etc.

I'm not sure if these changes are appropriate for this PR, or for a future one that will overhaul `rdp.c`. But I agree with @snacchus, make sure you follow the existing conventions in `rdp.c`

src/rdp.c
snacchus: This is very interesting. I encountered something like this before as well but wasn't sure whether it was an error on my side or the RDP documentation is inaccurate (recently I tend more and more towards just not trusting SGI's docs on anything). In theory, it should be fine to calculate the slopes in full precision, because the way `xh`, `xm` and `xl` are calculated should account for the fact that the Y-coords are truncated to 2 fractional bits. @anacierdem what did you end up doing in your code?

(Also don't worry about the implicit cast. Truncating to integer should be the right thing here)

anacierdem: Doing full precision arithmetic is just fine, though mine was based on 16.16 fixed numbers but I don't think using floats would be an issue either. So we should ideally do not have anything other than xh, xm, xl adjustments. If you are still having seams, I suggest checking color combiner/othe mode settings.

src/rdp.c
anacierdem: Actually this will be the most difficult part of the API to decide IMO. There are cases where specific flags should be set in order for a specific primitive to render correctly and it is not trivial to manage. Ideally we would want RDP state commands to be easily composable to be able to set whatever flag the user wants to set but OTOH that will expose too much detail for a high-level API such as we have now in `rdp.c`.

There are a few strategies that we can adopt;

- We can implement a standard media/gl library API but completely covering it will take a long time.
- @rasky's idea is embedding these RDP preparation commands into the primitive calls which will "cache" the state such that succesive calls will not set them over and over. IMO this is a really good idea but interoperability with the lower level commands is a little tricky as the cached versions might get stale if the user issues additional commands and the code may have to peek into all the commands issued, which is something I don't like very much.
- Some form of a compile time composition of commands that can produce command lists that can accept run-time parameters. C preprocessor might be an option but I'm not sure if it will enable enough flexibility (I'm no good at it at all). We might consider another tool or even a custom language implementation. The major drawback would be all calls should be made via that system such that it can decide which settings to re-issue and drop etc. I mean we can statically analyze as mush as we can and optimize beforehand. But this is not a hard requirement, we can use this for caching or use it to implement the gl library. The advantage is that we can change the implementation and even move it to RSP/RDP as we go without breaking existing code much (hopefully). If implementation requires doing command peeks, this will still enable that as all calls go throught the mechanism though we probably don't have to. We can still implement this in a crude way without any optimizations at first. Imagine we create this new method like this (this is totally made up at this point);

    ```
    rdp_draw_shaded_triangle(x1, y1, ..., v1G, v1R...) = rdp_draw_pipe(
    set_other_modes({
        .blend_enable = true,
        ...
    }),
    edge_coeffs(x1, y1,..),
    shade_coeffs(v1G, v1R,..),
    )
    ```

    Then it is possible to compose such constructs into bigger and bigger things;

    ```
    rdp_draw_blended_tri_strip(x1, y1, ..., v1G, v1R...) = rdp_draw_pipe(
    edge_coeffs(x1, y1,..),
    edge_coeffs(x2, y2,..),
    )
    ```
    This is still not a simple thing to achieve but would be really powerful if we have it.

rasky: First thing: I think that `rdp_enable_shading` is fine they way it is now, with more documentation and maybe a different naming. It follows the existing rdp.c convention and we can't fix that in the context of this PR.

Now thinking about the future. I have a large set of rdp.c changes that I'd like to submit soon, especially to improve support for 2D sprites and custom render modes. Those changes keep the current spirit / architecture of rdp.c and show even better the problem with the current rdp render modes functions, that I will try to summarize.

I think the real culprit is that the RDP command `set other modes` combines render flags in a way that makes it hard to make a composable API. Some flags are in fact related to the actual texture being rendered (eg. color mode, palette), while others are related to render modes which should be specificied at render time (do I want to render opaque or transparent?). On top of this, render modes *do affect* the commands that need to be sent to render; for instance switching from COPY to 1CYCLE is not just a render mode change, but you need to account for it when providing horizontal deltas in sprites.

So basically, if you want to have a "modern API" where you first set a render mode (eg. "I want to do 1CYCLE with this color combiner formula to achieve a cool effect") and then you draw something (eg. "Let me draw this texture which I don't even remember whether it's paletted or 16-bit"), you can't achieve that without having a render mode global state. The code needs to know how to merge the render mode state with the flag bits required to draw the specific texture pixel format, and then later it needs to check the current render mode to understand how to calculates the deltas.

So this is the unfortunate situation we live in. I don't think it's possible to have a powerful 2D (or 3D) API without keeping this state. So, in my local changes of rdp, I'm introducing this concept. There is a global variable where the code stores the current render mode (that is, the 64-bit `set other mode` command that was last sent), and then work on it incrementally when needed, or read it if required to understand how to perform calculations.

This is not a problem with the current rdp.c API, and in fact it works wonderfully. It can become a problem if we want to open up a API to let users enqueue custom RDP commands. If the user sent a `set other modes` command through this custom "enqueue whatever" API, they would effectively introduce a bug with the global cache. So my current solution is *not* to open up such API for now in my commit, and simply add one specific API to configure a custom render mode (which I *know* it will end up as a `set other modes`).

Going even further in the future, I don't think having rdp.c directly using CPU to put commands into RDP is that useful. We briefly discussed it before on Discord, but I think it would be much powerful to have RDP commands go through RSP anyway, and then let RSP grow a few features to speed up 2D games, such as applying the matrix stack to 2D sprite rendering (so that you can have static display lists for sprites and tilemaps and simply move them on the screen using the matrix stack). At that point, the RSP would be the obvious place where to keep the global state of the current render mode; and it would be even possible to add the "enqueue raw RDP commands" API as that would not break anything, as the RSP would see the render mode changes and act accordingly.

anacierdem: In the scope of this PR Im ok with a simple mode change function as well even though in the long run that will start to become a problem especially considering our backwards compatibility efforts. I am not opposing to caching the config either, I just dont like that... It might be something we have to have in the end. The composition should ideally enable different implementation strategies including caching - it may even result in better optimization strategies. Also we dont need to map the composable units 1 to 1 to the hardware like in my example, we can come up with a good abstraction only considering the "capabilities" of the sytem. I think this also closely relates to @snacchus's intermediate representation idea. OTOH I havent worked with RDP enough to decide on something.

src/rdp.c
space222: If I don't use my toy GL1.0 I wrote, I can make triangles touch in the obvious same-pixel-coordinates way without seams, so I've removed the alteration of the input values in the latest commit.

space222: Looks like there's something off about the colors now on hardware, so now something else to dig into.

anacierdem: I havn't worked on colors before, but if ou can share we may try to decipher the problem.

space222: After getting my new capture dongle working, it looks fine on capture, but on my TCL Series-5 it has a green stripe down the left side and white dots at the corners. Capture attached.
n64shot

anacierdem: I will try to run this ans see for myself

anacierdem:
I will review this as soon as possible. Did the color issue went away?

space222:
> I will review this as soon as possible. Did the color issue went away?

I don't have another (easy) way to test. I would not be at all surprised if it is just my tv. I can post the rom from the above screenshot on discord if you want to run the same thing and see.

rasky:
Closing this, because the [unstable branch](https://github.com/DragonMinded/libdragon/wiki/Unstable-branch) features a full implementation of RDP triangle drawing for all possibile formats (filled, shaded, textured, z-buffered), even with a RSP-accelerated ucode, that has been carefully tested to provide no seams. See [rdpq_tri.h](https://github.com/DragonMinded/libdragon/blob/unstable/include/rdpq_tri.h), which is part of the new `rdpq` library.

The old `rdp.h` library will be deprecated as soon as we start merging `rdpq` (will do that piecewise), so we are not going to add new features to it.

PR #219, opened 2021-10-31
ym64: switch to requiring filesystem-prefix for filenames

rasky:
Just like xm64, it's better to have public APIs to accept filenames
with filesystem prefix. In the case of ym64, it does also work on
all filesystems as the I/O is far from critical (while on xm64, we
currently only support rom://)


anacierdem (approved):
LGTM

PR #221, opened 2021-11-01
toolchain: compile newlib with optimizations

rasky:
Ooops...

I have broken it here:
https://github.com/DragonMinded/libdragon/commit/1c522f2da5928bb5bb79ac76436c2186091fef75#diff-4cdfece525412cfc651464566e6bc1e6a8c333a3d67831a751e8c78f3a82ab69L93-R93

but I had removed `CFLAGS="-O2"` because `CFLAGS` was being ignored in the compilation of Newlib (I knew because I had tried to add the `-DHAVE_ASSERT_FUNC` in `CFLAGS` and it was being ignored). So I moved to `CFLAGS_FOR_TARGET` and I assumed that nothing was necessary.

It looks like for unknown (to me) reasons, `CFLAGS="-O2"` was indeed having an effect.

I've also took the chance to modify the CI to upload all the built ROMs as artifacts, which costs nothing and since we build them, we can as well make them available for download from the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,14 @@ jobs:
           ghcr.io/dragonminded/libdragon:latest \
           ./build.sh
 
+      - name: "Upload built ROMs to artifacts"
+        uses: actions/upload-artifact@v2
+        with:
+          name: roms
+          path: |
+            ${{ github.workspace }}/examples/**/*.z64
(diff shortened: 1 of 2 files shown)


sp1187 (approved):

anacierdem:
I remember the day we decided to remove that, how did you find out?
PR LGTM.

PR #222, opened 2021-11-02
do_ctors: fix (again) constructors when using ld linker

rasky:
In #195, we fixed a bug (off-by-one) in __do_global_ctors that caused
the last constructor not to be called in C programs (using ld linker).

This in turn broke C++ software, that uses g++ as linker. We found out
that ld and g++ creates different constructor tables. We attempted a
first fix in #197 by switching n64.mk to use g++, but this broke the
usage of ld linker; so we fixed it again in #201 by separating the
codebase for ld (__do_global_ctors) vs g++ (__wrap___do_global_ctors).

Unfortunately, in doing so in #201, we reverted to the version of
__do_global_ctors that had the initial bug, effectively reverting #195.
So, all in all, we still have code built by ld broken because the
last constructor (eg: init_interrupts) is not called.

So this PR introduces again the fix in #195, hopefully fixing it finally
for good.

### DIFF
--- a/src/do_ctors.c
+++ b/src/do_ctors.c
@@ -35,7 +35,9 @@ void __do_global_ctors()
 {
 	func_ptr * ctor_addr = &__CTOR_END__ - 1;
 	func_ptr * ctor_sentinel = &__CTOR_LIST__;
-	while (ctor_addr > ctor_sentinel) {
+	assertf((uint32_t)*ctor_sentinel != 0xFFFFFFFF,
+		"Invalid constructor sentinel.\nWhen linking with g++, please specify:\n   --wrap __do_global_ctors");
+	while (ctor_addr >= ctor_sentinel) {
 		if (*ctor_addr) (*ctor_addr)();
 		ctor_addr--;
 	}


rasky:
I tested this by running on N64 vtest (built with ld), cpptest and audioplayer (built with g++). They all seem to work.

meeq:
["There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors." -- Jeff Atwood](https://twitter.com/codinghorror/status/506010907021828096?lang=en)

anacierdem:
AFAIR this will break the C++ example with a legacy build.

anacierdem:
Yes, I just verified that it does not work with legacy C++ builds. We either revert the auto init or live with it IMO. Though there is a relatively simple fix: if you want to update libdragon and still link with an old g++ setup, add `-Wl,--wrap -Wl,__do_global_ctors` to the flags. If we document this somewhere easily discoverable, I think we will be fine.

The condition we are hitting is trying to call a function at `0xFFFFFFFF` and get an address error. Maybe we can add an assert in `__do_global_ctors()` like;

    assertf(*ctor_addr = !0xFFFFFFFF, "When linking via g++, use the --wrap flag");

What do you think?

For some reason I tought going before #195 would maintain compatbility, I was so wrong.

rasky:
@anacierdem I've pushed the assert. Can you please verify again if a legacy g++ build now hits it? I've been unable to test it myself.

anacierdem:
@rasky It does hit that assert and still works ok with ` -Wl,--wrap -Wl,__do_global_ctors`. If you want to test it, you can use;
```Makefile
PROG_NAME = cpptest

ROOTDIR = $(N64_INST)
GCCN64PREFIX = $(ROOTDIR)/bin/mips64-elf-

CC = $(GCCN64PREFIX)gcc
CXX = $(GCCN64PREFIX)g++
AS = $(GCCN64PREFIX)as
LD = $(GCCN64PREFIX)ld
OBJCOPY = $(GCCN64PREFIX)objcopy
N64TOOL = $(ROOTDIR)/bin/n64tool
CHKSUM64 = $(ROOTDIR)/bin/chksum64

CFLAGS = -std=gnu99 -march=vr4300 -mtune=vr4300 -O2 -Wall -Werror -I$(ROOTDIR)/mips64-elf/include
CXXFLAGS = -std=c++11 -march=vr4300 -mtune=vr4300 -O2 -Wall -Werror -I$(ROOTDIR)/mips64-elf/include
LDFLAGS = -L$(ROOTDIR)/mips64-elf/lib -ldragon -lc -lm -lstdc++ -ldragonsys -Tn64.ld -Wl,--gc-sections -Wl,--wrap -Wl,__do_global_ctors
N64TOOLFLAGS = -l 1M -h $(ROOTDIR)/mips64-elf/lib/header -t "C++ Test"

ifeq ($(N64_BYTE_SWAP),true)
$(PROG_NAME).v64: $(PROG_NAME).z64
	dd conv=swab if=$^ of=$@
endif

$(PROG_NAME).z64: $(PROG_NAME).bin
	$(N64TOOL) $(N64TOOLFLAGS) -o $@ $^
	$(CHKSUM64) $@

$(PROG_NAME).bin: $(PROG_NAME).elf
	$(OBJCOPY) $< $@ -O binary

# Use g++ to link the compiled C++ object into an ELF executable
$(PROG_NAME).elf: $(PROG_NAME).o
	$(CXX) -o $@ $^ $(LDFLAGS)

.PHONY: clean
clean:
	rm -f *.v64 *.z64 *.elf *.o *.bin
```

As the C++ example's Makefile

src/do_ctors.c
```diff
@@ -35,15 +35,17 @@ void __do_global_ctors()
...
+	while (ctor_addr >= ctor_sentinel) {
 		if (*ctor_addr) (*ctor_addr)();
 		ctor_addr--;
 	}
 }
 
 /**
  * @brief Execute global constructors
- * This version is used by the new build system (n64.mk) via the --wrap linker
+* This version is used by the new build system (n64.mk) via the --wrap linker
```
anacierdem: I think we removed a space by accident?

anacierdem (approved):

PR #223, opened 2021-11-03
Minor fixes to SD writing support

anacierdem:
- Change PI writes to wait for the ready bit.

### DIFF
--- a/src/debug_sdfs_64drive.c
+++ b/src/debug_sdfs_64drive.c
@@ -74,11 +74,28 @@ static DRESULT fat_disk_write_64drive(const BYTE* buff, LBA_t sector, UINT count
 
 			uint32_t* dst = (uint32_t*)(D64_CIBASE_ADDRESS + D64_BUFFER);
 			u_uint32_t* src = (u_uint32_t*)buff;
+			disable_interrupts();
 			for (int i = 0; i < 512/16; i++)
 			{
 				uint32_t a = *src++; uint32_t b = *src++; uint32_t c = *src++; uint32_t d = *src++;
-				*dst++ = a;          *dst++ = b;          *dst++ = c;          *dst++ = d;
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = a;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = b;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = c;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = d;
+				MEMORY_BARRIER();
 			}
+			enable_interrupts();
 		}
 
 		usb_64drive_wait();
--- a/src/debug_sdfs_ed64.c
+++ b/src/debug_sdfs_ed64.c
@@ -542,11 +542,29 @@ static DRESULT fat_disk_write_everdrive(const BYTE* buff, LBA_t sector, UINT cou
 
 			uint32_t* dst = (uint32_t*)(ED64_BASE_ADDRESS + ED64_SD_IO_BUFFER);
 			u_uint32_t* src = (u_uint32_t*)buff;
+
+			disable_interrupts();
 			for (int i = 0; i < 512/16; i++)
 			{
 				uint32_t a = *src++; uint32_t b = *src++; uint32_t c = *src++; uint32_t d = *src++;
-				*dst++ = a;          *dst++ = b;          *dst++ = c;          *dst++ = d;
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = a;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = b;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = c;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = d;
+				MEMORY_BARRIER();
 			}
+			enable_interrupts();
 		}
 
 		everdrive_sd_crc16((void*)buff, crc);


src/debug_sdfs_64drive.c
```diff
@@ -74,11 +74,28 @@ static DRESULT fat_disk_write_64drive(const BYTE* buff, LBA_t sector, UINT count
...
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = c;
+				MEMORY_BARRIER();
+				dma_wait();
+				MEMORY_BARRIER();
+				*dst++ = d;
+				MEMORY_BARRIER();
 			}
+			enable_interrupts();
```
rasky, line 98: I'm a bit worried on the interrupt latency here. Can you please measure how long we are keeping the interrupts disabled here? Also `dma_wait()` seems wrong because of its name (though I know it's doing what we expect). Maybe we should simply switch to `io_write()` / `__io_write()` here (the second if we keep the interrupts disabled ourselves)?

So I think it would be better to measure with TICKS_READ() how long it takes to do a whole run in the two scenarios: 1) keep interrupt disabled, use `__io_write()`, 2) keep interrupts enabled, call `io_write()`. The second version would switch interrupts on/off between each word write. A third option would be to enable/disable within the for cycle so that at least we re-enable them once per loop.

src/debug_sdfs_ed64.c
```diff
@@ -159,6 +159,8 @@ static bool everdrive_sd_read_response(uint8_t res_buff[5]) {
 		res_buff[2] = everdrive_sd_read_command();
 		res_buff[3] = everdrive_sd_read_command();
 		res_buff[4] = everdrive_sd_read_command();
+	} else {
+		timeout += 4;
```
rasky: Why is this required?

anacierdem: To make sure we consume the data from the sd card even if we don't use it.

src/debug_sdfs_64drive.c
anacierdem: I initially disabled/enabled in the loop but then moved it outside b/c I thought it would be less overhead. I will try to measure different usages 👍🏼

rasky: In general, as the kernel approaches, we should be wary of interrupt latency. For instance, dma.c uses the wrong pattern:

```C
void io_write(uint32_t pi_address, uint32_t data) 
{
    volatile uint32_t *uncached_address = (uint32_t *)(pi_address | 0xa0000000);

    disable_interrupts();

    while (__dma_busy()) ;
```

were it waits *with interrupts disabled* if a DMA is in progress, which means that we block all the threads for no reason (and in case of single thread, we delay processing interrupts like timers and audio again for no good reason). 

The correct pattern would be something like:

```C
void io_write(uint32_t pi_address, uint32_t data) 
{
    volatile uint32_t *uncached_address = (uint32_t *)(pi_address | 0xa0000000);

    // Wait with interrupts enabled
    while (1) {
        while (__dma_busy()) ;
        disable_interrupts();
        if (!__dma_busy()) break;
        enable_interrupts();
    }
    [...]

    enable_interrupts();
```

This makes sense if a DMA is in progress, and we don't want to wait too much with interrupts disabled. If the PI is busy without DMA (DMA busy bit 0, IO busy bit 1), then it doesn't matter as much as that should become ready "very soon" anyway. But we currently don't differentiate between the two.

src/debug_sdfs_ed64.c
rasky: Obviously the semantic is a bit different. 

 * If you pass res_buff != NULL, you always read 4 bytes (with no checks), and then consume up to 16 bytes, stopping at the first 0xFF.
 * If you pass res_buff==NULL, you consume up to 20 bytes stopping at the first 0xFF.

I don't know if it matters. If passing res_buff==NULL is just a way for the caller to avoid reading a data that is not useful, then maybe we should do something like:

```C
for (int i=0;i<4;i++) {
    uint8_t data = everdrive_sd_read_command();
    if (res_buff) res_buff[i+1] = data;
}
```

which would preserve the semantic of checking for 0xFF just in the second loop. But again, I don't understand this protocol very well.

src/debug_sdfs_64drive.c
rasky: I was thinking that to generalize this, we could have an API `pi_lock()` / `pi_unlock()` to lock/unlock the bus, where locking does the loop dance above to try to acquire the lock without keeping interrupts disabled and pi_unlock() just reenables the interrupts. So all users of the PI bus have a simple API to handle it.

src/debug_sdfs_ed64.c
anacierdem: This is a combination of possible sized responses (depending on the command we can have up to 20 bytes), it makes it a little less ergonomic but simpler. The clock line is only toggled when we actually read/write so we always need to complete the data flow specified by the protocol to keep things as expected b/c we are the host device and we decide when to continue transfers.

That's a nice implementation suggestion as well I will do that instead.

src/debug_sdfs_64drive.c
anacierdem: Once we acquire the lock, if we have a long running code, it will still introduce latency if we just use interrupts for the lock. Maybe we should work on an alternative lockless solution. (Not that I have in mind ATM)
On the other hand I totally get your point we should clarify the intent of the API (I know dma_wait was not very appropriate but wanted to start discussions around). Once we do that the locking itself can be an implementation detail though it would still matter for this PR as it would change the implementation here.
Also for very small read/writes like we have in this PR we wouldn't want too much overhead.
Then if we utilize `C11` threads, I think we can implement safer and more performant threading, WDYT? The compiler can cooperate with our needs as well that way.

rasky: Yes the idea of pi_lock is to help programmers acquire the lock in the correct way. Right now, exactly zero code does that (well, just the RDP), so it looks like the pattern is easy to be mistaken.

pi_lock doesn't prevent us from the fact that, once a lock is taken, it must be retained for not too long. I don't think there's an easy way. We currently hide the lock pattern behind high level primitive like io_read but that's a bit too little (just one memory access). I can't think of a way to design a API that allows for some more packing of memory accesses but not too many.
It's also hard to differentiate standard cases from special cases as PI is just a bus. Maybe sometimes you really need to keep bus locked and ignore interrupt latency, so it would be wrong to design a API that handcuffs you in this regard.


C11 threads is just an API. We could decide to implement it on top of our kernel if we wanted, but it's the implementation that matters. In our case, I don't think we can allow anything to happen while the lock is held: almost anything could touch the PI bus. So the only way is to have the PI lock keep the interrupts disabled.

anacierdem: It should be possible to have interrupts fire and execute as long as they do not touch PI IINM. Once they do, they can start waiting for the release of the lock.

In theory it can be more than an API. I really don't know otimizations in GCC or the memory model details of C11 but in theory it can better (and more correctly) optimize code if it is thread aware. It is not possible if we don't cooperate with the compiler.

See http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4455.html for some possible optimization examples. (this is C++ IDK if any would apply) (also http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0062r1.html)

rasky: > It should be possible to have interrupts fire and execute as long as they do not touch PI IINM. 

Which is not possible to enforce, and, what's worse, it's a problem only in the rare case where the interrupt is fired with the exact timing within the lock, so it will cause hard to reproduce and debug bugs.

> Once they do, they can start waiting for the release of the lock.

If the interrupt wait for the lock to be released, it will deadlock because the lock is held in user space, which will not resume until the interrupt is finished. If instead the interrupt bypasses the lock but just waiting for PI ready, when the user space resume execution the lock would have been violated and maybe the PI is now busy because of a DMA started under interrupt... but the user space thinks it has acquired the lock.

anacierdem: Ok I was incorrectly referring to `dma_wait` by "wait" here. `pi_lock` is the necessary devil for such small one off read things then? IDK what would be best b/c with such a lock method, there would be too much overhead.

rasky:
Not applicable anymore after #374 (libcart).

PR #225, opened 2021-11-05
toolchain: Update toolchain (GCC 11.2, Binutils 2.37)

sp1187:


tools/build-toolchain.sh
```diff
@@ -110,7 +108,7 @@ cd ..
 rm -rf gcc_compile
 mkdir gcc_compile
 cd gcc_compile
-CFLAGS_FOR_TARGET="-G0 -O2" CXXFLAGS_FOR_TARGET="-G0 -O2" ../"gcc-$GCC_V"/configure \
+CFLAGS_FOR_TARGET="-O2" CXXFLAGS_FOR_TARGET="-O2" ../"gcc-$GCC_V"/configure \
```
meeq, line 111: What is the context behind this change?

sp1187: Not strictly a required part of the toolchain update, but figured out that this was likely a mistake as no other libdragon build scripts or makefiles disable the global data section, and I see no reason why it would be necessary or beneficial here.

rasky (approved):
Fine by me, I'll leave it to @anacierdem

PR #228, opened 2021-11-13
Improvements to toolchain builds and actions

rasky:
The first commit does some little mix improvement in toolchain building.

The second commit allows to push Docker images even during PRs, to simplify local testing of toolchain changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      # Create a lower cased version of the repo name. This is required
+      # because Docker supports only lowercase names in the registry, while
+      # a repo name on GitHub can have uppercase letters.
+      - name: Downcase repository name
+        run: |
+          echo "GITHUB_REPOSITORY_LOWERCASE=${GITHUB_REPOSITORY,,}" >>${GITHUB_ENV}
+
       # We first compare the default branch to the event commit. In case this is
       # the default branch, this should return 0, skipping the build and using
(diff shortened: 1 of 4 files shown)


rasky:
@anacierdem @networkfusion do you know why the CI is failing with authorization error? Any idea?

networkfusion:
@rasky I am getting the same, but have not had a chance to look into it yet. Although I suspect repo permissions as outlined in  https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio

At minimum, this seems solved in my PR against your repo.

anacierdem:
I have worked on this for quite some time but the thing we are trying to achieve is not possible as originally intended.
As I have described in more detail [here](https://github.com/DragonMinded/libdragon/issues/230), it is not possible to push a package from a PR workflow run initiated by a public fork.
I have fixed other issues and the workflow is now configured to push an image from the existing repository on a push event. This means that individual forks will create their own images and we will be able to use them if we want to test them.
For pull requests, a temporary image is created but not pushed as it was the case before.

anacierdem:
Ideally we probably want the image pushes on the `Docker Build` workflow to have it in a single place. Once this start working as expected, I can do that change as well.

anacierdem:
I have verified that everything is working as expected with the latest changes. @rasky you can merge whenever you are ok with the changes. I think it would make better sense to squash the last two commits, but we can go with a full squash merge as well if you like.
We can do more improvements later.

PR #229, opened 2021-11-13
Initial support for iQue/BBPlayer

rasky:
iQue / BBPlayer is a Chinese variant of Nintendo 64 which has a slightly different hardware and operating system. This PR adds some basic support to libdragon so that it is possible to at least boot the ROMs. ~~The macro N64_BBPLAYER is introduced, and it must be changed to 1 at compile-time to build a iQue-compatible version of libdragon~~. Detection is made at boot time, and at runtime it is possible to call `sys_bbplayer()` to check whether we are running on an iQue Player or not.

The CPU frequency is higher compared to a stock N64, so timers run faster. In this patch, the CPU_FREQUENCY is adjusted so that the timers example pass. Notice that `wait_ticks` is not changed, so games using `wait_ticks` and hardcoding a fixed number of ticks might break with iQue as the delay would be shorter there. We must decide whether to keep this approach, or rather transparently scale all timer-related functions to the original N64 frequency.

A bit is also changed in the display registers otherwise display is garbled. A change is also made to n64tool, to create ROMs which are 16Kb-padded by default, which makes them compatible with iQue. This doesn't affect people explicitly specifying a size (that is still respected), but it is the new default behavior.

This is just the beginning, there are many other differences, but we have to start somewhere.

This PR has been tested on real hardware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ programming and debugging. These are the main features:
   to use 64-bit registers and opcodes with it)
 * Can be developed with newer-generation emulators (cen64, Ares, Dillonb's n64,
   m64p) or development cartridges (64drive, EverDrive64).
+* Support both vanilla N64 and iQue Player (chinese variant). The support is
+  experimental and done fully at runtime, so it is possible to run ROMs built
+  with libdragon on iQue without modifying the source code.
 * In-ROM filesystem implementation for assets. Assets can be loaded with
   `fopen("rom://asset.dat")` without having to do complex things to link them in.
 * Efficient interrupt-based timer library (also features a monotone 64-bit
--- a/tools/n64tool.c
+++ b/tools/n64tool.c
@@ -28,9 +28,25 @@
 #include <sys/errno.h>
 #endif
 
+#define ROUND_UP(n, d) ({ \
+	typeof(n) _n = n; typeof(d) _d = d; \
+	(((_n) + (_d) - 1) / (_d) * (_d)); \
+})
+
+// Minimum ROM size alignment, used by default. We currently know of two constraints:
+//  * 64drive firmware has a bug and can only transfer chunks of 512 bytes. Some
+//    tools like UNFloader and g64drive work around this bug by padding ROMs,
+//    but others (like the official one) don't. So it's better in general to
+//    pad to 512 bytes.
+//  * iQue player only allows loading ROMs which are multiple of 16 Kbyte in size.
+//  
+// To allow the maximum compatibility, we pad to 16Kb by default. Users can still
+// force a specific length with --size, if they need to.
+#define PAD_ALIGN    16384
+
 #define WRITE_SIZE   (1024 * 1024)
-#define HEADER_SIZE  0x1000
-#define MIN_SIZE     0x100000
+#define HEADER_SIZE  0x1000          // Size of the header (which includes IPL3)
+#define MIN_SIZE     0x100000        // Minimum size of a ROM after header (mandated by IPL3)
 
 #define TITLE_OFFSET 0x20
 #define TITLE_SIZE   20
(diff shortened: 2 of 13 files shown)


CrashOveride95:
> transparently scale all timer-related functions to the original N64 frequency.

This is what the original devs chose to do originally, but bear in mind their priority was for it to work with old code.
Perhaps there's a better way to go about this today
I also remember a address being used to tell if the hardware was a iQue or not, that might be a good idea to avoid having to build seperate libraries for n64/ique, but unsure if this might affect performance

rasky:
We had a bit of discussion on Discord and it seems the majority of people feel that it would be vastly better if libdragon supported iQue at runtime, without a compile time switch, so to always generate ROMs which are iQue compatible.

I will need to investigate this if we can find a way without causing too much overhead; I'm concerned about the frequency change mainly, I'll do some tests.

rasky:
I have checked the performance. Basically, GCC is able to do what I hoped. For instance, this code:

```C
wait_ticks(TICKS_FROM_MS(50))
```

is expanded as follows:

```C
wait_ticks(50 * ((__bbplayer ? 140625000 : 93750000) / 2 / 1000))
```

which means that it would require at least one division at runtime (depending on the flag). But GCC compiles it to:

```C
__bbplayer ? wait_ticks(3515625) : wait_ticks(2343750);
```

so that the calculations are done at compile-time in the two versions. This is enough to dismiss the concerns I had (even if in this example obviously it doesn't matter if we do one division more if we're going to sleep anyway...).

The new implementation with runtime detection has been tested on real hardware and it works (timers test, which is for now the only one that has been extensively tested). I think we can review and merge this.

tools/n64tool.c
```diff
@@ -428,11 +453,11 @@ int main(int argc, char *argv[])
 			return print_usage(argv[0]);
 		}
 	}
-	else if((total_bytes_written % 512) != 0)
+	else if(((total_bytes_written+HEADER_SIZE) % PAD_ALIGN) != 0)
 	{
-		/* Pad size to 512 bytes to avoid possible incompatibilities with 64drive
+		/* Pad size to 16384 bytes 512 bytes to avoid possible incompatibilities with 64drive
```
sp1187: This looks like a typo, the old text should probably be removed.

rasky: Done

rasky:
Ok, now the whole test ROM passes on iQue. I've disabled testing the timing of tests as the numbers are different (plus there's something weird going on which I couldn't debug at this moment, there was a floating point exception in timing calculations that didn't make sense -- anyway, maybe it's too hard to debug it in this context). 

I've also adjusted `dma_read` to the slightly-different-and-still-undocumented behaviors of iQue PI DMA for misaligned transfers, so that they work correctly. The audioplayer example now plays beautiful music on both consoles.

README.md
```diff
@@ -15,6 +15,9 @@ programming and debugging. These are the main features:
   to use 64-bit registers and opcodes with it)
 * Can be developed with newer-generation emulators (cen64, Ares, Dillonb's n64,
   m64p) or development cartridges (64drive, EverDrive64).
+* Support both vanilla N64 and iQue Player (chinese variant). The support is
```
sp1187, line 18: Saying that "**all** ROMs will work on both systems" sounds a bit overpromising, it should probably still be considered unfinished and experimental for now. Also, even when it's more mature it will probably still depend on which features are used and if the ROM does anything stupid, buggy or just something out of the ordinary API usage patterns.

rasky: That's true but it's also overly generic. Can we say that all libdragon ROMs work on vanilla N64 either? Not at all, obviously the programmer can always do something that prevents them from working.

I guess it's more a styling issue here. The concept to convey is that you don't need to do anything specific to make ROMs work on iQue. Can you suggest a phrasing that avoids the word "all" which seems to be the one that can be misunderstood?

networkfusion: How about "current libdragon example ROMs and integration tests"

rasky: That's missing the information that we had not to change the examples to make them work. That is, what I would want to say is that the design choice is that user code shouldn't change at all to run on ique

rasky: I've sent a new wording, does this sound better?

sp1187: Sounds good enough to me.

rasky:
Pending #233, I've modified the n64tool commit to avoid the warning for explicit sizes which are not multiple of 16 KiB, which might be surprising for users for now. We're always on time to add it later. The rest is good to go.

PR #231, opened 2021-11-24
build.sh: build examples after clobber, not before

rasky:
For some reason, build.sh was running a make clobber before building
the tests, that was deleting also all built examples. This caused
the CI not to store the example ROMs as artifacts.

Thanks to @networkfusion for the analysis.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -31,9 +31,9 @@ makeWithParams install
 popd
 rm -Rf $LIBMIKMOD_DIR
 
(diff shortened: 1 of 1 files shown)


rasky:
@anacierdem do you know why there was that make clobber there?

rasky:
@anacierdem explained me the rationale, so I changed the patch to clobber before build examples. I included the rationale in a comment. Running a final test.

PR #232, opened 2021-11-28
Misc batch of audio fixes

rasky:
This PR includes several commits that I've done over the past weeks to fix several bugs. It should be reviewed commit by commit, as they are mostly unrelated changes.

The most important is `audio: avoid allowing to overwrite a buffer that was currently played`, which fixes a historical bug in audio.c where the currently-played audio buffer could be overwritten because of race conditions, causing audio cracks or silence. This typically happened at the begin or end of an audio playback.

`mixer: fix two undefined behaviors` fixes an uninitialized data bug that triggers when the RSP mixer was used together with other RSP ucodes (alternating them) and could find random data in DMEM. This will turn useful as we grow more ucodes in libdragon.

`audioconv64: fix off-tune bug with short forward loops` is an important fix for some out-of-tune playback in chiptune XM modules that used very short 8-bit waveforms with loops.

The others are really minor stuff.

### DIFF
--- a/tools/audioconv64/conv_xm64.c
+++ b/tools/audioconv64/conv_xm64.c
@@ -20,6 +20,11 @@
 
 #include "mixer.h"
 
+// Loops made by an odd number of bytes and shorter than this length are
+// duplicated to prevent frequency changes during playback. See below for more
+// information.
+#define XM64_SHORT_ODD_LOOP_LENGTH  1024
+
 // Bring libxm in
 #include "../../src/audio/libxm/play.c"
 #include "../../src/audio/libxm/context.c"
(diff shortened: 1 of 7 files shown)


tools/audioconv64/conv_xm64.c
```diff
@@ -77,10 +77,37 @@ int xm_convert(const char *infn, const char *outfn) {
...
+				// xm64.c will decrease the loop length by 1 byte to playback them,
+				// but this can affect the period in case of short loops:
+				// for instance, a 13-bytes loop shortened 12-bytes change the
+				// period by 7%, which can be several notes of difference at
+				// high frequencies.
+				// So for short loops (<1000 bytes), we just duplicate the loop
+				// itself to make it of even size. For longer loops, the period
+				// error made by xm64 when shortening is < 0.1%, which isn't
+				// audible.
+				if (bps == 1 && loop_length%2 == 1 && loop_length < 1024) {
```
snacchus: Defining the loop length threshold as a constant would be better I think. Maybe even go as far as making it configurable via CLI argument (If there is any benefit to that)?

snacchus (commented):
Looks good to me. Just added one suggestion in `conv_xm64.c`.

tools/audioconv64/conv_xm64.c
rasky: Done. I don't think it makes sense to expose it as a command line flag; it's a rather obscure implementation detail that is hard to explain and tuning it has very limited effect on file size.

Ryzee119:
I compiled this against my jam entry and all worked perfect. I heavily use this mixer work>
Code changes look good to me!

vieux:
Same here, used it on our jam entry, and worked fine! (only using wav64)

rasky:
Thanks! I'm merging this then

PR #234, opened 2021-12-04
joybus: implement asynchronous API and use it to implement controller autoscan

rasky:
This PR has two goals:

 * Improve interrupt latency during Joybus operations. Joybus operations are very slow and they currently run with interrupts fully disabled, which means that it's impossible for the application to implement background graphic/audio update as they are running. This would be the same also with the multithreading kernel, as disabling interrupts prevents scheduling a different thread.
 * `controller_scan()` currently takes 10% (!) of CPU frame time. Since it's a blocking function and there is no multithreading support, all applications are wasting 10% of their CPU frame budget on this.

To fix both issues, we rework Joybus support to use an asynchronous queue of messages. The messages are stored in a ring buffer and the buffer is executed under interrupt, using the SI DMA interrupt to synchronize with PIF. This is done via a new API called `joybus_exec_async`, which is private for now. `joybus_exec` is rewritten on top of `joybus_exec_async` so that all existing code continues to work (be it controller pak, EEPROM, RTC or whatever else), but waiting for the reply to arrive cam now be done with interrupts enabled without risking race conditions.

Then, we change `controller.c` to use the new internal API to implement a autoscan feature: controllers are scanned in background every frame (at vsync). `controller_scan()` is now used only to copy the state updated into interrupt
into a frozen state, to be able to inspect it without risking race conditions. This effectively fixes the performance issue, at the cost of a little (sub-frame) controller latency (as the game accesses a cache refreshed at each vsync, rather than reading the stata at the moment it is ready to process). If the latency proves to be a problem, we can add a API to ask to autoscan controllers multiple times per frame, or the application can still use the alternative blocking API `controller_read` which still works of course (and still wastes 10% of CPU frame time).

The PR is a bit noisy because I took the chance of moving EEPROM functions into their own file. Reviewing commit by commit might be easier.

### DIFF
--- a/src/joybus.c
+++ b/src/joybus.c
@@ -1,7 +1,7 @@
 /**
  * @file joybus.c
  * @brief Joybus Subsystem
- * @ingroup joybus
+ * @ingroup lowlevel
  */
 
 #include <string.h>
@@ -10,7 +10,7 @@
 
 /**
  * @defgroup joybus Joybus Subsystem
- * @ingroup libdragon
+ * @ingroup lowlevel
  * @brief Joybus peripheral interface.
  *
  * The Joybus subsystem is in charge of communication with all controllers,
@@ -19,18 +19,32 @@
  * for communicating with the serial interface (SI) registers to send commands
  * to controllers (including Controller Paks, Rumble Paks, and Transfer Paks),
  * the VRU, EEPROM save memory, and the cartridge-based real-time clock.
+ * 
+ * This module implements just the low-level protocol. You should use it
+ * only to implement an unsupported peripherals. Otherwise, refer to the
+ * higher-level modules such as:
  *
- * For Controller Pak functions, refer to the
- * @ref mempak "Mempak Filesystem Routines".
- *
- * For Transfer Pak functions, refer to the
- * @ref transferpak "Transfer Pak Interface".
- *
- * For all other controller-related functions, refer to the
+ * For controllers:
  * @ref controller "Controller Subsystem".
  *
- * For real-time clock functions, refer to the
- * @ref rtc "Real-time Clock Subsystem".
+ * For EEPROM, RTC and other peripherals:
+ * @ref peripherals "Peripherals Subsystem".
+ *
+ * Internally, the JoyBus subsystem communicates with the PIF controller via
+ * the SI DMA, via the JoyBus protocol which is a standard master/slave
+ * binary protocol. Each message of the protocol is a block of 64 bytes, and
+ * can contain multiple commands. Currently, there are no macros or functions
+ * to help composing a JoyBus message, so higher-level libraries currently
+ * hard code the binary messages. 
+ * 
+ * All communications is made asynchronously because SI DMA is quite slow:
+ * its completion is bound to the PIF actually processing the data, rather than
+ * just being the memory transfer. A queue of pending JoyBus messages is kept
+ * in a ring buffer, and is then executed under interrupt when the previous SI DMA
+ * completes. The internal entry point is #joybus_exec_async, that schedules
+ * a message to be sent to PIF, and calls a callback with the reply whenever
+ * it is available. A blocking API (#joybus_exec) is made available for
+ * simpler usage.
  *
  * @{
  */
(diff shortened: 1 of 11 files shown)


src/joybus.c
```diff
@@ -57,293 +71,224 @@ static volatile struct SI_regs_s * const SI_regs = (struct SI_regs_s *)0xa480000
...
-static void __SI_DMA_wait( void )
+__attribute__((constructor))
+void __joybus_init(void)
 {
-    while (SI_regs->status & (SI_STATUS_DMA_BUSY | SI_STATUS_IO_BUSY)) ;
+    // FIXME: this constructor requires the __init_interrupts constructor to be
+    // already run. Since we are not 100% sure of how GCC handles constructor
+    // ordering, we call it explicitly here (there's no harm in calling it
+    // multiple times anyway). Revisit this after gathering more information
+    // on constructor ordering.
```
anacierdem, line 109: Just a note: from my experience currently we call constructors in reverse order with respect to the order for which we provide the object files to the linker. This felt ok to me but I'm not sure on the spec. OTOH haven't experimented with per-translation-unit ordering yet.

src/joybus.c
```diff
@@ -57,293 +71,224 @@ static volatile struct SI_regs_s * const SI_regs = (struct SI_regs_s *)0xa480000
...
+/** @brief Pending messages read index */
+static volatile int msgs_ridx;
+
+static void si_interrupt(void);
+
+/**
+ * @brief Initialize the joybus subsystem
  */
-static void __SI_DMA_wait( void )
+__attribute__((constructor))
```
anacierdem, line 102: I think this is OK but if someone wants to disable this maybe we should provide a way?

src/joybus.c
```diff
@@ -57,293 +71,224 @@ static volatile struct SI_regs_s * const SI_regs = (struct SI_regs_s *)0xa480000
...
-    {
-        case EEPROM_16K: return 256;
-        case EEPROM_4K: return 64;
-        default: return 0;
+static void si_interrupt(void) {
+    joybus_msg_t *msg;
+
+    switch (joybus_state) {
+    case JOYBUS_STATE_SENDING:
+        // Message sending complete. Start receuving the reply
```
anacierdem: type: receiving

src/joybus.c
```diff
@@ -57,293 +71,224 @@ static volatile struct SI_regs_s * const SI_regs = (struct SI_regs_s *)0xa480000
...
 }
 
 /**
- * @brief Probe the EEPROM interface on the cartridge.
- *
- * Inspect the identifier half-word of the EEPROM status response to
- * determine which EEPROM save type is available (if any).
- *
- * @return which EEPROM type was detected on the cartridge.
+ * @brief Check where there are new messages to send
```
anacierdem, line 167: It might be nice to add a comment to signify interrupts must be disabled when calling `joybus_state` manipulating functions.

anacierdem (approved):
Looks good to me 👍

src/joybus.c
rasky: Which case are you thinking of? The constructor is not called if any other function of joybus.c is not called. So if you're not using joybus at all, the constructor is not called. Otherwise, you need the interrupt to be able to use joybus.

The alternatives are:

1) initialize the interrupt lazily on the first joybus_exec_async call. That requires a check every call. I would say the constructor approach achieves the same without the very little overhead, but I agree it's more obscure.
2) add an explicit joybus_init function, and call it from all joybus-related init functions (eg: controller_init). This works as well but breaks usages of joybus_exec in the wild as no init function was required before. joybus_exec is marked as experimental API though.

Any preference?

anacierdem: > The constructor is not called if any other function of joybus.c is not called.

I totally forgot about that. That makes this much less important and I think the constructor is fine 👍

src/joybus.c
rasky: I think C++ mandates that per-translation-unit order is defined to follow the source code order. By executing the constructors in reverse order, I think we're violating that and possibly causing bugs in C++ code. Anyway, I decided to punt for this PR to avoid the rabbit hole effect

src/joybus.c
rasky: Done

src/joybus.c
rasky: Done

PR #236, opened 2022-01-04
Clean rsp_load cache when calling rsp_load_code directly

lmancini:
Consider a program which calls `rsp_load` first and `rsp_load_code` later; `rsp_load` keeps a cache of the last loaded ucode, for performance reasons: so in order for a later invocation to `rsp_load` to actually execute, `rsp_load_code` should always clear the cache.

Otherwise, `rsp_load` would erroneously assume that the ucode trying to be loaded was already up, while in fact it was the one loaded by `rsp_load_code`.


snacchus (approved):
Looks good to me!

PR #238, opened 2022-01-04
Adding function to enable changing S and T

stefanmielke:
This enables users to draw only parts of a texture by changing the S and T of the texture coordinates.

Should help drawing tilesets (can have up to 32 tiles using 8x8 on 16bits).

Also added example on the spritemap example ROM.

### DIFF
--- a/examples/spritemap/spritemap.c
+++ b/examples/spritemap/spritemap.c
@@ -100,6 +100,9 @@ int main(void)
                 /* Display a stationary sprite of adequate size to fit in TMEM */
                 rdp_draw_sprite( 0, 20, 50, MIRROR_DISABLED );
 
+                /* Draw only a pat of the plane texture */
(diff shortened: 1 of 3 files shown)


rasky:
Hi Stefan, we're working a bit on the rdp.c API as part of the new RSP queue engine. I'd like to keep this on hold until we finish that work as we will have a better picture of this module. There's nothing inherently wrong in this PR, it makes sense, but before adding more public APIs that we need to maintain, I'd rather understand the module better. Thanks for the patience.

rasky:
Closing this PR, as the [unstable branch](https://github.com/DragonMinded/libdragon/wiki/Unstable-branch) contains a different new library (`rdpq`). that supersedes rdp.c and includes many more functionalities, included this one. As we're readying to merge that, it doesn't make sense to accept new APIs for rdp.c

PR #241, opened 2022-01-11
Refactor timer callback API to support context parameter

meeq:
This approach uses an anonymous union of function pointers and a flag to select between callback signatures. Thanks for the advice @rasky


meeq:
@anacierdem @rasky This PR has incorporated all of the feedback from Discord. Please let me know if you have any additional suggestions. Thanks!

anacierdem:
LGTM. Adding a simple test for context version would be awesome though 😊

src/timer.c
```diff
@@ -250,6 +258,47 @@ timer_link_t *new_timer(int ticks, int flags, void (*callback)(int ovfl))
...
+ * @param[in] ticks
+ *            Number of ticks before the timer should fire
+ * @param[in] flags
+ *            Timer flags.  See #TF_ONE_SHOT, #TF_CONTINUOUS and #TF_DISABLED
+ * @param[in] callback
+ *            Callback function to call when the timer expires
+ * @param[in] ctx
+ * 			  User data to pass as an argument to callback
+ *
+ * @return A pointer to the timer structure created
```
rasky, line 294: I would modify the existing documentation to mention that using the `context` version is possibly preferred. Also please add `@see` tag to reference related functions (context vs non-context).

I wouldn't mark the non-context version as deprecated or using a too strong wording. As long as we point new users to the new versions, I think the older versions are fine to stay without scaring too much.

meeq:
> @anacierdem Adding a simple test for context version would be awesome though :blush:

 Done in 24e7c02ea6efa5904737119fa8cbd95f05ed3dea

tests/testrom.c
```diff
@@ -189,22 +189,23 @@ static const struct Testsuite
 	uint32_t duration;
 	uint32_t flags;
 } tests[] = {
-	TEST_FUNC(test_exception,              	   5, TEST_FLAGS_NO_BENCHMARK),
```
meeq, line 192: This block was a mix of tabs and spaces, which was causing visual alignment issues. I have made it consistent regardless of tab size by using a leading tab (no change), and spaces for field alignment padding.

rasky (approved):
Thanks! PR looks OK. I left a small comment about docs, and then as @anacierdem said we would need at least some basic coverage test. You don't need to test in details all functionalities, that is already done with the previous callback. Just write a test that calls each function once and makes sure it doesn't crash.

PR #242, opened 2022-01-13
Adding font loading to draw text functions

stefanmielke:
This PR adds the `graphics_set_font` function that loads sprite fonts that will be used on the `graphics_draw_character` function.

It also adds font width and height calculation on `graphics_draw_text`, removing the fixed values.


vieux:
Very cool! could you maybe add an example ? or explain how to generate a format similar to .h for other fonts?

stefanmielke:
> Very cool! could you maybe add an example ? or explain how to generate a format similar to .h for other fonts?

Added an example (customfont).

vieux:
Thanks @stefanmielke

Acceptable67:
I added this feature in my library just very recently. I would say good timing. Good work!

src/graphics.c
```diff
@@ -616,10 +632,25 @@ void graphics_fill_screen( display_context_t disp, uint32_t c )
...
 /**
- * @brief Draw a character to the screen using the built-in font
+ * @brief Set the current font. Should be set before using any of the draw function.
  *
- * @todo Need a set font mechanism.
- * @todo Need to look up the width and height from the font instead of assuming it is 8x8.
+ * @param[in] font
+ *        Sprite font to be used. Should be using a standard font sprite (containing every ASCII
+ * character) and with a transparent background, and be loaded on mksprite with width and height
+ * of 16.
```
rasky: I don't understand very well this sentence. What it means "width and height of 16"? Do you mean "hslices and vslices" instead? My suggestion would be to just leave "Sprite font to be used" here, and write a small description of how it should look like in the description of the function (paragraph after the brief).

What I guess is that it should be a sprite arranged in square fashion, with exactly 256 characters of equal size arranged in a 16x16 layout. And those characters must map exactly to ASCII 0-256. Is this correct? Maybe I wouldn't use the word "standard" in this context -- I'm not fully in touch with game world but I wouldn't say that there is an obvious standard for this; I've seen font images arranged in any layout.

My only worry about this is that ASCII only defines 128 characters. Characters 129-255 are actually undefined by ASCII, and in old 8-bit encodings are codepage-dependent. I think we should eventually work towards assuming UTF-8 here (which I think it can be compatible with N64 performance wise), so I'd like here at least not to make that goal more difficult. Would you agree in declaring that custom fonts for now are 16x8 grid instead, with 128 characters, and only support ASCII?

src/graphics.c
```diff
@@ -641,62 +672,72 @@ void graphics_draw_character( display_context_t disp, int x, int y, char ch )
 
     int depth = __bitdepth;
 
+    // setting default font if none was set previously
+    if( sprite_font.sprite == NULL )
+    {
+        sprite_font.sprite = (sprite_t *)(depth == 2 ? __font_data_16 : __font_data_32);
```
rasky: Shouldn't this configure the whole `sprite_font` structure for the system font? Including `font_width`, `font_height` and `tab_width`. Maybe you could just call `graphics_set_font` here.

rasky: Also, I think we should possibly export the system font to the public API now. So maybe having a system_font.c file with its header file. That's because otherwise the user has no way to explicitly select the system font after selecting something else.

src/graphics.c
```diff
@@ -616,10 +632,25 @@ void graphics_fill_screen( display_context_t disp, uint32_t c )
...
+ *        Sprite font to be used. Should be using a standard font sprite (containing every ASCII
+ * character) and with a transparent background, and be loaded on mksprite with width and height
+ * of 16.
+ * @param[in] tab_width
+ *        Width of each \\t used. Is multiplied by the width of each character.
+ */
+void graphics_set_font( sprite_t *font, int tab_width )
+{
+    sprite_font.sprite = font;
+    sprite_font.tab_width = tab_width;
```
rasky: I know that tab_width is currently hardcoded but I think tab_width is more of a concern for text layouting / drawing routines and not specifically bound to a font. The fact that I want to use tabulations of 8 or 4 has more to do with how to want align the text and less with the used font (and the current implementation is also broken as the tab is just a substitute for a sequence of spaces rather than aligning to a certain column, but I digress).

If you don't have a pressing need on this, I would suggest to leave it out from the public API for now. You can keep it in the structure if you want with value 5, and then we can devise a better way to configure it and use it.

src/graphics.c
```diff
@@ -616,10 +632,25 @@ void graphics_fill_screen( display_context_t disp, uint32_t c )
...
- * @todo Need a set font mechanism.
- * @todo Need to look up the width and height from the font instead of assuming it is 8x8.
+ * @param[in] font
+ *        Sprite font to be used. Should be using a standard font sprite (containing every ASCII
+ * character) and with a transparent background, and be loaded on mksprite with width and height
+ * of 16.
+ * @param[in] tab_width
+ *        Width of each \\t used. Is multiplied by the width of each character.
+ */
+void graphics_set_font( sprite_t *font, int tab_width )
```
rasky: Would you mind calling this `graphics_set_font_sprite`? I think eventually we want to a have a public `font_t` type (maybe even shared between graphics.c and rdp.c) that can be configured with a set of `font_*` functions, maybe has its own serialization functions and tools, and I'd like to keep the name `graphics_set_font` available for that scenario. This API would then remain as a quick fallback for backward compatibilty, so mentioning `sprite` in the name makes it clear that it's for the cases where you do have a `sprite_t` around.

rasky (changes requested):
Thanks you Stefan! I've left a few comments mostly around API design. I'm trying to be very conservative in general with API definitions, as we need to look in the future to not lock ourselves into a corner for further functionalities, as we would prefer to avoid API breakage as much as possible.

As a further comment not made inline: would you please change `console_init()` to also configure the system font? I think the console should use it, rather than inherit whatever custom font was last configured, especially as the console is "asynchronously" initialized for asserts and crashes.

src/graphics.c
stefanmielke: Yeah, I was trying to use this as an opportunity to explain to the user how to use it. But since I created an example it became just more confusing.

And honestly the size doesn't matter, as long as the sprite imported has enough slices to use (I even used a 16x8 on the example).

Should it be better to just point the user toward the example to see how to use it instead of trying to come up with a good explanation?

src/graphics.c
stefanmielke: Given we have different font according to the bits used, wouldn't it be better to just have a function (eg. `graphics_set_default_font`) to set the default font?

src/graphics.c
stefanmielke: I set that mostly for personal preference of a "normal" `tab_width` of 4 and not 5.

Do you have in mind how would we change this kind of setting in the future?

stefanmielke:
> As a further comment not made inline: would you please change `console_init()` to also configure the system font? I think the console should use it, rather than inherit whatever custom font was last configured, especially as the console is "asynchronously" initialized for asserts and crashes.

Did you mean something else rather than on `console_init()`, like on `__console_render`? That way we would always render using that font, even if we change the font right before rendering the console.

I guess if we do that we should state somewhere that you have to set the font every frame if you use the console (or else we'd have to keep the last font used somewhere and revert back to it after drawing the console).

rasky:
No I meant console_init. What I meant is that it could get called by an assertion failure while the user is drawing something via graphics, and may have changed the font.

After console_init, I don't expect graphics.c API to be used directly: the idea of the console is to just use printf.

src/graphics.c
rasky: Pointing to the example is always good, but I think also the example shows "one possible way". I think the documentation should state the exact requirements for the font bitmap, so that users can evaluate their options. I don't think a single example can capture all possibilities

src/graphics.c
rasky: Ok that's fine as well

src/graphics.c
rasky: Honestly I don't know. I think a more complex layouting API is due, as users would surely need to have their texts centered for instance, auto-split in multiple lines and so on. I think tabulation size is one possible options of this text layout function / engine. Also layouting is probably the same between graphics.c and rdp.c which is why it should probably be moved to a common module like a font.c.

An idea would be that this layout engine would render into a different buffer (another sprite) so that it can then be shown on screen as a simple textured quad. In this case, we would need some analysis to see whether its better to have a CPU-only renderer so that the quad can be I4/I8 for instance (after we support I4/I8 fonts) or it's better to try to use RDP and have this quad as RGB16. Rendering with RDP to a CI8 sprite is also an option which is worth investigating probably.

This is just a braindump of course, nothing concrete. Just to answer the question of why I think tabulation is more related to layout and less to the font.

src/graphics.c
```diff
@@ -641,62 +680,72 @@ void graphics_draw_character( display_context_t disp, int x, int y, char ch )
...
-            unsigned char c = __font_data[(ch * 8) + row];
+            const register int run = yp * sprite_font.sprite->width;
 
-            for( int col = 0; col < 8; col++ )
+            for( int xp = sx; xp < ex; xp++ )
             {
+                const char c = sp_data[xp + run];
                 if( trans )
                 {
                     if( c & 0x80 )
```
rasky: What does it mean `c & 0x80` in the context of a 16bpp or 32bpp font?

src/graphics.c
```diff
@@ -641,62 +680,72 @@ void graphics_draw_character( display_context_t disp, int x, int y, char ch )
...
-                        /* Only draw it if it is active */
-                        __set_pixel( buffer, x + col, y + row, f_color );
+                        __set_pixel( buffer, tx + xp, ty + yp, f_color );
                     }
                 }
                 else
                 {
-                    /* Display foreground or background depending on font data */
-                    __set_pixel( buffer, x + col, y + row, (c & 0x80) ? f_color : b_color );
+                    __set_pixel( buffer, tx + xp, ty + yp, (c & 0x80) ? f_color : b_color );
```
rasky: Same here

src/graphics.c
stefanmielke: Honestly I'm not even sure how that check works, but seemed to work correctly to grab the transparent pixel, so I left it there instead of trying to use a function to check.

rasky: In the previous code, the font was 1bpp. So the code was reading one byte, shifting it left 8 times, and checking each time the highest bit (0x80) to check whether it was 0 or 1 (transparent or solid). 

In the new code, where the font is 16bpp or 32bpp, I would expect it to check the alpha component. In case of 16bpp it is a single bit, while for 32bpp fonts it would be a full byte. Does that make sense to you?

stefanmielke:
@rasky I added the check used inside `__is_transparent`. Hoping this removes that block.

rasky (approved):
I think it's fine. Please rebase it for a last CI run before merge

PR #243, opened 2022-01-14
Fix `mixer_remove_event` entry identity matching

meeq:
This allows mixer events to use a single callback function with multiple contexts.

Previously, if the same callback was added multiple times with different contexts, attempting to remove a specific event context would only remove the first entry (for that callback) in the event list, which may not actually be the entry associated with the context provided.

This is especially problematic in C++ where:
1. GCC's nested function syntax is not available
2. C++ Lambdas require lifecycle management for asynchronous callbacks
3. `std::function` is not compatible with function pointers
4. Trampolines are the easiest way to interact with C-style APIs

This is not an API change, but is definitely a behavior change that may cause breakage, however I consider the old behavior to be a bug. The API implies that the identifying values of an event should be both the callback and the context, otherwise there would be no point in providing the context to `mixer_remove_event`.

@rasky

### DIFF
--- a/src/audio/mixer.c
+++ b/src/audio/mixer.c
@@ -608,7 +608,7 @@ void mixer_add_event(int64_t delay, MixerEvent cb, void *ctx) {
 
 void mixer_remove_event(MixerEvent cb, void *ctx) {
 	for (int i=0;i<Mixer.num_events;i++) {
-		if (Mixer.events[i].cb == cb) {
+		if (Mixer.events[i].cb == cb && Mixer.events[i].ctx == ctx) {
 			memmove(&Mixer.events[i], &Mixer.events[i+1], sizeof(mixer_event_t) * (Mixer.num_events-i-1));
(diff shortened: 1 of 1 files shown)


meeq:
The only existing usage of this function is unaffected by this change: https://github.com/DragonMinded/libdragon/blob/63ee2dceb5e0cc25f93e7c5c8709372eff5a1240/src/audio/xm64.c#L237

My hope is that the mixer still too new for a change like this to have significant impact to the community, and that this change has such overwhelming benefits that it's worth adopting.

anacierdem:
This change looks correct and necessary to me as well. Though might be nice to hear @rasky s opinion as well. Another idea: we can remove all listeners if the provided ctx is NULL

rasky:
Yes I agree it is just a bug. Also the way it is used in xm64.c implies that the expected behavior is the one fixed in this PR. Thanks!

PR #244, opened 2022-01-18
mixer: add mixer_throttle

rasky:
This function allows to throttle the mixer to a specific pace, to help
implementing correct syncing with a video.

Normally, once the mixer is initiated and assuming mixer_poll is called
frequently enough, the audio will playback uninterrupted, irrespective of
any slow down in the main loop. This is the expected behavior for background
music for instance, but it does not work for video players or cut-scenes in
which the music must be perfectly synchronized with the video. If the video
happens to slowdown, the music will desynchronize.

mixer_throttle sets a budget of samples that the mixer is allowed to
generate. Every time the function is called, the specified number of samples
is added to the budget. Every time the mixer playbacks the channel, the
budget is decreased. If the budget reaches zero, the mixer will automatically
pause playback until the budget is increased again, possibly creating
audio cracks.

### DIFF
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -198,7 +198,7 @@ void mixer_ch_play(int ch, waveform_t *wave);
  * By default, the frequency is the one required by the waveform associated
  * to the channel, but this function allows to override.
  * 
- * This function must be called after mixer_ch_play, as otherwise the
+ * This function must be called after #mixer_ch_play, as otherwise the
  * frequency is reset to the default of the waveform.
  * 
  * @param[in]   ch              Channel index
@@ -213,7 +213,7 @@ void mixer_ch_set_freq(int ch, float frequency);
  * The position must be specified in number of samples (not bytes). Fractional
  * values account for accurate resampling position.
  *
- * This function must be called after mixer_ch_play, as otherwise the
+ * This function must be called after #mixer_ch_play, as otherwise the
  * position is reset to the beginning of the waveform.
  * 
  * @param[in]   ch              Channel index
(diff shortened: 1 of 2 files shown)


include/mixer.h
```diff
@@ -300,6 +348,10 @@ void mixer_ch_set_limits(int ch, int max_bits, float max_frequency, int max_buf_
  * Since the N64 AI can only be fed with an even number of samples, mixer_poll
  * does not accept odd numbers.
  * 
+ * This function will respected throttling, if configured via #mixer_throttole.
```
meeq: Typo: `mixer_throttole` -> `mixer_throttle`

include/mixer.h
```diff
@@ -283,6 +283,54 @@ bool mixer_ch_playing(int ch);
...
+/**
+ * @brief Unthrottle the mixer
+ * 
+ * Switch back the mixer to the default unthrottled status, after some calls to
+ * #mixer_throttle.
+ * 
+ * After calling #mixer_unthrottle, the mixer will no longer be limited and
+ * will produce all the samples requested via #mixer_poll.
+ * 
+ * @see #mixter_throttle
```
meeq: Typo: `mixter_throttle` -> `mixer_throttle`

rasky:
I'm experimenting a bit with audio/video syncing. This PR works correctly in allowing a videoplayer to keep sync with audio even when it slows down. I would say in general the pull-based / callback-based approach of the mixer is not the best for a videoplayer; in a muxed video, the videoplayer would probably want to decode and push audio into the mixer at its own pace, rather than setting up for the being called back at "random" moments for audio decoding. OTOH, this works good if video and audio are in separate files, which also means that it makes it easier to use the same decoder for both videoplayers and normal background music.

I'll keep this PR in draft for now.

rasky:
This was merged to the preview branch long time ago, it is used in the videoplayer example. There's no need to keep this PR around.

PR #245, opened 2022-01-18
debug: improve assert reporting in low memory conditions

rasky:
Currently, assertions initialize the console and display the message
there. This is not a reliable solution in a low-memory scenario:
unfortunately, initializing the console resets the display and causes
memory allocations to create the new buffers, which might fail,
causing in turn an infinite crash/assert loop that prevents the original
assertion to ever being displayed.

The correct solution is possibly to have an "emergency console" that
somehow does not allocate. But this is hairy to implement in the
current architecture.

This PR implements a small stopgap: it forces assert to first prints
the assert on stderr (aka: USB / ISViewer), and only after initialize
the console. At least, we get to see the assertion on the debug log
before the console freezes, which is better than now.

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -494,11 +494,37 @@ void debug_close_sdfs(void)
 
 void debug_assert_func_f(const char *file, int line, const char *func, const char *failedexpr, const char *msg, ...)
 {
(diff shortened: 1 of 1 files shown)


anacierdem (approved):
LGTM

PR #246, opened 2022-01-30
Improve example Makefiles

anacierdem:
- Add missing builds for eepromfs and rtc examples
- Add `$(SOURCE_DIR)` prefix for `testrom_emu.o` recipe in test Makefile.
Without this the paths for the two sources become inconsistent when building
with a `SOURCE_DIR` override. Providing it is useful when using tools that parse
the build output. You can provide `SOURCE_DIR=../tests` to make such that the
build output looks like;
`[CC] ../tests/testrom.c`
This makes it possible to discriminate which error output is coming from which
path. Without consistent paths, there will be a second `[CC] testrom.c`.
- To be able to benefit from the same path consistency, changed the input list
to object files for all examples. When provided like `$(BUILD_DIR)/object.o`, it
ensures they all use the same `SOURCE_DIR` prefix and makes the Makefile a
little less verbose.
- We dont need to manually create a list of `.d` files. Start using a wildcard
instead.


Acceptable67:
Nice to see the new examples added.

PR #248, opened 2022-02-05
rsp.inc: improve support for opcodes vrndn/vrndp/vmacq

rasky:
These obscure opcodes are tailored for MPEG1 inverse quantization. They
have a weird syntax: vrndn/vrndp use one of the register fields as
boolean flag, and vmacq (which, despite the name, as nothing to do with
multiplications) only works on the destination register.

This is technically a breaking change but I'll eat my hat if somebody
is already using these opcodes... anyway the fix is trivial.


snacchus (approved):
Looks good!

PR #249, opened 2022-02-06
Two small changes in wav64

rasky:
* Add missing `wav64_play` (was documented in header file, but non existing)
 * Align `wav64_open` to the new convention of requesting filesystem-qualified filenames. Keep the bare filenames working for backward compatibility


meeq:
Looks good. I couldn't find a GCC warning for "function prototype declared without a definition", but it seems like there should be one.

(I was able to find the opposite: "global function defined without a prototype" [`-Wmissing-prototypes`], but that's not helpful in this scenario)

PR #250, opened 2022-02-07
audio: use the new malloc_uncached API

rasky:
The internal buffers are meant to be used with uncached accesses as
they are primarily consumed via DMA (AI or RSP).

This is a slight API change as the recent audio_write_begin() will
now return an uncached pointer, but hopefully it will not change much
for most use cases.

@meeq @ryzee119 I would appreciate if you could test this PR in your games. It should not break audio :)

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -170,13 +170,13 @@ static void audio_callback()
         }
 
         if (_fill_buffer_callback) {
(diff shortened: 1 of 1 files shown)


meeq:
Tested with FlappyBird, no issues detected.

Ryzee119:
Tested on Cosmo64 and Commander Keen and zzgunner. All work with no issues

PR #253, opened 2022-02-17
Add a high performance, general purpose RSP command queue engine

snacchus:
This PR introduces the rspq library (short for "RSP command queue"), which provides the basic infrastructure to allow a very efficient use of the RSP coprocessor. On the CPU side, it implements an API to enqueue "commands" to be executed by RSP into a ring buffer, that is concurrently consumed by RSP in background. On the RSP side, it provides the core loop that reads and execute the queue prepared by the CPU, and an infrastructure to write "RSP overlays", that is libraries that plug upon the RSP command queue to perform actual RSP jobs (eg: 3D graphics, audio, etc.).

The library is extremely efficient. It is designed for very high throughput and low latency, as the RSP pulls by the queue concurrently as the CPU  fills it. Through some complex synchronization paradigms, both CPU and RSP run fully lockless, that is never need to explicitly synchronize with each other (unless requested by the user). The CPU can keep filling the queue and must only wait for RSP in case the queue becomes full; on the other side, the RSP can keep processing the queue without ever talking to the CPU.

The library has been designed to be able to enqueue thousands of RSP commands per frame without its overhead to be measurable, which should be more than enough for most use cases.

### Commands

Each command in the queue is made by one or more 32-bit words (up to 15 currently). The MSB of the first word is the command ID. The higher 4 bits are called the "overlay ID" and identify the overlay that is able to execute the command; the lower 4 bits are the command index, which identify the command within the overlay. For instance, command ID 0x37 is command index 7 in overlay 3.

As the RSP executes the queue, it will parse the command ID and dispatch it for execution. When required, the RSP will automatically load the RSP overlay needed to execute a command. In the previous example, the RSP will load into IMEM/DMEM overlay 3 (unless it was already loaded) and then dispatch command 7 to it.

### Higher-level libraries and overlays

Higher-level libraries that come with their RSP ucode can be designed to use the RSP command queue to efficiently coexist with all other RSP libraries provided by libdragon. In fact, by using the overlay mechanism, each library can obtain its own overlay ID, and enqueue commands to be executed by the RSP through the same unique queue. Overlay IDs are allocated dynamically by rspq in registration order, to avoid conflicts between libraries.

End-users can then use all these libraries at the same time, without having to arrange for complex RSP synchronization, asynchronous execution or plan for efficient context switching. In fact, they don't even need to be aware that the libraries are using the RSP. Through the unified command queue, the RSP can be used efficiently and effortlessly without idle time, nor wasting CPU cycles waiting for completion of a task before switching to another one.

Higher-level libraries that are designed to use the RSP command queue must:

  * Call `rspq_init` at initialization. The function can be called multiple times by different libraries, with no side-effect.
  * Call `rspq_overlay_register` to register a `rsp_ucode_t` as RSP command queue overlay, obtaining an overlay ID to use.
  * Provide higher-level APIs that, when required, call `rspq_write` and `rspq_flush` to enqueue commands for the RSP. For instance, a matrix library might provide a "matrix_mult" function that internally calls `rspq_write` to enqueue a command for the RSP to perform the calculation.

To be compatible with the queue engine, ucodes must simply include `rsp_queue.inc` at the top of the file and define a header and a command table at the beginning of their data section using `RSPQ_BeginOverlayHeader`, `RSPQ_DefineCommand` and `RSPQ_EndOverlayHeader`. An overlay ucode doesn't have a single entry point: it exposes multiple functions bound to different commands, that will be called by the queue engine when the commands are enqueued. See `tests/rsp_test.S` for an example.

### Blocks

A block (`rspq_block_t`) is a prerecorded sequence of RSP commands that can be played back. Blocks can be created via `rspq_block_begin` / `rspq_block_end`, and then executed by `rspq_block_run`. It is also possible to do nested calls (a block can call another block), up to 8 levels deep.

A block is very efficient to run because it is played back by the RSP itself. The CPU just enqueues a single command that "calls" the block. It is thus much faster than enqueuing the same commands every frame.

Notice that this library does not support static (compile-time) blocks. Blocks must always be created at runtime once (eg: at init time) before being used.

### Syncpoints

The RSP command queue is designed to be fully lockless, but sometimes it is required to know when the RSP has actually executed an enqueued command or not (eg: to use its result). To do so, this library offers a synchronization primitive called "syncpoint" (`rspq_syncpoint_t`). A syncpoint can be created via `rspq_syncpoint` and records the current writing position in the queue. It is then possible to call `rspq_check_syncpoint` to check whether the RSP has reached that position, or `rspq_wait_syncpoint` to wait for the RSP to reach that position.

Syncpoints are implemented using RSP interrupts, so their overhead is small but still measurable. They should not be abused.

### High-priority queue

This library offers a mechanism to preempt the execution of RSP to give priority to very urgent tasks: the high-priority queue. Since the moment a high-priority queue is created via `rspq_highpri_begin`, the RSP immediately suspends execution of the command queue, and switches to the high-priority queue, waiting for commands. All commands added via standard APIs (`rspq_write`) are then directed to the high-priority queue, until `rspq_highpri_end` is called. Once the RSP has finished executing all the commands enqueued in the high-priority queue, it resumes execution of the standard queue.

If required, it is possible to call `rspq_highpri_sync` to wait for the high-priority queue to be fully executed.

### Final notes
For an explanation of implementation details, see documentation in `src/rspq.c`.

In addition to the core library, this PR also ports the existing RSP mixer library to be compatible with rspq.

The implementation of this library was a combined effort between @rasky and me. I prototyped the command queue, implemented overlay support and ported the mixer library. Rasky added the blocks and highpri queue, drove the API design and wrote most of the absolutely incredible rspq ucode, which is both extremely fast *and* as efficient on instruction size as humanly possible, leaving plenty of IMEM for overlays.


include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * 
+ * Normally, end-users will not need to manually enqueue commands in the RSP
+ * queue: they should only call higher-level APIs which internally do that.
+ * 
+ * ## Blocks
+ * 
+ * A block (#rspq_block_t) is a prerecorded sequence of RSP commands that can
+ * be played back. Blocks can be created via #rspq_block_begin / #rspq_block_end,
+ * and then executed by #rspq_block_run. It is also possible to do nested
+ * calls (a block can call another block), up to 8 levels deep.
```
anacierdem, line 82: Is the minimum depth of 8 is a requirement for opengl?

anacierdem, line 82: Is the minimum depth of 8 is a requirement for opengl?

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * enqueued in the standard queue. This can be useful for running tasks that
+ * require immediate execution, like for instance audio processing.
+ * 
+ * If required, it is possible to call #rspq_highpri_sync to wait for the
+ * high-priority queue to be fully executed.
+ * 
+ * Notice that the RSP cannot be fully preempted, so switching to the high-priority
+ * queue can only happen after a command has finished execution (before starting
+ * the following one). This can have an effect on latency if a single command
+ * has a very long execution time; RSP overlays should in general prefer
```
anacierdem, line 134: In the future, we may provide an actual number here as we get more experience.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * that position or not.
+ * 
+ * To create a syncpoint, use #rspq_syncpoint that returns a syncpoint that
+ * references the current position. Call #rspq_check_syncpoint or #rspq_wait_syncpoint
+ * to respectively do a single check or block waiting for the syncpoint to be
+ * reached by RSP.
+ * 
+ * Syncpoints are implemented using interrupts, so they have a light but non
+ * trivial overhead. Do not abuse them. For instance, it is reasonable to use
+ * tens of syncpoints per frame, but not hundreds or thousands of them.
```
anacierdem, line 184: Then it is a good idea to document libdragon's ucode's syncpoints in their public docs.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * 
+ * The function returns the overlay ID, which is the ID to use to enqueue
+ * commands for this overlay. The overlay ID must be passed to #rspq_write
+ * when adding new commands. rspq allows up to 16 overlays to be registered
+ * simultaneously, as the overlay ID occupies the top 4 bits of each command.
+ * 
+ * Notice that it is possible to occupy multiple consecutive IDs with the
+ * same overlay in case the overlay exposes more than 16 commands. For instance,
+ * registering an overlay that handles up to 32 commands could return ID 0x6, 
+ * meaning that the whole range of command IDs 0x60-0x7F would be assigned to it.
```
anacierdem: I didn't get the logic here just by reading the description.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ *             
+ * @param      overlay_ucode  The overlay to register
+ *
+ * @return     The overlay ID that has been assigned to the overlay. Note that
+ *             this value will be preshifted by 28 (eg: 0x60000000 for ID 6),
+ *             as this is the expected format used by #rspq_write.
```
anacierdem, line 240: Very nice touch!

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+
+/**
+ * @brief Register an overlay into the RSP queue engine assigning a static ID to it
+ * 
+ * This function works similar to #rspq_overlay_register, except it will
+ * attempt to assign the specified ID to the overlay instead of automatically
+ * choosing one. Note that if the ID (or a consecutive IDs) is already used
+ * by another overlay, this function will assert, so careful usage is advised.
+ * 
+ * Assigning a static ID can mostly be useful for debugging purposes.
```
anacierdem, line 252: Or might be a tool for building static compiled command queues for specific requirements for a project. I presume the libdragon ucode libraries will do `rspq_overlay_register` when they are initialized, so it will be possible to allocate an id before initializing them?
Not because it is important, just trying to clarify.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * 
+ * @see #rspq_overlay_register
+ */
+void rspq_overlay_register_static(rsp_ucode_t *overlay_ucode, uint32_t overlay_id);
+
+/**
+ * @brief Unregister a ucode overlay from the RSP queue engine.
+ * 
+ * This function removes an overlay that has previously been registered
+ * with #rspq_overlay_register from the queue engine. After calling this
```
anacierdem: nit: `rspq_overlay_register` & `rspq_overlay_register_static`

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ */
+void rspq_overlay_register_static(rsp_ucode_t *overlay_ucode, uint32_t overlay_id);
+
+/**
+ * @brief Unregister a ucode overlay from the RSP queue engine.
+ * 
+ * This function removes an overlay that has previously been registered
+ * with #rspq_overlay_register from the queue engine. After calling this 
+ * function, the specified overlay ID (and consecutive IDs in case the overlay
+ * occupied multpiple slots) is no longer valid and must not be used to write
```
anacierdem: typo: `multiple`. How do we know the overlay slot length? It is not still clear just reading the header.

doxygen-public.conf
```diff
@@ -905,7 +905,7 @@ RECURSIVE              = NO
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                =
+EXCLUDE                = ./src/audio ./src/fatfs
```
anacierdem: Hmm why did we need to exclude `src/audio` I wonder? Maybe we can just open an issue to remember this.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * @brief Make sure that RSP starts executing up to the last written command.
+ * 
+ * RSP processes the command queue asynchronously as it is being written.
+ * If it catches up with the CPU, it halts itself and waits for the CPU to
+ * notify that more commands are available. On the contrary, if the RSP lags
+ * behind it might keep executing commands as they are written without ever
+ * sleeping. So in general, at any given moment the RSP could be crunching
+ * commands or sleeping waiting to be notified that more commands are available.
+ * 
+ * This means that writing a command via #rspq_write is not enough to make sure
```
anacierdem, line 402: I think this only applies to the case where the RSP is halted before we atomically add new commands right? In any case the number of written commands will be unknown and require a flush as we don't have another way of knowing RSP state. Just asking to clarify.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * 		}
+ * 		
+ * 		// After enqueuing multiple commands, it is sufficient
+ * 		// to call rspq_flush once to make sure the RSP runs them (in case
+ * 		// it was idling).
+ * 		rspq_flush();
+ * @endcode
+ *
+ * @note This is an experimental API. In the future, it might become 
+ *       a no-op, and flushing could happen automatically at every #rspq_write.
```
anacierdem, line 438: This makes it clearer! I like the approach here.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+/**
+ * @brief Create a syncpoint in the queue.
+ * 
+ * This function creates a new "syncpoint" referencing the current position
+ * in the queue. It is possible to later check when the syncpoint
+ * is reached by the RSP via #rspq_check_syncpoint and #rspq_wait_syncpoint.
+ *
+ * @return     ID of the just-created syncpoint.
+ * 
+ * @note It is not possible to create a syncpoint within a block
```
anacierdem: We may also add the reason why here?

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * order relative to other commands in the queue.
+ * 
+ * Blocks can call other blocks. For instance, if a block A has been fully
+ * created, it is possible to call `rspq_block_run(A)` at any point during the
+ * creation of a second block B; this means that B will contain the special
+ * command that will call A.
+ *
+ * @param block The block that must be run
+ * 
+ * @note The maximum depth of nested block calls is 8.
```
anacierdem, line 567: In the future, we may statically analyze before compiling the code.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+ * queue (after #rspq_highpri_end closes it), it resumes processing the normal
+ * queue from the exact point that was left.
+ * 
+ * The goal of the high-priority queue is to either schedule latency-sensitive
+ * commands like audio processing, or to schedule immediate RSP calculations
+ * that should be performed right away, just like they were preempting what
+ * the RSP is currently doing.
+ * 
+ * It is possible to create multiple high-priority queues by calling
+ * #rspq_highpri_begin / #rspq_highpri_end multiples time with short
```
anacierdem: typo: multiples

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+
+/**
+ * @brief Finish building the high-priority queue and close it.
+ * 
+ * This function terminates and closes the high-priority queue. After this
+ * command is called, all following commands will be added to the normal queue.
+ * 
+ * Notice that the RSP does not wait for this function to be called: it will
+ * start running the high-priority queue as soon as possible, even while it is
+ * being built.
```
anacierdem, line 631: Does this also mean we don't need `rspq_flush` when inside a `rspq_highpri_begin`?
Something to keep in mind regarding multithreading: multiple threads might call this before one ended the high prio queue. I know we don't support re-entrancy for rspq ATM but it looked like a little trickier than other cases here.

A question: will this cause a bottleneck b/c CPU is unable to push enough commands to the high priority queue? OTOH I don't think it will ever be an issue because each command will take enough time on RSP that CPU will probably never lag behind.

include/rspq.h
```diff
@@ -0,0 +1,704 @@
...
+/**
+ * @brief Wait for the RSP to finish processing all high-priority queues.
+ * 
+ * This function will spin-lock waiting for the RSP to finish processing
+ * all high-priority queues. It is meant for debugging purposes or for situations
+ * in which the high-priority queue is known to be very short and fast to run,
+ * so that the overhead of a syncpoint would be too high.
+ * 
+ * For longer/slower high-priority queues, it is advisable to use a #rspq_syncpoint_t
+ * to synchronize (though it has a higher overhead).
```
anacierdem: So why it is advisable to use it instead? If we only want to wait for high prio to end is it better to use a syncpoint as last command or `rspq_highpri_sync`?

include/rsp_dma.inc
```diff
@@ -120,6 +120,7 @@ DMAInEnd:
     andi $1, s0, 7
     add s4, $1
     andi t2, 0xF   # isolate DMA wait flags
+SpStatusWait:
```
anacierdem, line 123: Why not use `DMAWaitLoop`?

include/rsp_queue.inc
```diff
@@ -0,0 +1,447 @@
...
+# Globally reserved registers
+#define rspq_dmem_buf_ptr gp
+
+# Can be used with l* instructions to get contents of the current command at the specified offset.
+# The total command size needs to be specified as well.
+#define CMD_ADDR(offset, cmdsize) (%lo(RSPQ_DMEM_BUFFER) + (offset) - (cmdsize)) (rspq_dmem_buf_ptr)
+
+#define OVERLAY_HEADER_SIZE 0x8
+
+.macro RSPQ_BeginOverlayHeader stateStart, stateEnd
```
anacierdem: It would be great to have this documented as well. (Sorry if I was unable to find it)

src/audio/mixer.c
```diff
@@ -562,22 +580,25 @@ void mixer_exec(int32_t *out, int num_samples) {
...
-	SP_DMEM[1] = (num_samples << 16) | Mixer.num_channels;
-	SP_DMEM[2] = (uint32_t)out;
-	SP_DMEM[3] = (uint32_t)Mixer.ucode_state;
-
 	uint32_t t0 = TICKS_READ();
-	rsp_run();
+
+	rspq_highpri_begin();
+	rspq_write(__mixer_overlay_id, 0,
+		(((uint32_t)MIXER_FX16(Mixer.vol)) & 0xFFFF),
```
anacierdem, line 637: How do we make sure MSB is all zeros here? (now knowing mixer ucode yet)

anacierdem (commented):
This is really well implemented and documented! Loved it!

I was only able to review the headers files for now (and a few other places here and there) and tried to write down whenever I have a question in mind. I will continue with the implementation but that's something I'm not very proficient. Still eventually I will go over the whole thing but I'm ok to merge once someone goes over my current comments if I lag behind.

This is huge! Congrats to both of you @rasky & @snacchus! Can't tell how happy I am having you building upon frankenGAS :)

The only downside is we have more to do on the pre-emptive multithreading now :) One step at a time though.

include/rspq.h
rasky: No it's just arbitrary. We can grow it (at compile time) if we feel it's required, wasting some DMEM (4 bytes per level).

include/rspq.h
rasky: yes that's the idea. You initialized your own libraries first to reserve the fixed IDs. Or, knowing that rspq allocates in growing order, you can simply reserve as static the last IDs (0xF, 0xE, etc.) so that will work independently from libdragon's internal libraries.

doxygen-public.conf
rasky: When I revamped doxygen, I didn't realize that the option `RECURSIVE=OFF` meant that `src/audio` wasn't being processed by doxygen. The mixer docs were still be present because the header files are in the `include` dir, but the C files weren't being processed. In this PR, I'm activating `RECURSIVE=ON` to document `src/rspq/*.c`, but that causes a CI failure because `src/audio` is missing some docs now. I need to add missing docs there as well.

include/rspq.h
rasky: yes correct.

include/rspq.h
rasky: The reason is that a syncpoint is a one-time event. Adding it within a block means that the same syncpoint could be triggered multiple times, which in turns makes it unclear how to use the API (what should `rspq_check_syncpoint` returns?). We can add some text explaining this.

include/rspq.h
rasky: I'm not sure what your comment means wrt the quoted line?

include/rspq.h
snacchus: Also note that for the purpose of forwarding RDP commands, IDs 0x2 and 0x3 are "soft reserved", which means when registering overlays, IDs 0x2 and 0x3 will be assigned last until no others are available. The upcoming "gfx" overlay will statically register itself with those IDs. But if gfx is not actually needed, those IDs are still available for other custom overlays.

include/rspq.h
rasky: > Does this also mean we don't need rspq_flush when inside a rspq_highpri_begin?

We still need `rspq_flush` in highpri. What the comment is trying to explain is that the act of calling `rspq_highpri_begin` as an immediate effect on RSP: ASAP, it will switch to the highpri queue and wait there for commands. To add command, you need to do write+flush as usual, but I wanted to clarify that rspq_highpri_begin will immediately halt the RSP, so the client shouldn't call it much in advance wrt the first write (or they can do it, if they want to absolutely reduce latency and make sure that the RSP is idle when the first write+flush arrives).

> multiple threads might call this before one ended the high prio queue.

yeah in general as you notices rspq is not reentrant. For highpri or even block creation which exposes a stateful API, I think the best solution is to keep those state as thread-local. Eg. a thread might be preempted while building a block; if another thread writes to the queue, it will writes to the normal queue, not to the block being created by the first thread. Similarly, if a thread is building a highpri queue and is preempted, the second thread will write to the lowpri queue (and those commands will stay there pending, waiting for RSP to go back to lowpri mode).

rasky: > A question: will this cause a bottleneck b/c CPU is unable to push enough commands to the high priority queue? OTOH I don't think it will ever be an issue because each command will take enough time on RSP that CPU will probably never lag behind.

Yeah I mean it could cause a small delay on RSP, but it's the price to pay if you have high priority aka low latency. I wouldn't say a bottleneck in general. If the CPU needs to write 1 highpri command, then do a long calculation, then write another highpri command, it can safely turn off highpri mode after the first command and turn it back again when the second command is ready to be written. What happens in this case is that if the RSP was still executing the first command, it will then smoothly go to the second command before going back to lowpri queue. Otherwise, if the RSP had already finished the first command and get back to lowpri, it will have to switch again to highpri. So it's up to the caller to decide whether use a single longer highpri mode (less latency, but possible RSP idle time) or two smaller highpri modes (potentially more latency, but no RSP idle time).

include/rspq.h
snacchus: instead of detecting nesting depth overflow at runtime and crashing, build some custom analyzer that detects at build time?

include/rspq.h
snacchus: The first byte of each command contains the overlay ID and the command ID. The 4 MSB are overlay ID, 4 LSB are command ID. This means that there are 16 unique overlay IDs, and per overlay ID there is a maximum of 16 commands. To support overlays that have more than 16 commands, it is possible for an overlay to have multiple IDs assigned to it. `rspq_register_overlay` handles this automatically by examining the ucode itself and determining the number of commands that are defined by it. The number of commands (rounded up to multiple of 16) divided by 16 will give the required number of ID "slots" it needs to occupy for all commands to be callable.
So for example if an overlay with 24 commands is registered and assigned ID 0x50, It is possible to call the twenty-first command of this overlay by using `0x50 + 0x14 = 0x64`. The next overlay would then be registered at 0x70 in this case.
I'll expand the description a bit to make this clearer.

include/rspq.h
rasky: rspq_highpri_sync will spin loop. If you know it's going to take a long time, it might be better to add a syncpoint as last command in the highpri queue, go doing something else, and later check whether it has finished or not. Obviously you could instead call rspq_highpri_sync later -- if you know that no other highpri stuff has happened.

include/rsp_dma.inc
rasky: DMAWaitLoop is a misnomer, because the code just waits for a specific status bit to be set. In the context of rsp_dma.inc it is used to wait for DMA to be not full or idle, but we realized we could reuse the code for a generic "wait for a bit in SP status". Since we can't rename an existing label in a shared assembly library (as it's basically public API), we added a new label with a name that it's cleared in the context of rspq. If you read the code in rspq that uses it, it would have been highly confusing to see a call to DMAWaitLoop as that code has nothing to do with DMA.

src/audio/mixer.c
rasky: Mixer.vol is supposed to be in the 0..1 range. If it can't go beyond that it's a bug, but I wouldn't clamp it here -- I would do that where it can wrongly assume that out-of-bound value. This is the inner execution loop where the global state is assumed to be consistent and valid.

include/rspq.h
snacchus: Actually is it even possible to use syncpoints in highpri? Because they might then be triggered out of sequence?

rasky: oh sorry that's right, you can't do that and is probably also covered by an assert. So this comment should be removed.

include/rspq.h
snacchus: I rewrote those docs in an attempt to make the logic more clear. What do you think?

include/rspq.h
snacchus: fixed

include/rspq.h
snacchus: fixed. Also, as mentioned in another comment, I rewrote some docs to make it more clear.

doxygen-public.conf
snacchus: Can this be resolved?

include/rspq.h
snacchus: Added some docs to explain

include/rspq.h
snacchus: fixed

include/rspq.h
snacchus: It wasn't covered by an assert so far, so I added one.

include/rsp_queue.inc
snacchus: I realized that rsp_queue.inc and writing overlays wasn't documented very well in general, so I added a bunch of docs. Also I changed the API a little bit to make it less error prone and a little bit nicer as well. What do you think?

include/rspq.h
snacchus: So far, syncpoints are not used internally anywhere. The mixer used to, but since it always pushes onto the high-priority queue, it uses highpri_sync instead. I feel like that is an implementation detail and doesn't really need to be documented anywhere.

snacchus:
As mentioned in a reply above, I revised some of the API in rsp_queue.inc to make writing overlays a little bit easier and less error prone. The location of the saved state now doesn't need to be explicitly passed to the `RSPQ_BeginOverlayHeader` macro. Instead, you just place the data that needs to be persisted between a pair of `RSPQ_BeginSavedState` and `RSPQ_EndSavedState`. The overlay registration now also asserts whether the size of the saved state is bigger than zero, which would cause undefined behavior before. If no data needs to be persisted, then `RSPQ_EmptySavedState` can be used instead.

I also added a bunch of documentation to the macros in rsp_queue.inc and a short guide on how to write overlays.

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+RSPQ_LOG:                    .ds.l 16
+RSPQ_LOG_END:                .long 0xFFFFFFFF
+#endif
+
+    .align 3
+RSPQ_DMEM_BUFFER:            .ds.b RSPQ_DMEM_BUFFER_SIZE
+
+
+    .align 4
+# Overlay data will be loaded at this address
```
anacierdem, line 225: Overlay code?

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+    # Check if new commands were added in the display list (SIG_MORE)
+    mfc0 t0, COP0_SP_STATUS
+    andi t0, SP_STATUS_SIG_MORE
+    bnez t0, wakeup
+    li t0, SP_WSTATUS_CLEAR_SIG_MORE
+
+    # No new commands yet, go to sleep
+    break
+wakeup:
+    mtc0 t0, COP0_SP_STATUS
```
anacierdem, line 245: Why do we write back to the status register here?

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+    # starting from the current position.
+    # Notice that we use "bge" instead of "bgt" so we actually refetch the buffer
+    # also if the current command ends exactly at buffer end; this is slighly
+    # wasteful but saves us a few instructions (that would be required to check
+    # whether we are then trying to load a command outside of the buffer).
+    addu t0, rspq_dmem_buf_ptr, cmd_size
+    bge t0, RSPQ_DMEM_BUFFER_SIZE, rspq_fetch_buffer
+
+    # Load second to fourth command words (might be garbage, but will never be read in that case)
+    # This saves some instructions in all overlays that use more than 4 bytes per command.
```
anacierdem, line 375: How?

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+
+    .globl _start
+_start:
+    li rspq_dmem_buf_ptr, 0
+
+    .func RSPQCmd_WaitNewInput
+RSPQCmd_WaitNewInput:
+    # Check if new commands were added in the display list (SIG_MORE)
+    mfc0 t0, COP0_SP_STATUS
+    andi t0, SP_STATUS_SIG_MORE
```
anacierdem, line 238: In the worst case, the CPU sets the signal at this instance and that would cause an additional sleep right?

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+    # a0: command opcode + call slot in DMEM to recover
+    j rspq_fetch_buffer_with_ptr
+    lw s0, %lo(RSPQ_POINTER_STACK)(a0)
+    .endfunc
+
+    #############################################################
+    # RSPQCmd_TestWriteStatus
+    #
+    # Like RSPQCmd_WriteStatus, writes COP0 SP status register with a specified value,
+    # but first waits until the SP status AND'ed with the specified bitmask is zero.
```
anacierdem, line 504: I presume this register is atomically written to by the CPU. Is it the the same mechanism as the 32bit RDRAM writes?

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+
+########################################################
+# RSPQ_BeginOverlayHeader
+# 
+# Begins the definition of an overlay header. Must be
+# followed by at least one call to RSPQ_DefineCommand
+# and closed with RSPQ_EndOverlayHeader. Only calls
+# to RSPQ_DefineCommand are allowed inside the header definition.
+########################################################
+.macro RSPQ_BeginOverlayHeader
```
anacierdem, line 47: Might be nice to cross reference `rspq_overlay_header_t` here.

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+    uint32_t nesting_level;     ///< Nesting level of the block
+    uint32_t cmds[];            ///< Block contents (commands)
+} rspq_block_t;
+
+/** @brief A RSPQ overlay ucode. This is similar to rsp_ucode_t, but is used
+ * internally to managed it as a RSPQ overlay */
+typedef struct rspq_overlay_t {
+    uint32_t code;              ///< Address of the overlay code in RDRAM
+    uint32_t data;              ///< Address of the overlay data in RDRAM
+    uint32_t state;             ///< Address of the overlay state in RDRAM (within data)
```
anacierdem, line 347: doensn't the CPU know the state size? Might be usfeul to have bound checking in debug mode maybe...

include/rsp_queue.inc
```diff
@@ -0,0 +1,566 @@
...
+# overlay does not need any persistent state.
+#
+# It is only possible to use either this or 
+# RSPQ_BeginSavedState/RSPQ_EndSavedState, but not both.
+# Do not use RSPQ_BeginSavedState and RSPQ_EndSavedState
+# without any data directives between them.
+########################################################
+.macro RSPQ_EmptySavedState
+    RSPQ_BeginSavedState
+    .quad 0
```
anacierdem, line 105: Why do we need to have at least this amount of data?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+    rspq_overlay_t overlay_descriptors[RSPQ_MAX_OVERLAY_COUNT];
+} rspq_overlay_tables_t;
+
+/**
+ * @brief RSP Queue data in DMEM.
+ * 
+ * This structure is defined by rsp_queue.S, and represents the
+ * top portion of DMEM.
+ */
+typedef struct rsp_queue_s {
```
anacierdem, line 375: Is this the data synced back from RSP on overlay switches?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+    assertf(rspq_ctx != &highpri, "cannot create syncpoint in highpri mode");
+    rspq_int_write(RSPQ_CMD_TEST_WRITE_STATUS, 
+        SP_WSTATUS_SET_INTR | SP_WSTATUS_SET_SIG_SYNCPOINT,
+        SP_STATUS_SIG_SYNCPOINT);
+    return ++rspq_syncpoints_genid;
+}
+
+bool rspq_check_syncpoint(rspq_syncpoint_t sync_id) 
+{
+    return sync_id <= rspq_syncpoints_done;
```
anacierdem: How does this behave on an overflow?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+}
+
+/** @brief Stop the RSP queue engine in the RSP */
+static void rspq_stop(void)
+{
+    MEMORY_BARRIER();
+    *SP_STATUS = SP_WSTATUS_SET_HALT;
+    MEMORY_BARRIER();
+
+    rspq_is_running = 0;
```
anacierdem, line 696: How many cycles does it take for the RSP to actually halt, I think it is not possible to know e.g if a command was already running? Might be a good idea to warn about doing a `rspq_stop` + `rspq_start` in close succession might not work correctly. Would adding a a syncpoint help here?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+ * for efficiently check when it is time to switch to the other buffer: basically,
+ * it is sufficient to check whether "cur > sentinel".
+ * 
+ * The current queue is stored in 3 global pointers: #rspq_ctx, #rspq_cur_pointer
+ * and #rspq_cur_sentinel. #rspq_cur_pointer and #rspq_cur_sentinel are 
+ * external copies of the "cur" and "sentinel" pointer of the
+ * current context, but they are kept as separate global variables for
+ * maximum performance of the hottest code path: #rspq_write. In fact, it is
+ * much faster to access a global 32-bit pointer (via gp-relative offset) than
+ * dereferencing a member of a global structure pointer.
```
anacierdem, line 405: How do we keep the ones in the struct in sync?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+    MEMORY_BARRIER();
+    *SP_STATUS = rspq_ctx->sp_wstatus_clear_bufdone;
+    MEMORY_BARRIER();
+
+    // Switch current buffer
+    rspq_ctx->buf_idx = 1-rspq_ctx->buf_idx;
+    uint32_t *new = rspq_ctx->buffers[rspq_ctx->buf_idx];
+    volatile uint32_t *prev = rspq_switch_buffer(new, rspq_ctx->buf_size, true);
+
+    // Terminate the previous buffer with an op to set SIG_BUFDONE
```
anacierdem, line 983: Couldn't we append these commands before waiting for the prev buffer to finish?

src/rspq/rspq.c
```diff
@@ -0,0 +1,1235 @@
...
+    // Most of the times, the above is enough. But there is a small and very rare
+    // race condition that can happen: if the above status change happens
+    // exactly in the few instructions between RSP checking for the status
+    // register ("mfc0 t0, COP0_SP_STATUS") RSP halting itself("break"),
+    // the call to rspq_flush might have no effect (see command_wait_new_input in
+    // rsp_queue.S).
+    // In general this is not a big problem even if it happens, as the RSP
+    // would wake up at the next flush anyway, but we guarantee that rspq_flush
+    // does actually make the RSP finish the current buffer. To keep this
+    // invariant, we wait 10 cycles and then issue the command again. This
```
anacierdem, line 1009: I think this explains my concerns there :)

anacierdem (commented):
I went over the implementation rather quickly as well. Please forgive me if some of the questions does not make sense. I believe most of the things might feel trivial to you in hindsight.

Can't believe how much work you have put on this!

I'm ready to merge this whenever you feel comfortable. I will also run the tests on my hardware, just in case we find something different with the setup.

include/rsp_queue.inc
snacchus: That's at the end of the file, after the common text segment.

anacierdem: I mean dont we load the code here? Let me read it again probably an oversight by me.

include/rsp_queue.inc
snacchus: We prefetch the first 16 bytes of each command at this point into registers `a0-a3`, so overlays don't have to load those themselves. So if a command wants to access bytes 0x8-0x11, they can just read the contents of `a2` for example. We decided to include this convenience feature because the internal commands benefit from it as well (most notably `RSQPCmd_Dma`, which uses all 16 bytes). This ends up saving instructions overall because the code for loading any of the first 16 bytes doesn't need to be duplicated anywhere.

src/rspq/rspq.c
snacchus: Yes, the CPU knows the state size, but it's not relevant for this struct. Since the saved state is always located within the overlay data (it should be anyway, I'll add some asserts for this), the RSP will load both in a single DMA, so it doesn't need to know the state size for this step. When the overlay is unloaded however (or rather a new overlay is loaded), it does need the size for saving the state back to RDRAM. But because the overlay is loaded at that point, it can get the state size from the header. See rsp_queue.inc, lines 326-328.

Some basic checking of the state size is done in `rspq_overlay_register_internal` (lines 832-834). But as I said, it wouldn't hurt to also check that the state is contained within the overlay data.

include/rsp_queue.inc
snacchus: If the size could be zero, we'd need an additional branch for saving the state to RDRAM (rsp_queue.inc, lines 326-328). Also because the state size is stored as `size - 1` (so it can be used for DMA straight away), a size of zero would cause an overflow, which in turn causes undefined behavior because the RSP would attempt to DMA 0xFFFF + 1 = 4k of DMEM (so all of it I guess).

In short: This is an artificial restriction so we can save yet another instruction :)

include/rsp_queue.inc
anacierdem: Thats seems reasonable. I guess its not worth a jump either.
i.e we could have a common function to get the parameters to save a few cycles when they are not needed.

src/rspq/rspq.c
snacchus: No, this is general data that the ucode needs to run the queue engine. It's initialized once by the CPU in `rspq_init`.

src/rspq/rspq.c
snacchus: I'd say it breaks briefly, and as soon as both `rspq_syncpoints_genid` and `rspq_syncpoints_done` have overflown it should go back to working correctly. `rspq_syncpoints_genid` will overflow first, which means during the period in which `rspq_syncpoints_done` hasn't, syncpoints that are created after the overflow will falsely be reported to have been reached.

Assuming we constantly generate one syncpoint every millisecond (at 60Hz that's 16 per frame), the overflow would happen after roughly 50 days of continuous uptime, based on my quick calculation. If that's is not acceptable we could make the syncpoint be 64 bits I guess?

include/rsp_queue.inc
snacchus: Yes, that's the race condition we prevent by writing to `SP_STATUS` twice in `rspq_flush_internal`.

include/rsp_queue.inc
snacchus: This resets the `SP_STATUS_SIG_MORE` signal (`SP_WSTATUS_CLEAR_SIG_MORE` is loaded into `t0` a couple of lines above), so we'll know if the CPU sets that signal in the time between the RSP waking up and checking the signal the next time. If it's still unset when checking again, the RSP goes back to sleep. Otherwise, it keeps fetching new commands. If we didn't reset the signal here, there would be cases where the RSP unnecessarily fetches new commands even though none are available. That wouldn't break anything, but would waste some DMA bandwidth.

src/rspq/rspq.c
anacierdem: Consumers can also create syncpoints that might get triggerred many times a frame, so this might become a concern.
Still if we can verify it briefly breaks and then continue working, I think that is acceptable. Should be easy to test by starting at a high id initially.

include/rsp_queue.inc
snacchus: Yes, reading and writing the SP status register is atomic. The only instance where that matters are the signals though I think, which can be set/unset arbitrarily by both CPU and RSP. This command is primarily used for synpoints. In addition to raising an interrupt, syncpoints will set a signal when triggered (`SP_STATUS_SIG_SYNCPOINT`). Before triggering a syncpoint, this command will check if that signal is still set (it gets reset by CPU when handling the interrupt). This is done to prevent these interrupts from being coalesced, which would cause syncpoints to be skipped.

I'm not sure which RDRAM writes you are referring to? I think the SP status register works differently though, since it is wired up directly to a specific virtual address instead of going through RDRAM.

src/rspq/rspq.c
snacchus: The global pointers are saved back to the struct when switching context (see `rspq_switch_context`) and vice versa it updates the global pointers with the ones from the new context that is switched to. This happens when entering/exiting high-priority mode (see `rspq_highpri_begin` / `rspq_highpri_end`) or block mode (see `rspq_block_begin` / `rspq_block_end`).

src/rspq/rspq.c
snacchus: Can you elaborate? In case you are concerned about the fact that this "hack" is even necessary, it's pretty much the only way to prevent a race condition (see my other comment in `RSPQCmd_WaitNewInput`). At least to my knowledge.

src/rspq/rspq.c
snacchus: I think @rasky will have to explain this one :)

include/rsp_queue.inc
snacchus: That struct is mostly an implementation detail so I wouldn't add it to the "public" docs (I'm not sure if there is even a way to include these docs in doxygen?). I'll add a comment about it within the macro definition though.

src/rspq/rspq.c
snacchus: `rspq_stop` is only used to tear down the entire system when calling `rspq_close`, so it should be on the user to make sure that all commands they care about have finished executing. Also, `rspq_close` will probably never be called in an actual game because there isn't really any reason to do so. The system can be used continuously between scenes, and synchronisation can be done with syncpoints by the user. We only added this function for completeness's sake and to make the tests work basically.

Calling `rspq_init` straight after `rspq_close` shouldn't really cause issues because the former will always perform the entire setup necessary to bootstrap the rspq system.

And just in case you might have missed it: `rspq_start` and `rspq_stop` are not public API :)

src/rspq/rspq.c
snacchus: True, there isn't really an upper bound on how many syncpoints can be created. OTOH, we _do_ warn about abusing them in the docs :)

Jokes aside, even if they only break briefly that could have unforeseen and dramatic consequences for an application. So we should do our best to prevent that situation. @rasky It should be possible to bump the syncpoints IDs to 64 bit, right?

src/rspq/rspq.c
anacierdem: I was referring to my comment at https://github.com/DragonMinded/libdragon/pull/253/files/2feb4242db3097518ca09a28b5c373ff73597083#diff-8f33e64dc050f62d7f9988d1952992863c56dca0a901e9bddd099c76825e816bR237 When reading the PR, this actually answered my question there, as you have also mentioned `RSPQCmd_WaitNewInput`

src/rspq/rspq.c
rasky: We could append `WRITE_STATUS`, but not `JUMP`, because `JUMP` requires the new buffer to be cleared (or the RSP could pick up the jump before you clear it, and you would end up with the RSP re-execing the previous contents of the new buffer).

Now that I think of it, one option would be to clear just the first word of the new buffer, which is enough to keep the RSP waiting there even in the worst case. So the sequence might be:

 * Clear first word of the new buffer (crossing finger it's been already executed -- very likely but we have no explicit knowledge that this was done)
 * Write the epilogue of the current buffer
 * Wait for the signal to be set
 * Clear the new buffer and set it as current

I'm a little hesitant on this because it'd introduce another delicate synchronization process, for something which I feel might be a marginal performance improvement. I would say the largest time that you would parallelize to RSP is the memset that clears the new buffer, which I previously already attempted to parallelize by using DMA, but then I decided to punt for simplicity.

src/rspq/rspq.c
rasky: With the 32-bit counter and 100 syncpoints per frame on a 60 fps application, we get an overflow after 7 days. This said, if we want to fix it, we can either switch to 64bit (I don't think there's any problem with that), or use a difference rather than a comparison to handle the overflow condition, just like `TICKS_BEFORE`. Using the latter would mean that the only limit is not to have more than 2 billion syncpoints which are enqueued and not yet executed, which I would say it's unlikely :)

snacchus: I like the difference method, which I hadn't even thought of. It guarantees syncpoints always working correctly, because it's not even physically possible to have that many commands in the queue. It's not even possible to have that many syncpoints in all of RDRAM with the expansion pack! So I'd say we do that?

include/rsp_queue.inc
snacchus: Yeah that would add a lot of extra instructions to the common code, which we are trying to keep as small as possible, trading off a little bit of performance in multiple places.

src/rspq/rspq.c
snacchus: I mean it's also pretty much guaranteed that a 64 bit counter will never overflow, but it does involve doubling the storage size of syncpoints.

include/rsp_queue.inc
anacierdem: Ok this is where we clear the signal, I totally missed that. So is this example flow correct?

- `RSPQCmd_WaitNewInput` does the `mfc0 t0, COP0_SP_STATUS` and sees `SIG_MORE` is not set
- CPU writes `SIG_MORE` and starts waiting for 10 instructions
- RSP goes into sleep
- CPU re-writes `SIG_MORE`
- RSP wakes and clears `SIG_MORE`
- RSP executes the written command and the command jumps to `RSPQ_Loop` until there is no `RSPQCmd_WaitNewInput`
- RSP does the `mfc0 t0, COP0_SP_STATUS` sees his cleared value (i.e CPU didn't write anything and did not flush again) and sleeps

snacchus: Yes, exactly!

include/rsp_queue.inc
anacierdem: It is possible to dispatch 32bit memory mapped writes to the RDRAM which all go through RI and I thought there might be something in common between CPU-RI and CPU-SP_STATUS that does the arbitration, probably this is a different mechanism though. Maybe @rasky knows.
Regardless, the important thing these being atomic, which means either SPU or RSP stalls when the other is dispatching an SP_STATUS update. I wonder how this works :)
Probably a non-issue for this PR if we know it is atomic.

include/rspq.h
```diff
@@ -0,0 +1,716 @@
...
+ * @brief Initialize the RSPQ library.
+ * 
+ * This should be called by the initialization functions of the higher-level
+ * libraries using the RSP command queue. It can be safely called multiple
+ * times without side effects.
+ * 
+ * It is not required by applications to call this explicitly in the main
+ * function.
+ */
+void rspq_init(void);
```
anacierdem, line 200: One important thing to note; once the system is initialized the cpu should not DMA anything to the RSP unless it is known that the rspq is halted on the CPU, which can be achieved by `rspq_wait`.
This can include a scenario where the user wants to switch over to a custom ucode for some reason. There is no DMA synchronization and doing a DMA while rspc is running is UB.
Might be obvious but wanted to point it out anyways :)

snacchus: Yeah pretty much. If the user wants to load some custom ucode when rspq is running, they have to do the synchronization themselves.

This made me think - should we check whether rspq is actually running when the user accesses the API? Or maybe just add a function `rspq_is_running` or something so other components that depend on it (like the mixer) can perform this check? Or we document the dependency on those components to warn the user about calling `rspq_close` before the dependent components have been closed properly.

rasky: I think we should come up with an answer to the main question: "how do I use rspq together with non-rspq ucodes within the same application, and more specifically within the same frame?". I think in the medium term, most ucodes will switch to rspq as it's generally easier to write a rspq overlay than writing a ucode from scratch (many less things to care about -- rspq is a joy to use as RSP programmer!), but we can't rule out mixing with non-rspq ucodes.

For instance, is it safe to switch ucode after rspq_wait()? And if so, how do you switch back to rspq when the other ucode has finished? My feeling is that rspq_wait() is not enough because the overlay state has not been saved at that point and would be lost if something is loaded into DMEM. So if we want to explicitly support this we need something like `rspq_pause()` or `start/stop` or similar APIs, plus tests :)

snacchus: So ideally, you would just call `rspq_stop`, load your own ucode and run it, and then call `rspq_start` and it would just be ready to go from where it left off, without the need to re-register overlays again and such. I think a reasonable restriction is that there cannot be any commands in the queue while stopping that will resume executing after restarting. So internally it would first call `rspq_wait` and then maybe load some dummy overlay so the current overlay's state is saved. I can look into implementing this.

anacierdem: I think we don't need the pause functionality at this time and we can develop it later. For now, it should be ok to just have some warnings to note manual handling is necessary for opting out of rspq. Using mixer also means it will be inited so a small note there would be nice as well.

rasky: > I think a reasonable restriction is that there cannot be any commands in the queue while stopping that will resume executing after restarting

Yes, agreed. If you think it's easy to implement go ahead, but as Anacierdem says, it's not a big deal, as long as we document it.

src/rspq/rspq.c
rasky: Yeah, I think the difference is fine, it's easy enough to implement :)

include/rsp_queue.inc
rasky: I'm not sure what you're referring to, wrt RDRAM.

Writes to SP_STATUS are atomic as in the fact that the write masks have separate set/reset flags means that you don't need a read-modify-write cycle which can introduce race conditions. For instance, let's assume that SP_STATUS didn't even the special SP_WSTATUS_* bits and could just be written like a regular memory location, using SP_STATUS_* bits. To turn off SP_STATUS_SIG_SYNCPOINT,  the CPU would have to do:

```C
   *SP_STATUS = *SP_STATUS & ~SP_STATUS_SIG_SYNCPOINT;
```

which means that the CPU would first read the register and then write it. If the RSP races with this and turn on a different bit, that write would be lost.

Instead, given that the CPU can simply do `*SP_STATUS = SP_WSTATUS_CLEAR_SIG_SYNCPOINT` and everything is handled internally, it means that the RSP can at the same time turn on a different bit and everything works as expected. Likely, what happens if the two memory writes happens exactly at the same clock cycle, is that either of the two gets stalled one cycle more, and so they are actually serialized.

Obviously, this doesn't mean that modifying *the same bit* by both CPU and RSP has a well defined behavior. That is a true race condition where there is no amount of hardware suppor that can save you. If the CPU is setting one bit and the RSP is resetting it at the same time, it's undefined behavior whether the bit ends up being set or cleared. After all, there is no way to synchronize the two cores, so what we think it's happening "at the same time" could be slightly drifted in any direction, and we would have no way to prove it or control it. 

In other words, this is the same guarantee of an atomic variable in modern programming. You can surely access it from two threads without any other locking, but you must behave well while accessing it, following a programming pattern that takes into account that either side can access it at any moment.

anacierdem: wrt RDRAM what I meant was it is not possible to partially write a 32 bit value by CPU memory-mapped writes.

> Writes to SP_STATUS are atomic as in the fact that the write masks have separate set/reset flags means that you don't need a read-modify-write cycle which can introduce race conditions.

This is totally true if different cores were accessing on-chip SP_STATUS interfaces. But in this case the RSP is a separate chip and when the CPU tries to set some bits via the SP_STATUS flags, they need to propagate to the RSP chip and will actually physically change bits on some digital interface. If this digital interface is physically the same with the one used internally by RSP (it probably is) then there need to be a mechanism to tell the CPU that this is being written to stall it or an intermediate buffer to keep the flags coming from the CPU until the RSP operation finishes, if any.

If the CPU-RSP interface is not 32 bit and instead if the CPU is setting the flags 8 bits at a time and there isn't a mechanism to sync the two, RSP can partially override the *flags* that are being written before they are latched to update the register state.

The same is probably not a concern when reading the register contents even if the CPU can execute the flags partially because we are not interested in flags changing in tandem. Though it would mean halt and sig more may not be set together causing other problems...

But if the propagation of the flags from CPU to RSP is in 32bit chunks (like how RI interface works - 32bit at a time - like we are certain we cannot write a partial command to the RDRAM) while mutually stalling each other we are all fine! This is what I was trying to explain with a few poorly written sentences before 😆 

 >  everything is handled internally, it means that the RSP can at the same time turn on a different bit and everything works as expected

Then we are allright! I know I maybe overthinking a little but this is N64 with its various hardware quirks!

> If the CPU is setting one bit and the RSP is resetting it at the same time, it's undefined behavior

This made me think of a potential scenario;

- CPU kicks off the first sig more
- RSP loads this update by `mfc0 t0, COP0_SP_STATUS` and decides to wake up
- CPU kicks off second sig more in tandem with `mtc0 t0, COP0_SP_STATUS`

If RSP wins, sig more is reset and RSP will sleep when it is done.

If CPU wins, sig more is set, even though the RSP is already awake, potentially trying to fetch new commands and this may cause partially written commands to execute?

I really hope I am wrong on this one.

snacchus: > If CPU wins, sig more is set, even though the RSP is already awake, potentially trying to fetch new commands and this may cause partially written commands to execute?

As we already discussed on Discord, this is not the case since we explicitly protect against such a scenario by always writing the first word of each command last.

include/rspq.h
snacchus: It's simple in theory, but in practice I'd have to do some more plumbing to make it work which is not entirely trivial. For example we would need to consider the case in which all overlay slots are occupied. Then one of those overlays needs to be temporarily unregistered or somehow backed up, then the dummy overlay takes its place to trigger the state-saving mechanism. Also things like the pointer stack and the current RDRAM address need to be backed up from DMEM.

I've decided to leave this for a follow up PR for now.

src/rspq/rspq.c
snacchus: I've changed `rspq_check_syncpoint` to use the difference.

include/rspq.h
snacchus: I actually forgot to remove the comment, so I've done that now :)

snacchus:
@meeq @Ryzee119 since this PR also affects the mixer, it would be fantastic if you guys could test it in your games to verify we didn't break anything. It should just work out of the box, no additional code required. And sound the same of course :)

include/rspq.h
```diff
@@ -0,0 +1,713 @@
...
+ *       is meant to be a one-time event. Otherwise the same syncpoint would
+ *       potentially be triggered multiple times, which is not supported.
+ * 
+ * @note It is not possible to create a syncpoint from the high-priority queue
+ *       due to the implementation requiring syncpoints to be triggered
+ *       in the same order they have been created.
+ * 
+ * @see #rspq_syncpoint_t
+ */
+rspq_syncpoint_t rspq_syncpoint(void);
```
rasky: I did a final pass on rspq.h and I suggest that we rename this function. I find it a bit confusing to have both `rspq_syncpoint` and `rspq_syncpoint_t`. I would say that this should be called `rspq_syncpoint_new()`. Also, we should probably rename the other functions into `rspq_syncpoint_check` and `rspq_syncpoint_wait`, to adhere to the convention of having a hierarchical prefix.

snacchus: I renamed those functions like you suggested

PR #262, opened 2022-03-27
Get rid of the second workflow

anacierdem:
We used to have a separate workflow for the image build for historical
reason. We now already have have multiple build and push steps, so it
is ok to have an additional one to push the latest image on the default
branch.

This enables us removing the secondary build & push step saving possible
headaches in the future.

We always compare files to default branch except for a push to the default
branch, which uses `event.before`.

I have also split different workflows into their own file to ease readability.

Also enabled concurrency for doc and image updates to prevent any race
conditions, just in case.

The second commit splits the workflows into files and can be reviewed
separately.


.github/workflows/build-documentation.yml
```diff
@@ -0,0 +1,39 @@
...
+# Make sure we don't have more than one active workflow to prevent race conditions
+# e.g a previous build may update docs later and overwrite if we don't have this.
+# It is ok to have parallel runs for push and PR events and from different branches.
+# We can cancel previous runs for non-trunk events.
+concurrency:
+  group: ${{ github.ref }}${{ github.event_name }}
+  cancel-in-progress: ${{ github.ref != format('refs/heads/{0}', github.event.repository.default_branch) }}
+
+jobs:
+  Build-Tools-Windows:
```
rasky: This should probably be renamed?

rasky (approved):
Fine by me! As I understand this is just a refactoring?

anacierdem:
Correct, it is fundementally a refactor but I wouldn't call it "just" a refactor :) I expect the same behaviour though and didn't find anything unexpected on my tests (including a PR merge to the trunk on my fork)

PR #263, opened 2022-03-28
dragonfs: assert out with a clean error message on broken emulators

rasky:
Some emulators do not still fully emulate behaviors of PI DMA
(notably, Project 64). When a libdragon ROM that uses DFS is run on
those emulators, the ROM fails typically when opening files with a
confusing error message such as "file not found".

This commit introduces an explicit check that the PI DMA behaves as
expected. It is obviously moot on real hardware, but on broken emulators
it fails the check, exposing a readable error message to users.

### DIFF
--- a/src/dragonfs.c
+++ b/src/dragonfs.c
@@ -1302,6 +1302,31 @@ static filesystem_t dragon_fs = {
 };
 
 /**
+ * @brief Verify that the current emulator is compatible with DFS, and assert
+ *        if it is not.
+ */
+static void __dfs_check_emulation(void)
+{
+    const uint32_t ROM_ADDR = 0x10000000;
+
+    uint32_t data = io_read(ROM_ADDR);
+    
+    /* DFS requires working support for short DMA transfers. Verify that they
+     * work and if they don't, assert out. */
+    uint8_t buf[4] __attribute__((aligned(8))) = { 0xFF, 0xFF, 0xFF, 0xFF };
+    data_cache_hit_writeback_invalidate(buf, 4);
+    dma_read(buf, ROM_ADDR, 3);
+
+    if (buf[0] == ((data >> 24) & 0xFF) &&
+        buf[1] == ((data >> 16) & 0xFF) &&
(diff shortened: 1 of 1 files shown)


anacierdem:
I'm into emulator detection and asserting if one is detected that
won't be able to run libdragon but I'm not sure if this is the way
to go. I still think the DMA fix-ups for short length and non-standard
alignments is limiting faster ROMs which should be allowed as the
hardware is capable of doing faster PI transfers. There are no official
ROMs that did that but we know DD used faster speeds for example.

Also considering we might implement a detection mechanism that
might solve differences at different ROM speeds, this change might
not be future compatible. Is there another way to detect emulators
without using these ver specific DMA implementations?

rasky:
This test verifies if the current implementation of dragonfs is compatible. The current implementation uses short odd transfers, hence this test verifies that the emulator is capable of transferring 3 bytes. The current implementation does not use misaligned transfers, hence the test doesn't do it. I don't think we should "detect emulators" but rather "detect whether what we are currently doing is working or not". We currently do short odd transfers, so we check that.

My plan would be to add similar tests for other subsystems that are broken in common emulators. For instance wav64 / xm64 currently rely on misaligned transfers so I would add detection for that on the first opening of a wav64 or xm64 file. This means you get a nice error message instead of corrupted audio.

MPEG1 uses the weird RSP opcode; I'd test those when the user opens their first MPEG1 movie file, so users get a nice error message rather than a corrupted video.

And so on.

We currently don't have a feature to bump ROM speeds, and honestly I wouldn't be 100% sure it breaks short odd transfers. If it does, we can analyze that and decide to stop doing that; if we do, that day we will remove this test. In general, I think that if we evolve the code, we can adapt the emulator checks as well as the code evolves. I don't see this PR as doubling down on short odd transfers at all; it just helps confused users, but it doesn't make that feature more irreplaceable than before.

anacierdem:
That's fair. I think I may have objected impetuously knowing the odd transfers are not working at higher speeds, but doing feature detection is the correct approach. I can prepare a test for odd transfers at some point.

PR #264, opened 2022-03-31
Optimize joybus_exec callback, remove nested function

meeq:
Use `joybus_callback_t` to unify usage of callback function pointer


meeq:
Closing this; after discussing on Discord it was determined that there is no point in optimizing a function that spends most of its time waiting in a spinlock.

PR #265, opened 2022-04-06
rsp.inc: unify support for e(x) syntax

snacchus:
The vector element syntax that was introduced for computational vector opcodes is now supported in `mtc2`/`mfc2` and vector load/store opcodes as well. Until now, you had to specify the element as the byte offset into the vector register for those opcodes. Now it is possible to specify the lane index directly instead. For example:
```
# Before:
lsv $v00,4,  0x10,s4
mfc2 t0, $v01,6

# After:
lsv $v00,e(2),  0x10,s4
mfc2 t0, $v01,e(3)
```

### DIFF
--- a/include/rsp.inc
+++ b/include/rsp.inc
@@ -270,6 +270,349 @@ makeNotImplemented tnei
     .set $v31, 0x11F
 .endm
 
+# Encode vector elements for use in e(x) style syntax.
+.macro veVectorElements
+    .set VE_v,  0x0
+    .set VE_0q, 0x2
+    .set VE_1q, 0x3
+    .set VE_0h, 0x4
+    .set VE_1h, 0x5
+    .set VE_2h, 0x6
+    .set VE_3h, 0x7
+    .set VE_0,  0x8
+    .set VE_1,  0x9
+    .set VE_2,  0xA
+    .set VE_3,  0xB
+    .set VE_4,  0xC
+    .set VE_5,  0xD
+    .set VE_6,  0xE
+    .set VE_7,  0xF
+.endm
+
+# Encode byte offsets for use in e(x) style syntax.
+.macro byteVectorElements
+    .set VE_v,  -1
+    .set VE_0q, -1
+    .set VE_1q, -1
+    .set VE_0h, -1
+    .set VE_1h, -1
+    .set VE_2h, -1
+    .set VE_3h, -1
+    .set VE_0,  0x0
+    .set VE_1,  0x2
+    .set VE_2,  0x4
+    .set VE_3,  0x6
+    .set VE_4,  0x8
+    .set VE_5,  0xA
+    .set VE_6,  0xC
+    .set VE_7,  0xE
+.endm
+
+# Encode byte offsets for use in e(x) style syntax.
+.macro laneAsByteVectorElements
+    byteVectorElements
+.endm
+
+# Encode lane indices for use in e(x) style syntax.
+.macro laneVectorElements
+    .set VE_v,  -1
+    .set VE_0q, -1
+    .set VE_1q, -1
+    .set VE_0h, -1
+    .set VE_1h, -1
+    .set VE_2h, -1
+    .set VE_3h, -1
+    .set VE_0,  0x0
+    .set VE_1,  0x1
+    .set VE_2,  0x2
+    .set VE_3,  0x3
+    .set VE_4,  0x4
+    .set VE_5,  0x5
+    .set VE_6,  0x6
+    .set VE_7,  0x7
+.endm
+
+# This will encode element accessors for computational
+# opcodes, which interpret the element as either lane index + 8
+# or broadcast specifiers. For these, the .bN accessors are not valid.
+.macro veAccessorEncoder prefix, base
+    .set \prefix\().v,   (\base + 0x0)
+    .set \prefix\().q0,  (\base + 0x2)
+    .set \prefix\().q1,  (\base + 0x3)
+    .set \prefix\().h0,  (\base + 0x4)
+    .set \prefix\().h1,  (\base + 0x5)
+    .set \prefix\().h2,  (\base + 0x6)
+    .set \prefix\().h3,  (\base + 0x7)
+    .set \prefix\().e0,  (\base + 0x8)
+    .set \prefix\().e1,  (\base + 0x9)
+    .set \prefix\().e2,  (\base + 0xA)
+    .set \prefix\().e3,  (\base + 0xB)
+    .set \prefix\().e4,  (\base + 0xC)
+    .set \prefix\().e5,  (\base + 0xD)
+    .set \prefix\().e6,  (\base + 0xE)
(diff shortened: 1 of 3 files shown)


src/audio/rsp_mixer.S
```diff
@@ -276,9 +276,9 @@ command_exec:
 	#define samples_left    t4
 	#define outptr          s8
 
-	vxor v_zero, v_zero, v_zero,0
+	vxor v_zero, v_zero, v_zero
 	li t0, %lo(VCONST_1)
-	lqv v_const1,0, 0,t0
+	lqv v_const1,e(0), 0,t0
```
rasky: Honestly I don’t know if I’d make these changes. It makes the code more verbose and honestly the real solution here would be making that argument optional. If that’s not easy to do, maybe just leave 0?

include/rsp.inc
```diff
@@ -474,52 +539,52 @@ makeOpInstruction vsubc, 0b010101
 makeOpInstruction vxor, 0b101100
 
 /** @brief Load Byte into Vector Register */
-makeLsInstructionByte load, lbv, 0b00000
+makeLsInstructionByte load, byte, lbv, 0b00000
```
rasky, line 954: I’ll review these properly

src/audio/rsp_mixer.S
snacchus: Yeah you're right. I'll look into this again. I wonder if there is a good, general approach to providing multiple "overloads" of GAS macros

snacchus:
Update: In addition to the `e(x)` syntax, all relevant opcodes now also support a new "dot syntax":
```
vadd $v01, $v02, $v03.h2   # equivalent to $v03,e(2h)
vsub $v04, $v05.e4         # equivalent to $v05,e(4)

lsv $v10.e5, 0x10,s0      # equivalent to $v10,10
ssv $v11.b4, 0x00,s1      # equivalent to $v11,4

mfc2 t0, $v20.e7           # equivalent to $v20,14
```

Also, the element is now optional everywhere. For computational opcodes, it defaults to the normal vector access (like before). For all others, the default is lane 0 (or byte offset 0).

src/audio/rsp_mixer.S
snacchus: The element is now optional

include/rsp.inc
```diff
@@ -473,138 +732,229 @@ makeOpInstruction vsubc, 0b010101
...
 /** @brief Load Double into Vector Register */
-makeLsInstructionDouble load, ldv, 0b00011
+makeLsInstructionDouble load, byte, ldv, 0b00011
 /** @brief Load Packed Fourth into Vector Register */
-makeLsInstructionQuad load, lfv, 0b01001
+makeLsInstructionQuad load, lane, lfv, 0b01001
 /** @brief Load Packed Half into Vector Register */
-makeLsInstructionQuad load, lhv, 0b01000
+makeLsInstructionQuad load, lane, lhv, 0b01000
```
rasky: Both lhv and lfv do accept byte offsets.

include/rsp.inc
```diff
@@ -473,138 +732,229 @@ makeOpInstruction vsubc, 0b010101
...
+makeLsInstructionQuad load, byte, lqv, 0b00100
 /** @brief Load Quad (Rest) into Vector Register */
-makeLsInstructionQuad load, lrv, 0b00101
+makeLsInstructionQuad load, byte, lrv, 0b00101
 /** @brief Load Short into Vector Register */
-makeLsInstructionShort load, lsv, 0b00001
+makeLsInstructionShort load, byte, lsv, 0b00001
 /** @brief Load Transpose into Vector Register */
-makeLsInstructionQuad load, ltv, 0b01011
+makeLsInstructionQuad load, byte, ltv, 0b01011
```
rasky: ltv and stv are an exception :)

In theory the element is treated as a byte offset, but in practice the lowest bit is ignored. So semantically speaking I would say that it should only accept a lane, but encoding-wise a byte offset should be written.

include/rsp.inc
```diff
@@ -473,138 +732,229 @@ makeOpInstruction vsubc, 0b010101
...
+makeLsInstructionByte store, byte, sbv, 0b00000
 /** @brief Store Double from Vector Register */
-makeLsInstructionDouble store, sdv, 0b00011
+makeLsInstructionDouble store, byte, sdv, 0b00011
 /** @brief Store Packed Fourth from Vector Register */
-makeLsInstructionQuad store, sfv, 0b01001
+makeLsInstructionQuad store, lane, sfv, 0b01001
 /** @brief Store Packed Half from Vector Register */
-makeLsInstructionQuad store, shv, 0b01000
+makeLsInstructionQuad store, lane, shv, 0b01000
```
rasky: Both sfv and shv accept a byte offset.

include/rsp.inc
snacchus: fixed

include/rsp.inc
snacchus: I've added a special encoding for ltv and stv

include/rsp.inc
snacchus: fixed

include/rsp.inc
```diff
@@ -464,147 +770,270 @@ makeOpInstruction vrsq, 0b110100
...
+    .endif
+
+    .ifnb \vs
+        # \vs is element
+        vsarInternal \vd, \vd, \vd, \vs
+        .exitm
+    .endif
+
+    .error "Element is required"
+.endm
```
rasky, line 811: `vsar` basically only cares about `vd` and `elem`. `vt` and `vs` are ignored, to the best of my understanding.

So the only good syntax should be 2-arg. Given that we've forced 4-args up until now, I would probably support that as backward compatibility as well, marking it as deprecated in comments. I would not introduce the 3-arg syntax since it's not required.

Also I'd change `vsarInternal` to use the 2-arg syntax, so that it's clear that the other elements are ignored (can be forced to 0 while building the opcode).

.github/workflows/build-documentation.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           echo "::set-output name=default_ref::${{ format('refs/heads/{0}', github.event.repository.default_branch) }}"
 
       - name: Run Doxygen
-        uses: mattnotmitt/doxygen-action@v1
+        uses: mattnotmitt/doxygen-action@master
```
meeq: This change seems orthogonal to the rest of the PR, and is not good practice: "master" is not a tagged version, so it can cause breakage without warning and provides no guarantees of compatibility.

It looks like `1.9.2` is the latest stable version of the package. `1.9.3` is not considered stable because it depends on an `edge` version of Alpine Linux.

snacchus: The reason for this was the doxygen build being broken at the time, which made the CI build fail. Rasky and I then came up with this solution. That issue seems resolved now though (some other PRs made it through without this change in the meantime) so I'll roll it back.

PR #267, opened 2022-04-19
mkdfs: added error message on empty directories

msklywenn:
I was super confused why mkdfs would not work until I stumbled upon the comment in the source about empty directories being not allowed. So I added an error message.

### DIFF
--- a/tools/mkdfs/mkdfs.c
+++ b/tools/mkdfs/mkdfs.c
@@ -253,6 +253,10 @@ uint32_t add_directory(const char * const path)
     closedir(dirp);
 
     /* Will return 0 if we don't find any entries (don't support directories without files) */
+    if (first_entry == 0)
+    {
+        fprintf(stderr, "Directory %s is empty, not allowed.\n", path);
+    }
     return first_entry;
 }
 


rasky:
Can you please explain what was the behavior that you were expecting and didn't work? Having an empty directory bundled in the ROM and then visibile via directory walking?

msklywenn:
No, my program just stopped building and I didn't understand why, that's all. I had moved files around and left a sub directory empty. "Error creating filesystem" isn't super helpful in that case, especially if you have no idea that empty directories are not allowed.

rasky:
Oh I see now, thanks.

I've opened https://github.com/DragonMinded/libdragon/pull/268 which is similar to yours, but doesn't give an error message on empty subdirectory, it just shows the error when the top-level directory is empty. I'll merge it as soon as CI finishes.

msklywenn:
@rasky this was not just for the top level directory. If any subdirectory is empty, it will also fail. The change you pushed will be confusing because it will say the top level is empty while it’s a subdirectory that is missing files

rasky:
Oh you're right, I misread the code. I would change it to ignore empty subdirs, I don't see any need to error out in that case. Do you agree?

msklywenn:
That would be even better indeed

rasky:
OK see #269

PR #270, opened 2022-04-27
rspq: implement RSP assertion support for overlays

rasky:
With this commit, rspq forwards to overlay's assert handler all assert
codes that it doesn't know of. This allows each overlays to handle their
own assertions.


src/rspq/rspq.c
```diff
@@ -482,7 +485,13 @@ static void rspq_crash_handler(rsp_snapshot_t *state)
...
     printf("RSPQ: Current DRAM address: %08lx + GP=%lx = %08lx\n", 
         rspq->rspq_dram_addr, state->gpr[28], cur);
-    printf("RSPQ: Current Overlay: %02x\n", rspq->current_ovl / sizeof(rspq_overlay_t));
+
+    int ovl_idx = rspq->current_ovl / sizeof(rspq_overlay_t);
+    const char *ovl_name = "?";
+    if (ovl_idx < RSPQ_MAX_OVERLAY_COUNT && rspq_overlay_ucodes[ovl_idx])
+        ovl_name = rspq_overlay_ucodes[ovl_idx]->name;
+    printf("RSPQ: Current Overlay: %s (%02x)\n", ovl_name, ovl_idx);
+    printf("%p %s\n", rspq_overlay_ucodes[0], rspq_overlay_ucodes[ovl_idx]->name);
```
snacchus: This seems like a leftover from something IINM?

src/rspq/rspq.c
```diff
@@ -506,15 +515,22 @@ static void rspq_crash_handler(rsp_snapshot_t *state)
 static void rspq_assert_invalid_command(rsp_snapshot_t *state)
 {
     rsp_queue_t *rspq = (rsp_queue_t*)state->dmem;
+
+    // Get overlay index a name. Be defensive against DMEM corruptions.
+    int ovl_idx = rspq->current_ovl / sizeof(rspq_overlay_t);
+    const char *ovl_name = "?";
+    if (ovl_idx < RSPQ_MAX_OVERLAY_COUNT && rspq_overlay_ucodes[ovl_idx])
+        ovl_name = rspq_overlay_ucodes[ovl_idx]->name;
```
snacchus: This could go into a helper function as it's a duplicate from the crash handler.

src/rspq/rspq.c
```diff
@@ -506,15 +515,22 @@ static void rspq_crash_handler(rsp_snapshot_t *state)
...
+
     uint32_t dmem_buffer = RSPQ_DEBUG ? 0x140 : 0x100;
     uint32_t cur = dmem_buffer + state->gpr[28];
-    printf("Invalid command\nCommand %02x not found in overlay %02x\n", state->dmem[cur], rspq->current_ovl / sizeof(rspq_overlay_t));
+    printf("Invalid command\nCommand %02x not found in overlay %s (0x%01x)\n", state->dmem[cur], ovl_name, ovl_idx);
```
snacchus, line 532: Could add an additional check for internal commands here, so it says "builtin" or something instead of "?".

snacchus (changes requested):
I tested it with rdpq and it works. The assert handler defined in the overlay ucode is correctly called. I Also found a couple of things that could be improved.

src/rspq/rspq.c
snacchus: Actually since the command is not valid anyway maybe not.

src/rspq/rspq.c
snacchus: Just referring to line 494 in case it's not clear.

rasky: Yes, removed.

src/rspq/rspq.c
rasky: Done

src/rspq/rspq.c
rasky: It still makes sense once it's moved into a separate function. Done.

PR #279, opened 2022-05-14
Adding option to change DFS offset

stefanmielke:
This PR adds an option to override the DFS file location offset without having to copy the whole `n64.mk` file in your project.

### DIFF
--- a/n64.mk
+++ b/n64.mk
@@ -1,5 +1,6 @@
 BUILD_DIR ?= .
 SOURCE_DIR ?= .
+N64_DFS_OFFSET ?= 1M # Override this to offset where the DFS file will be located inside the ROM
 
 N64_ROM_TITLE = "Made with libdragon" # Override this with the name of your game or project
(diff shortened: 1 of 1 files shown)


n64.mk
```diff
@@ -1,5 +1,6 @@
 BUILD_DIR ?= .
 SOURCE_DIR ?= .
+N64_DFS_OFFSET ?= 1M # Override this to offset where the DFS file will be located inside the ROM
```
anacierdem, line 3: I wonder if `?=` does make a difference here?

anacierdem (approved):
LGTM

n64.mk
stefanmielke: Yeah, you can't override on your Makefile if it's not there (at least it didn't work on my tests).

anacierdem: Interesting b/c we can override the title anyways, merging this one

anacierdem: Hmm I think the difference is in how it is used. One option is setting the var like a dependency and the other is as an individual statement

PR #280, opened 2022-05-14
Fix fseek bug on SD card

stefanmielke:
This PR fixes a bug on `debug.c` that would cause `fseek` to not return its current poisition, causing a bug when trying to get the size of a file on the SD card, as verified by the test changed to check for the behavior (fails before the fix).

It also adds an example that uses the SD card to load and display sprites (that only works with this fix).


examples/loadspritefromsd/loadspritefromsd.c
```diff
@@ -0,0 +1,121 @@
...
+		debugf("Error loading background\n");
+		return NULL;
+	}
+
+	const int size = filesize(fp);
+	debugf("Size: %d\n", size);
+
+	sprite_t *sprite = malloc(size);
+
+	setvbuf(fp, NULL, _IONBF, 0);
```
rasky: Do you really need this?

examples/loadspritefromsd/loadspritefromsd.c
```diff
@@ -0,0 +1,121 @@
...
+		sprite = spritesheet_load(sprites_sd[cur_sprite]);
+	else
+		sprite = spritesheet_load(sprites_rom[cur_sprite]);
+}
+
+int main(void) {
+	/* Initialize peripherals */
+	display_init(RESOLUTION_320x240, DEPTH_16_BPP, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE);
+	dfs_init(DFS_DEFAULT_LOCATION);
+	rdp_init();
```
rasky: This is probably not needed

examples/loadspritefromsd/loadspritefromsd.c
```diff
@@ -0,0 +1,121 @@
...
+	controller_init();
+	timer_init();
```
rasky: Also this

rasky (approved):
Thanks!

examples/loadspritefromsd/loadspritefromsd.c
stefanmielke: Nope, removed (along with the other unused code).

PR #282, opened 2022-05-31
New interrupt handler

rasky:
This commit series introduces a big refactoring of the interrupt handler, with the following benefits:

* It is much faster because fewer registers are saved (18 now vs 60 before). The fixed cost of an interrupt (before invoking the registered handler) is ~50% less now, so the interrupts are 2x faster.
* FPU registers are saved optimistically. If interrupt handlers don't touch FPU, they are not saved at all; otherwise, as soon as a handler does a COP1 operation, and exception is triggered that saves the registers on the stack and resumes execution. This is a good tradeoff because interrupt handles rarely need FPU anyway.
* The context is saved on the user stack, rather than a dedicated interrupt stack. This is required for the kernel anyway (and the implementation was in fact taken from the kernel branch), and it saves ~64K of RDRAM that was previously allocated for interrupt stack.

Possible future work in this area:

 * More micro-optimization in the code, like improving icache usage.
 * Introduce the concept of a interrupt handler, with a specific ABI, so that even fewer registers are saved when only interrupt handlers need to be saved.

### DIFF
--- a/include/exception.h
+++ b/include/exception.h
@@ -19,33 +19,33 @@
 enum
 {
     /** @brief Unknown exception */
-	EXCEPTION_TYPE_UNKNOWN = 0,
+    EXCEPTION_TYPE_UNKNOWN = 0,
     /** @brief Reset exception */
-	EXCEPTION_TYPE_RESET,
+    EXCEPTION_TYPE_RESET,
     /** @brief Critical exception */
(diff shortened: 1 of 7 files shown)


rasky:
I've updated the PR, as I had forgotten to port over an important bugfix for a potential race condition in interrupt.c.
Thanks to @stefanmielke for testing it and discovering the bug!

PR #283, opened 2022-06-03
controller: add assert if controller_init is not called

rasky:
In #234, we changed the controller module to do the scanning in
background.

Before #234, controller_init was just a big glorified
nop as it was just clearing global variables which were already cleared
in BSS. So a ROM could get away forgetting to call controller_init.

After #234, controller_init registers the interrupt callback to do
the scanning. So not calling controller_init equals to never read
any controller, so input is broken.

Add an assert to catch this case, to help people porting their code
to latest libdragon.

### DIFF
--- a/src/controller.c
+++ b/src/controller.c
@@ -65,6 +65,8 @@ static struct controller_data current;
 static struct controller_data prev;
 /** @brief True if there is a pending controller autoscan */
 static volatile bool controller_autoscan_in_progress = false;
+/** @brief True if the module was initialized */
+static bool controller_inited = false;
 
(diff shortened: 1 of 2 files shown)


src/controller.c
```diff
@@ -65,6 +65,8 @@ static struct controller_data current;
 static struct controller_data prev;
 /** @brief True if there is a pending controller autoscan */
 static volatile bool controller_autoscan_in_progress = false;
+/** @brief True if the module was initialized */
+static bool contoller_inited = false;
```
sp1187: contoller -> controller

rasky: Fixed thanks

PR #284, opened 2022-06-11
audio: add workaround for AI DMA hardware bug

rasky:
The AI DMA has a curious bug: if the address being transferred ends
exactly at a 0x2000 address boundary, next transfer gets corrupted
as the transfer happens from a location different than that requested.

The reason is possibly that the address register is split in two halves
with the lower part being 13 bit, and the carry from lower to higher is
delayed.

Add a workaround for it, adjusting the pointer in the rare case we get
such a pointer from the allocator.

Manually tested by forcing a reproduction of the bug, and then fixing it.


rasky:
Finally tested on real hardware, forcing a reproduction of the bug. It works.

PR #285, opened 2022-06-13
rspq: add support for big commands (up to 63 words)

rasky:
rspq currently supports command up to 16 words long. The limit is
mostly coming from the fact that we optimize checks for end of buffer
using a sentinel which is checked *after* writing each command (for
performance reasons) and it must be put to a position distant at least
like the biggest command from the end of the buffer. By increasing the
command size, the sentinel would have to move early in the buffer, and
more space would be lost.

16 words are not sufficient for RDP triangle commands, so we need to
increase it anyway.

With this commit, the command size is increased to 63 words. We introduce
a new API made of three functions:

 * rspq_write_begin to begin writing a big command
 * rspq_write_arg to add a word to the big command
 * rspq_write_end to finish writing a big command

With this API, the end-of-buffer condition is checked before writing,
so that it does not create constraints on the sentinel position. This
is slightly slower, but a good compromise for larger commands anyway.
Moreover, the new API is probably more versatile, rather than calling
rspq_write with 20+ arguments.

Written with @snacchus, in preparation for the new rdpq library.


include/rsp_queue.inc
```diff
@@ -158,6 +158,9 @@ _RSPQ_SAVED_STATE_END:
 # except to read command arguments.
 #define rspq_dmem_buf_ptr gp
 
+# Is set to the command's size in bytes when jumping to the command function
+#define rspq_cmd_size t7
```
anacierdem, line 163: Do we intentionally moved this over here? Also it is not `undef`d anymore?

include/rspq.h
```diff
@@ -388,6 +401,109 @@ void* rspq_overlay_get_state(rsp_ucode_t *overlay_ucode);
...
+ * 
+ * This function adds one more argument to the command currently being
+ * enqueued. This function must be called after #rspq_write_begin; it should
+ * be called multiple times (one per argument word), and then #rspq_write_end
+ * should be called to terminate enqueuing the command.
+ * 
+ * See also #rspq_write for a more straightforward API for command enqueuing.
+ * 
+ * @param       w       The write cursor (returned by #rspq_write_begin)
+ * @param       value   New 32-bit argument word to add to the command.
```
anacierdem, line 473: It might be nice to include the first argument restriction we have here as well. I mean the first byte should not be used for the first argument.

What about including the first argument on `rspq_write_begin` instead, to be more consistent with `rspq_write` API? We could get rid of the additional `is_first` state as well. This is not something I'm against but just thinking out loud.

Also do you think we should assert if the first byte of the first arg is not zero?

include/rspq.h
```diff
@@ -151,8 +151,14 @@ extern "C" {
 #endif
 
 /** @brief Maximum size of a command (in 32-bit words). */
-#define RSPQ_MAX_COMMAND_SIZE          16
+#define RSPQ_MAX_COMMAND_SIZE          63
 
+/** @brief Maximum size of a command that it is writable with #rspq_write
```
anacierdem, line 156: This is technically a breaking change if someone is depending on the old name so maybe we can loosen the semantics a little and call the other one `RSPQ_MAX_LONG_COMMAND_SIZE`?

tests/test_rspq.c
```diff
@@ -238,14 +244,16 @@ void test_rspq_flush(TestContext *ctx)
     test_ovl_init();
     DEFER(test_ovl_close());
 
+    // This is meant to verify that the fix in rspq_flush actually
+    // prevents the race condition (see the comment in that function).
+    // If the race condition does happen, this test will fail very quickly.
     uint32_t t0 = TICKS_READ();
-    while (TICKS_DISTANCE(t0, TICKS_READ()) < TICKS_FROM_MS(10000)) {
+    while (TICKS_DISTANCE(t0, TICKS_READ()) < TICKS_FROM_MS(1000)) {
```
anacierdem, line 251: Is 1 sec enough for it to fail consistently? that's a 10 fold decrease from the older value. Is is the randomness that improves it?

tests/rsp_test.S
```diff
@@ -104,3 +109,23 @@ command_reset_log:
     # Reset the big log pointer to the start
     jr ra
     sw zero, %lo(BIG_LOG_PTR)
+
+command_big:
+    addi s1, rspq_dmem_buf_ptr, -128
+    move s2, zero
+command_big_loop:
+    lw t0, %lo(RSPQ_DMEM_BUFFER)(s1)
```
anacierdem, line 117: What about using `CMD_ADDR`?

tests/test_rspq.c
```diff
@@ -685,3 +698,49 @@ void test_rspq_highpri_overlay(TestContext *ctx)
...
+    
+
+    uint32_t output[32] __attribute__((aligned(16)));
+    data_cache_hit_writeback_invalidate(output, 128);
+
+    rspq_write_t wptr = rspq_write_begin(test_ovl_id, 0x8, 33);
+    rspq_write_arg(&wptr, 0);
+    for (uint32_t i = 0; i < 32; i++)
+    {
+        rspq_write_arg(&wptr, i | i << 8 | i << 16 | i << 24);
```
anacierdem, line 722: Why do you prefer to do an xor rather than writing and reading the same arguments?
Also do we depend on the initial state being all zeros when rspq is inited?

anacierdem (commented):
Thank you @rasky! Dropped my comments/questions.

include/rsp_queue.inc
rasky: It becomes public API: it's where each command can read its size, if needed. It can be used to reuse implementation against different-sized commands, or in the future to allow for variable-sized commands.

include/rspq.h
rasky: > It might be nice to include the first argument restriction we have here as well. I mean the first byte should not be used for the first argument.

Sure, added.

> What about including the first argument on rspq_write_begin instead, to be more consistent with rspq_write API? We could get rid of the additional is_first state as well. This is not something I'm against but just thinking out loud.

I think it would make the API less orthogonal: for instance, if you check rdp_triangle in the current gfx-rdp branch (https://github.com/rasky/libdragon/blob/4ff51e9a487a0885a375893827c3e44b510eae56/src/rdpq/rdpq.c#L669) you will see that the first argument is generated within the `__rdpq_write_edge_coeffs` function, so forcing the first argument in rspq_write_begin would make it for a more awkward implementation.

> Also do you think we should assert if the first byte of the first arg is not zero?

If we do it, we should do it also in `rspq_write`. I think I've had a few cases where I actually abused the system already because the assert was not there, but I'm not strongly opposed to adding it. It would have a measurable code-size / performance impact though (unless we limit it to compile-time constants). I think we could discuss it as part of a separate PR anyway.

include/rspq.h
rasky: I think it depends on whether you interpret `RSPQ_MAX_COMMAND_SIZE` as "maximum size of commands that you can write with rspq_write" or "maximum size of commands, in general". I would say the latter is a more obvious interpretation (for somebody that just knows of the current code and never saw this PR), and if that's true, this PR is actually increasing the maximum size of commands in general, by coming up with a new API to handle them.

I'm not opposed to change it, I think I originally had it as you are suggesting, but then I thought this version might be safer.

tests/test_rspq.c
rasky: Yes, the comment says that if we remove the race condition workaround in rspq_flush, the bug triggers very fast, which is also what I've measured. 1s is even still too much for this test.

You could try and verify it yourself, though.

tests/rsp_test.S
rasky: unfortunately that macro only accepts a constant as argument, so you cannot use it in a for loop. If there's a suggestion in this regard, it's well accepted

tests/test_rspq.c
rasky: > Why do you prefer to do an xor rather than writing and reading the same arguments?

No specific reason, just simulating a "do something on RSP".

> Also do we depend on the initial state being all zeros when rspq is inited?

Not that I see?

anacierdem: > Not that I see?

I mean would it behave the same if the initial dram contents were not all 0s? I think we do that when we do `rspq_init` before the test but I was just trying to validate my thinking.

include/rsp_queue.inc
anacierdem: Is it safe to use that register in custom ucode if it doesn't need that information? Still we should document it IMHO.

anacierdem:
Feel free to merge it whenever you think itmis ready!

include/rsp_queue.inc
rasky: Agreed, I've improved the docs

tests/test_rspq.c
rasky: Oh right, but `TEST_BIG` is in `.data` so it is guaranteed to be zero-initialized. It would be different if it was in `.bss`, as in that case it wouldn't be initialized.

PR #286, opened 2022-06-19
Improve installation instructions

rasky:
The instructions did not mention the N64_INST environment variable which
is mandatory for the build system. Moreover, the build-toolchain.sh
was still allowing to be run without it, giving it a default PATH, even
though the build system doesn't know of this default.

Change build-toolchain.sh to make N64_INST mandatory to run, and then
update the README accordingly. Since we are at it, improve also the
instructions by being more detailed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,28 +47,36 @@ programming and debugging. These are the main features:
 See [the libdragon CLI](https://github.com/anacierdem/libdragon-docker) to
 quickly get libdragon up and running. Basically:
 
-1. Download the CLI (as a pre-built binary, or build from source)
-2. Run `libdragon init` to create a skeleton project
-3. Run `libdragon make` to compile a build a ROM
+1. Make sure that you have Docker installed correctly (on Windows and Mac, use
+   Docker Desktop). You can run `docker system info` to check that it is working
+   correctly.
+2. Install the [the libdragon CLI](https://github.com/anacierdem/libdragon-docker).
+   You have two options:
+
+   1. Download the [pre-built binary](https://github.com/anacierdem/libdragon-docker/releases/tag/v10.8.0), 
+      and copy it into some directory which is part of your system PATH.
+   2. If you have `npm` installed (at least verstion 14), run `npm install -g libdragon`.
+3. Run `libdragon init` to create a skeleton project
+4. Run `libdragon make` to compile a build a ROM
 
 If you want, you can also compile and run one of the examples that will
 be found in `libdragon/examples` in the skeleton project.
 
 ### Option 2: Compile the toolchain (Linux only)
 
-1. Create a directory and copy the `build-toolchain.sh` script there from the `tools/` directory.
-2. Read the comments in the build script to see what additional packages are needed.
-3. Run `./build-toolchain.sh` from the created directory, let it build and install the toolchain.
-4. Install libpng-dev if not already installed.
+1. Export the environment variable N64_INST to the path where you want your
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -47,28 +47,37 @@ programming and debugging. These are the main features:
...
+1. Make sure that you have Docker installed correctly (on Windows and Mac, use
+   Docker Desktop). You can run `docker system info` to check that it is working
+   correctly.
+2. Install the [the libdragon CLI](https://github.com/anacierdem/libdragon-docker).
+   You have three options:
+
+   1. Download the [pre-built binary](https://github.com/anacierdem/libdragon-docker/releases/tag/v10.8.0), 
+      and copy it into some directory which is part of your system PATH.
+   2. If you have `npm` installed (at least verstion 14), run `npm install -g libdragon`.
+   3. Build it from source (by running `npm install` after cloning the repository).
```
anacierdem: This might not be ideal to have here, rather it might be better to just provide a link to the original repository or remove this option to reduce confusion.

tools/build-toolchain.sh
```diff
@@ -16,7 +16,7 @@ set -euo pipefail
 IFS=$'\n\t'
 
 # Set N64_INST before calling the script to change the default installation directory path
-INSTALL_PATH="${N64_INST:-/usr/local}"
+INSTALL_PATH="${N64_INST}"
```
anacierdem, line 19: Should we really change the default? There might be setups (potentially automated) depending on the default location. Also what happens if `N64_INST` does not exist in the current environment with this setup?

anacierdem (commented):
Looks much better, thank you! I have left a few comments. Also I'm planning to improve the instructions for PATH setup on the cli readme, which will further help with the process.

tools/build-toolchain.sh
rasky: > Should we really change the default? There might be setups (potentially automated) depending on the default location.

I don't think there are many CIs that are building the toolchain via this script. The point is that everything you do with libdragon *after* building the toolchain requires this variable to be set. So maybe it is better to ask the user to set it before building the toolchain, so they don't get confused. Also, this gives explicit visibility to the fact that you can change the installation directory (and it has to be declared before building -- you cannot change it after building).

> what happens if N64_INST does not exist in the current environment with this setup?

The script aborts with an error because we are in strict mode:
> build-toolchain.sh: line 19: N64_INST: unbound variable

README.md
rasky: OK do you prefer to remove the 3 or leave it as "build from the source" but just sends to the repo instead of hinting the command?

anacierdem: Let's remove it

tools/build-toolchain.sh
anacierdem: Looks good!

Hazematman: I forgot Make's rules, will this error if N64_INST is not set? If not could we do something like
```Makefile
ifndef N64_INST
$(error Please set the environment variable N64_INST to the location to install the tool chain. Check README.md for more details)
endif
```

rasky: We could do that but it would have to be duplicated in all Makefiles of all libdragon projects. In general I would say the Makefile is under the project responsibility, so we’re not helping in solving the problem to new users, which might hit the problem while trying to compile eg Flappy Bird

README.md
rasky: Done

PR #287, opened 2022-06-24
Two small fixes to n64.mk for RSP builds

rasky:


anacierdem:
I think this was not onto the latest trunk. Can you verify you are in sync and all the changes are intended?

rasky:
Looking at the diff, it looks OK to me. What changes seem unintended?

anacierdem:
There are diffs on `audio.c` on the lines you have inteoduced the AI h/w fix. They seem to be reverted. Maybe there is something wrong with the GitHub app if that was not intended.

anacierdem:
Commits look ok must be something weird with the app. I will check on web later.

PR #288, opened 2022-07-05
Explicitly set C++ version in cpptest example

sp1187:
Also modernize cpptest code a bit to use smart pointers and make_unique in order to showcase modern C++ support.

### DIFF
--- a/n64.mk
+++ b/n64.mk
@@ -52,7 +52,7 @@ CFLAGS+=-MMD     # automatic .d dependency generationc
 CXXFLAGS+=-MMD     # automatic .d dependency generationc
 ASFLAGS+=-MMD    # automatic .d dependency generation
 
-N64_CXXFLAGS := $(N64_CFLAGS) -std=c++11
+N64_CXXFLAGS := $(N64_CFLAGS)
 N64_CFLAGS += -std=gnu99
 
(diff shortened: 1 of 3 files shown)


n64.mk
```diff
@@ -52,7 +52,7 @@ CFLAGS+=-MMD     # automatic .d dependency generationc
 CXXFLAGS+=-MMD     # automatic .d dependency generationc
 ASFLAGS+=-MMD    # automatic .d dependency generation
 
-N64_CXXFLAGS := $(N64_CFLAGS) -std=c++11
```
anacierdem, line 55: There might be existing code depending on this. I think we can change it in a way that the consumer can override without changing the default

anacierdem (changes requested):

sp1187:
The default since GCC 11 is C++17 anyway, and C++17 should in most (practically all except if you are using features that were already deprecated in C++11) cases be backwards compatible with C++11 anyway, so I'm not sure if that's necessary. Old makefiles could in rare cases change their flag settings to be explicitly C++11 if necessary.

n64.mk
rasky: C++11 has been the default since GCC 4.8.1, and C++14 has been default since GCC 6. So what this did was actually downgrading the default. I don't think it's going to break anything, and the fix is trivial anyway.

PR #289, opened 2022-07-08
Add support for building toolchain with macOS (plus other fixes)

rasky:
This PR adds support for building the toolchain via macOS. It also includes several tweaks and incremental improvements to our build scripts to make them more solid in common scenarios.


tools/toolchain/build-toolchain.sh
```diff
@@ -15,6 +15,13 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+# Check that N64_INST is defined
+if [ -z "${N64_INST-}" ]; then
+  echo "N64_INST environemnt variable is not defined."
```
vieux:
```suggestion
  echo "N64_INST environment variable is not defined."
```

vieux:
@rasky where you able to try on intel and/or M1 macs?

tools/toolchain/build-toolchain.sh
rasky: Done, thanks

rasky:
@vieux yes, I tested this on a M1 Mac. It is able to correctly generate a working toolchain, install libdragon and compile everything. I also tested it with my prefix (`/opt/n64`) owned by my user or by root, and in both cases everything works.

build.sh
```diff
@@ -4,12 +4,29 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+if [[ -z ${N64_INST-} ]]; then
+  echo N64_INST environment variable is not defined
+  echo Please set N64_INST to point to your libdragon toolchain directory
+  exit 1
+fi
```
danbolt, line 11: :clap: :clap: :clap:

meeq:
Minor breakage note: FlappyBird's Makefile builds the toolchain in-project if it doesn't already exist by copying the `build-toolchain.sh` script out of the tools directory: https://github.com/meeq/FlappyBird-N64/blob/ea05f08e9cdfd4318fe99b321efd74b428995923/Makefile#L224

Other external scripts that depend on the location of this file may also break now that it has moved into a sub-directory.

rasky:
An alternative would be to leave the script where it is, and change it so that I creates an empty directory itself and runs from there. It doesn't have to be a temp directory, it could simply be a `build-toolchain/` subdirectory (that we could pre-commit to gitignore).

Would this be better?

anacierdem (approved):
Looks good to me. Expecting the user to have `brew` is totally ok as well, thanks for making this possible!

rasky:
I've changed the build script as suggested. It is not moved anymore from its original location, and just builds automatically into a subdirectory (that can be configured). After the tests pass, I'll merge this.

PR #290, opened 2022-07-12
Build Docker images for linux/arm64 as well

rasky:
While Macbooks with M1 can run linux/amd64 images, it is unbearably
slow. Building and publishing a linux/arm64 image provides a much
better experience.

### DIFF
--- a/.github/workflows/build-toolchain-library-and-roms.yml
+++ b/.github/workflows/build-toolchain-library-and-roms.yml
@@ -51,6 +51,9 @@ jobs:
 
       # Build the toolchain if toolchain files changed w.r.t target and we can
       # use from registry o/w
+      - name: Set up QEMU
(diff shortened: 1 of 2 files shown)


vieux:
Looks like a custom builder might help https://github.com/docker/buildx/issues/495#issuecomment-761562905

rasky:
@vieux thanks, I'm trying with a suggestion from that thread

rasky:
Unfortunately, just like using docker images for amd64 is too slow on Apple M1, running docker arm64 on a amd64 server (GitHub action) is just as slow, and the CI hits the hard limit of 6 hours per job.

I think for now I’ll just see if I can push the ARM image manually into the registry. For CI, we will have to either wait for GitHub to provide native ARM runners, or to cross-compile the ARM->MIPS compiler from amd64, which is a complex modification of the build script. Either way, this PR won’t fly for now.

PR #291, opened 2022-07-15
interrupt: add support for CART interrupts

rasky:
This PR allows to enable the CART interrupt and connect handlers to it. This is used by 64DD but in general peripherals on the cartridge bus can trigger it.

### DIFF
--- a/src/interrupt.c
+++ b/src/interrupt.c
@@ -152,12 +152,16 @@ struct callback_link * VI_callback = 0;
 struct callback_link * PI_callback = 0;
 /** @brief Linked list of DP callbacks */
 struct callback_link * DP_callback = 0;
-/** @brief Linked list of TI callbacks */
-struct callback_link * TI_callback = 0;
 /** @brief Linked list of SI callbacks */
 struct callback_link * SI_callback = 0;
 /** @brief Linked list of SP callbacks */
 struct callback_link * SP_callback = 0;
+/** @brief Linked list of TI callbacks */
+struct callback_link * TI_callback = 0;
+/** @brief Linked list of CART callbacks */
+struct callback_link * CART_callback = 0;
+
+static int last_cart_interrupt_count = 0;
 
 /** 
  * @brief Call each callback in a linked list of callbacks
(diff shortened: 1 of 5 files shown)


src/interrupt.c
```diff
@@ -472,11 +473,84 @@ void unregister_SP_handler( void (*callback)() )
...
+/**
+ * @brief Register a timer callback
+ * 
+ * The callback will be used when the timer interrupt is triggered by the CPU.
+ * This happens when the COP0 COUNT register reaches the same value of the
+ * COP0 COMPARE register.
+ * 
+ * This function is useful only if you want to do your own low level programming
+ * of the internal CPU timer and handle the interrupt yourself. In this case,
+ * also remember to activate the timer interrupt using
```
anacierdem: The end of the sentence is missing. Thanks for the additional documentation!

src/interrupt.c
```diff
@@ -472,11 +473,84 @@ void unregister_SP_handler( void (*callback)() )
...
+
+/**
+ * @brief Register a timer callback
+ * 
+ * The callback will be used when the timer interrupt is triggered by the CPU.
+ * This happens when the COP0 COUNT register reaches the same value of the
+ * COP0 COMPARE register.
+ * 
+ * This function is useful only if you want to do your own low level programming
+ * of the internal CPU timer and handle the interrupt yourself. In this case,
```
anacierdem, line 487: Is it possible to disable the mechanism we have at inthandler.S?

src/interrupt.c
```diff
@@ -314,10 +318,29 @@ void __MI_handler(void)
...
+void __CART_handler(void)
+{
+    /* Call the registered callbacks */
+    __call_callback(CART_callback);
+
+     /* CART interrupts must be acknowledged by handlers. If the handler fails
+       to do so, the console freezes because the interrupt will retrigger
+       continuously. Since a freeze is always bad for debugging, try to 
+       detect it, and show a proper assertion screen. */
+    if (!(C0_CAUSE() & C0_INTERRUPT_CART))
```
anacierdem, line 338: Should we also keep these behind `#ifndef NDEBUG` to reduce a little bit of code at final builds?

anacierdem (approved):
LGTM. Only minor comments.

src/interrupt.c
rasky: Which one?

anacierdem: This was a comment I forgot to delete. I meant how the users would not use the library version but instead implement their own. Then I realized, they just don't use the library timers...

src/interrupt.c
rasky: Done

src/interrupt.c
rasky: Sure, done thanks.

PR #294, opened 2022-08-20
Add infrastructure to handle reset button process

rasky:
Currently, libdragon just shows a crash screen as soon as the reset button is pressed.

This PR implements a general infrastructure to handle reset button handling:

  * Ability to register a handler when the reset is triggered (register_reset_handler)
  * Ability to just poll for the reset button (exception_reset_time)

Moreover, it hooks this up in the audio system as follows:

 * audio.c: avoid scheduling new buffers if they'd be truncated by the reset signal
 * mixer.c: perform a fadeout during reset

### DIFF
--- a/include/exception.h
+++ b/include/exception.h
@@ -105,6 +105,21 @@ typedef struct
     volatile reg_block_t* regs;
 } exception_t;
 
+
+/** 
+ * @brief Guaranteed length of the reset time.
+ * 
+ * This is the guaranteed length of the reset time, that is the time
+ * that goes between the user pressing the reset button, and the CPU actually
+ * resetting. See #exception_reset_time for more details.
+ * 
+ * @note The general knowledge about this is that the reset time should be
+ *       500 ms. Testing on different consoles show that, while most seem to
+ *       reset after 500 ms, a few EU models reset after 200ms. So we define
+ *       the timer shorter for greater compatibility.
+ */
+#define RESET_TIME_LENGTH      TICKS_FROM_MS(200)
+
 /** @} */
 
 #ifdef __cplusplus
@@ -113,6 +128,9 @@ extern "C" {
 
 void register_exception_handler( void (*cb)(exception_t *) );
 void exception_default_handler( exception_t* ex );
+
+void register_reset_handler( void (*cb)(void) );
+uint32_t exception_reset_time( void );
 
 #ifdef __cplusplus
 }
--- a/src/audio/mixer.c
+++ b/src/audio/mixer.c
@@ -13,6 +13,7 @@
 #include "samplebuffer.h"
 #include "audio.h"
 #include "n64sys.h"
+#include "exception.h"
 #include <memory.h>
 #include <stdlib.h>
 #include <math.h>
@@ -631,11 +632,23 @@ static void mixer_exec(int32_t *out, int num_samples) {
 		settings->rvol[ch] = rvol32[ch];
 	}
 
+	// Check if we the user pressed RESET. If so, we can apply
+	// a simple global volume ramp to fade out the volume.
+	// This is just a user-level feature. audio.c will truncate
+	// DMA transfers to AI anyway.
+	float gvol = Mixer.vol;
+	uint32_t reset_time = exception_reset_time();
+	if (reset_time) {
+		const float FADE_OUT_TIME = (float)RESET_TIME_LENGTH / TICKS_PER_SECOND;
+		float elapsed = (float)reset_time / TICKS_PER_SECOND;
+		gvol *= (FADE_OUT_TIME - MIN(elapsed, FADE_OUT_TIME)) / FADE_OUT_TIME;
+	}
+
 	uint32_t t0 = TICKS_READ();
 
 	rspq_highpri_begin();
 	rspq_write(__mixer_overlay_id, 0,
-		(((uint32_t)MIXER_FX16(Mixer.vol)) & 0xFFFF),
+		(((uint32_t)MIXER_FX16(gvol)) & 0xFFFF),
 		(num_samples << 16) | Mixer.num_channels,
 		PhysicalAddr(out),
 		PhysicalAddr(&Mixer.ucode_settings));
--- a/src/exception.c
+++ b/src/exception.c
@@ -6,7 +6,8 @@
 #include "exception.h"
 #include "console.h"
 #include "n64sys.h"
-
+#include "debug.h"
+#include "regsinternal.h"
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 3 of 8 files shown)


include/exception.h
```diff
@@ -105,6 +105,21 @@ typedef struct
...
+/** 
+ * @brief Guaranteed length of the reset time.
+ * 
+ * This is the guaranteed length of the reset time, that is the time
+ * that goes between the user pressing the reset button, and the CPU actually
+ * resetting. See #exception_reset_time for more details.
+ * 
+ * @note The general knowledge about this is that the reset time should be
+ *       500 ms. Testing on different consoles show that, while most seem to
+ *       reset after 500 ms, a few EU models reset after 200ms. So we define
```
anacierdem, line 118: Is the timer on the console or the cartridge? I remember a discussion about it being implemented on the CIC chip. If your dev cartridges are using 3rd party chips that implement the timer differently, it might be only like this for your carts.

rasky: I think this is pending further investigations, there are many question marks. I myself don't fully believe this but it is what has been measured. 

This PR has been pending these investigations for a long time, and I don't know when they will be done. Meanwhile, the reset handling on trunk is broken (we show an exception screen).

My suggestion is to go through with this as-is and then we can change this specific macro once we fully understand the situation. Meanwhile we can grow code that does the right thing, using this macro for calculations so that it will adapt to new discoveries.

If you prefer I can change this to 500 and ignore the findings until we better frame them.

anacierdem: Anything under 500ms is safe so we can continue as is 👍🏼 I can review the code in general soon but on a quick glance it all looks good.

rasky:
Followup required to better handle reset:

 * Stop rumble (or the rumble will stay on after reset)
 * Make sure VI Y scaling is set to 0x400 (or the VI will not work after reset)

src/exception.c
```diff
@@ -281,16 +289,97 @@ void __onCriticalException(volatile reg_block_t* regs)
...
+}
 
-	__fetch_regs(&e, EXCEPTION_TYPE_RESET, regs);
-	__exception_handler(&e);
+/** 
+ * @brief Check whether the RESET button was pressed and how long we are into
+ *        the reset process.
+ * 
+ * This function returns how many ticks have elapsed since the user has pressed
+ * the RESET button, or 0 if the user has not pressed it.
```
anacierdem, line 337: Does the counter start when the button is pressed or released? I remember you provided a test rom at some point but I don't remember the behavior RN. Maybe we can include that small piece of information as well.

src/exception.c
```diff
@@ -281,16 +289,97 @@ void __onCriticalException(volatile reg_block_t* regs)
...
+ * 
+ * It can be used by user code to perform actions during the RESET
+ * process (see #register_reset_handler). It is also possible to simply
+ * poll this value to check at any time if the button has been pressed or not.
```
anacierdem, line 341: In that case we would expect a larger than 0 value right? Being able to call this without an actual reset feels a little weird to me. Would making the initial value -1 (thus changing this to an int) make sense to detect via polling? Or maybe I'm overthinking...

src/exception.c
```diff
@@ -281,16 +289,97 @@ void __onCriticalException(volatile reg_block_t* regs)
...
+ * @brief Respond to a reset exception.
+ * 
+ * Calls the handlers registered by #register_reset_handler.
+ */
+void __onResetException( volatile reg_block_t* regs )
+{
+	/* This function will be called many times becuase there is no way
+	   to acknowledge the pre-NMI interrupt. So make sure it does nothing
+	   after the first call. */
+	if (__prenmi_tick) return;
```
anacierdem, line 377: It is theoretically possible to write a 0 ticks so this may execute the callbacks more than one. Not very important or critical but maybe another reason for a negative initial value?

src/inthandler.S
```diff
@@ -177,13 +177,21 @@ interrupt:
...
+	# stay pending in CR. Let's disable it in SR to avoid
+	# looping here. If another unrelated interrupt triggers, 
+	# CR will still have 0x1000 set, but __onResetException will
+	# do nothing after the first call.
+	li t0, ~0x1000
+	lw t1, STACK_SR(sp)
+	and t1, t0
+	sw t1, STACK_SR(sp)
+
+	# Reload cause register and test for other interrupts
```
anacierdem: Why we need a reload now? Honestly I didn't get the later ones as well.

src/audio/mixer.c
```diff
@@ -631,11 +632,23 @@ static void mixer_exec(int32_t *out, int num_samples) {
...
+	// Check if we the user pressed RESET. If so, we can apply
+	// a simple global volume ramp to fade out the volume.
+	// This is just a user-level feature. audio.c will truncate
+	// DMA transfers to AI anyway.
+	float gvol = Mixer.vol;
+	uint32_t reset_time = exception_reset_time();
+	if (reset_time) {
+		const float FADE_OUT_TIME = (float)RESET_TIME_LENGTH / TICKS_PER_SECOND;
+		float elapsed = (float)reset_time / TICKS_PER_SECOND;
+		gvol *= (FADE_OUT_TIME - MIN(elapsed, FADE_OUT_TIME)) * (1.0f / FADE_OUT_TIME);
```
anacierdem: This maybe a little clearer?
```gvol *= (FADE_OUT_TIME - MIN(elapsed, FADE_OUT_TIME)) / FADE_OUT_TIME;```

anacierdem (commented):
Looks good to me, just a few questions primarily for clarification.

src/exception.c
rasky: Done

src/exception.c
rasky: Yeah I know that -1 feels "a bit more" out of band than 0, but practically speaking 0 is out of band too (there is no way to observe a 0 value returned by this function after RESET was pressed), and 0 value is false in C which means that code is a bit more readable IMO.

src/exception.c
rasky: This is a bit difference. I can see the value -1 as return value for exception_reset_time, but __prenmi_tick is a uint32_t returned by TICKS_READ so it cannot be negative, it would not work with macros like TICKS_DISTANCE. 

You have a point that there is 1 possibility in 4 billions that libdragon reads the COP0 count when it is exactly 0, and thus misses the reset event. I have fixed this now by always forcing the last bit to 1 when reading. As a side-effect, this will decrease the reset time by 0.5 cycles on average which has no practical effect.

src/inthandler.S
rasky: It's part of the commit https://github.com/DragonMinded/libdragon/pull/294/commits/a17748a40d177e30aa3eef8af899fe390c544fef. 

There is an explanation there. Basically, on trunk we currently process one interrupt per exception. That is, when an exception fires, we look for which interrupt has triggered; as soon as we find one interrupt, we call the handler, and then exit the exception. If another interrupt was pending, a new exception will immediately retrigger.

This is acceptable for the rare case of two semi-simultaneous interrupts, but once we activate the pre-NMI interrrupt it doesn't work anymore: in fact, pre-NMI cannot be acked and will always trigger. This means that, if the exception handler finds pre-NMI set, process it, and then exit, it will enter a deadlock where an exception is continuously generated.

So we change the exception handler to process all interrupts that it finds set. To do so, after calling a handler, we need to reload the COP0 cause register as the register in which it was stored might have been reused by the handler itself.

src/audio/mixer.c
rasky: Done.

src/inthandler.S
rasky: BTW this is the most scary part of the whole PR, make sure it makes sense :)

danbolt:
I remember noticing the crash screen when issuing an NMI, so this is nice to see!

PR #295, opened 2022-08-22
Batch of small changes

rasky:
Various small changes from the incoming rdpq branch. Flush a bit of stuff to simplify review later.

### DIFF
--- a/doxygen-public.conf
+++ b/doxygen-public.conf
@@ -628,7 +628,7 @@ INLINE_INFO            = YES
 # name. If set to NO, the members will appear in declaration order.
 # The default value is: YES.
 
-SORT_MEMBER_DOCS       = YES
+SORT_MEMBER_DOCS       = NO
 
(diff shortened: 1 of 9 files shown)


doxygen-public.conf
```diff
@@ -628,7 +628,7 @@ INLINE_INFO            = YES
 # name. If set to NO, the members will appear in declaration order.
 # The default value is: YES.
 
-SORT_MEMBER_DOCS       = YES
+SORT_MEMBER_DOCS       = NO
```
anacierdem, line 631: Is our intended order more appropriate?

anacierdem (approved):
Looks good to me, thanks!

doxygen-public.conf
rasky: Yes in general most (all?) header files list functions in some rationale order, and we can correct those which don't. Alphabetical order doesn't seem very useful for a digital format where users will CTRL+F anyway.

PR #296, opened 2022-08-30
Clean up a bit dma.c on physical vs virtual and PI address space

rasky:
This PR contains 4 commits which globally cleans up a bit of confusion in dma.c. In particular:

 * PI bus has a full 32-bit address space, part of which is accessible via direct CPU I/O. This PR clarifies this a bit in the documentation (also adding a io_accessible macro), and also fixes dma_read_async to do misalign fixups only for accessible ranges.
 * dma_read / dma_write has always clamped their address into the ROM area (since 2012). Unfortunately this doesn't allow to use them for the full 32-bit address space. To fix this while striving to keep backward compatibility, we do the clamping in dma_read / dma_write and leave dma_read_async / dma_write_async working on the full 32-bit address range. This is technically a functional breakage of the async functions that were introduced in Sep 2021, and we're just hoping nobody has tried using them with "wrong addresses" given the limited time passed since their introduction.
 * We are trying to move libdragon to use C pointers for virtual addresses, and `uint32_t` for physical addresses. This PR moves forward in that direction.

### DIFF
--- a/src/dma.c
+++ b/src/dma.c
@@ -47,6 +47,54 @@ static volatile int __dma_busy(void)
     return PI_regs->status & (PI_STATUS_DMA_BUSY | PI_STATUS_IO_BUSY);
 }
 
+/**
+ * @brief Check whether the specified PI address can be accessed doing I/O from CPU
+ * 
+ * The PI bus covers the full 32-bit address range. The full range is only accessible
+ * via DMA, though. A part of the range is also memory mapped to the CPU and can be
+ * accessed via #io_read and #io_write.
+ * 
+ * The ranges of PI address that can be accessed via CPU are:
+ * 
+ *  * 0x0500_0000 - 0x0FFF_FFFF: used by N64DD and SRAM on cartridge
+ *  * 0x1000_0000 - 0x1FBF_FFFF: cartridge ROM
+ *  * 0x1FD0_0000 - 0x1FFF_FFFF: no known PI peripherals use this
+ * 
+ * The rest of the 32-bit address range is only accessible via DMA.
+ * 
+ * Notice also that the range 0x2000_0000 - 0x7FFF_FFFF is theoretically accessible
+ * by the CPU but only via 64-bit addressing, so it requires assembly instructions
+ * (as the libdragon toolchain uses 32-bit pointers). No known PI peripherals use this
+ * range anyway.
+ * 
+ * This function checks whether the specified address falls into the range accessible
+ * via CPU or not.
+ * 
+ * @param pi_address        PI address to check
+ * @return                  True if the address is memory mapped, false if it is not
+ */
+bool io_accessible(uint32_t pi_address)
+{
+    // Below 0x0500_0000, there is RDRAM and RCP registers.
+    if (pi_address < 0x05000000)
+        return false;
+
+    // Using 32-bit addresses (like those available from within our C code),
+    // the CPU can only access addresses up to 0x1FFF_FFFF. 
+    // FIXME: we could theoretically lift this limit up to 0x7FFF_FFFF if the
+    // I/O functions were rewritten in assembly using 64-bit addresses, but
+    // there is currently no known PI peripheral that operates in that range anyway.
+    if (pi_address > 0x1FFFFFFF)
+        return false;
+
+    // The SI bus is partially covering the PI range in the CPU memory map
+    if (pi_address >= 0x1FC00000 && pi_address <= 0x1FCFFFFF)
+        return false;    
+
+    // All other addresses are memory mapped and can be accessed via CPU.
+    return true;
+}
+
 /** 
  * @brief Return whether the DMA controller is currently busy
  *
@@ -364,6 +412,11 @@ void dma_write(const void * ram_address, unsigned long rom_address, unsigned lon
  *            Memory address of the peripheral to read from
  *
  * @return The 32 bit value read from the peripheral
+ * 
+ * @note This function only works if the specified PI address falls within a range
+ *       which is memory mapped on the CPU. See #io_accessible for more information.
+ * 
+ * @see #io_accessible
  */
 uint32_t io_read(uint32_t pi_address)
 {
@@ -387,6 +440,11 @@ uint32_t io_read(uint32_t pi_address)
  *            Memory address of the peripheral to write to
  * @param[in] data
  *            32 bit value to write to peripheral
+ *
+ * @note This function only works if the specified PI address falls within a range
+ *       which is memory mapped on the CPU. See #io_accessible for more information.
+ *
+ * @see #io_accessible
  */
 void io_write(uint32_t pi_address, uint32_t data) 
 {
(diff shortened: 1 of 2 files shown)


src/dma.c
```diff
@@ -47,6 +47,54 @@ static volatile int __dma_busy(void)
...
 }
 
+/**
+ * @brief Check whether the specified PI address can be accessed doing I/O from CPU
+ * 
+ * The PI bus covers the full 32-bit address range. The full range is only accessible
+ * via DMA, though. A part of the range is also memory mapped to the CPU and can be
+ * accessed via #io_read and #io_write.
+ * 
+ * The ranges of PI address that can be access via CPU are:
```
anacierdem: nit: `accessed`

src/dma.c
```diff
@@ -47,6 +47,54 @@ static volatile int __dma_busy(void)
...
+ * 
+ *  * 0x0500_0000 - 0x0FFF_FFFF: used by N64DD and SRAM on cartridge
+ *  * 0x1000_0000 - 0x1FBF_FFFF: cartridge ROM
+ *  * 0x1FD0_0000 - 0x1FFF_FFFF: no known PI peripherals use this
+ * 
+ * The rest of the 32-bit address range is only accessible via DMA.
+ * 
+ * Notice also that the range 0x2000_0000 - 0x7FFF_FFFF is theoretically accessible
```
anacierdem, line 65: Just a question: why is it only accessible via 64 bit addressing?

src/dma.c
```diff
@@ -135,17 +183,17 @@ void dma_write_raw_async(const void * ram_address, unsigned long pi_address, uns
  * 
  * @note This function must be called with interrupts disabled.
  */
-static uint32_t __io_read32(unsigned long pi_address) {
+static uint32_t __io_read32(void* pi_pointer) {
```
anacierdem, line 186: So this means whenever we access something from the CPU it should be a virtual address?

src/dma.c
```diff
@@ -47,6 +47,54 @@ static volatile int __dma_busy(void)
...
+ * (as the libdragon toolchain uses 32-bit pointers). No known PI peripherals use this
+ * range anyway.
+ * 
+ * This function checks whether the specified address falls into the range accessible
+ * via CPU or not.
+ * 
+ * @param pi_address        PI address to check
+ * @return                  True if the address is memory mapped, false if it is not
+ */
+bool io_accessible(uint32_t pi_address)
```
anacierdem, line 76: What about `io_cpu_accessible`? I think it is less confusing. Otherwise people might think it is doing something else just by looking at the name.

src/dma.c
```diff
@@ -217,27 +267,39 @@ static uint8_t __io_read8(unsigned long pi_address) {
...
+    uint32_t ram_address = (uint32_t)ram;
 
     assert(len > 0);
-    assert(((ram ^ rom) & 1) == 0); 
+    assert(((ram_address ^ pi_address) & 1) == 0); 
 
     disable_interrupts();
+
+    // Check if the PI address can be accessed with CPU.
+    // If so, we cannot perform a misaligned transfer.
```
anacierdem: shouldn't be the other way around? `if not, we cannot perform...`

src/dma.c
```diff
@@ -316,45 +398,64 @@ void dma_read(void *ram_address, unsigned long pi_address, unsigned long len)
...
+ * @note This function has always had an historical mistake: the pi_address is mangled
+ *       to be forced into the ROM area (0x10000000-0x1FFFFFFF). This is wrong as the
+ *       PI bus has full 32-bit address, and the same function could have been used
+ *       to access the whole range.
+ *       If you need to read outside the ROM area, use #dma_write_raw_async instead.
  */
-void dma_write(const void * ram_address, unsigned long pi_address, unsigned long len) 
+void dma_write(const void * ram_address, unsigned long rom_address, unsigned long len) 
 {
-    assert(len > 0);
```
anacierdem, line 326: Should we remove the assertion as well?

src/dma.c
```diff
@@ -316,45 +398,64 @@ void dma_read(void *ram_address, unsigned long pi_address, unsigned long len)
...
- * @brief Read a 32 bit integer from a peripheral
+ * @brief Read a 32 bit integer from a peripheral using the CPU.
  *
  * @param[in] pi_address
  *            Memory address of the peripheral to read from
  *
  * @return The 32 bit value read from the peripheral
+ * 
+ * @note This function only works if the specified PI address falls within a range
+ *       which is memory mapped on the CPU. See #io_accessible for more information.
```
anacierdem, line 428: Maybe we can assert if it is accessible? Would it be too much a hit for performance on a debug build?

anacierdem (commented):
Thank you for the consistency improvements @rasky, a few minor comments here and there...

src/dma.c
rasky: Fixed

src/dma.c
rasky: In 32-bit mode, addresses are made of 29-bit address + 3 bit of segment. The direct-mapped segments like what we call CachedAddr / UncachedAddr only allow access to physical address up to 29 bits, that is 0x1FFF_FFFF. To access addresses in the full 32-bit physical range you either need TLBs, or 64-bit addressing.

src/dma.c
rasky: Yes, CPU only access virtual addresses by definition. A virtual address is 3 bit segment + 29 bit address, and is mapped to a physical address either directly or indirectly via TLB.
See also: https://n64brew.dev/wiki/Memory_map

src/dma.c
rasky: The rationale here is to stay similar to `io_read` / `io_write`. I can add the `cpu` work in there if you feel it's better, even though it's missing from `io_read` / `io_write`.

src/dma.c
rasky: Fixed.

src/dma.c
rasky: The assert is pertinent, but it's duplicated at the top of `dma_write_raw_async` which is called immediately after this line. This is why I removed it.

src/dma.c
rasky: I preferred not to for performance reasons. There are rationales here:

 * Currently there are no known PI peripherals that operate outside of the accessible areas. I don't know what PicoCart64 is going to do for instance, but I guess they'd rather fall within the range for flexibility. So `pi_accessible` is mostly for documentation purposes, and checking `pi_accessible` in `dma_read_async` is just... uhm.. pedantry. The assert in io_read would never trigger (if not for programmer bugs, that would trigger anyway in other forms).
 * Normally `io_read` is called with some "fixed" / "hardcoded" address. Outside of library code, It is very uncommon to call it with an address which is "unknown" at compile time. So the address would either be correct (and the code will work) or wrong because eg a typo during development (and the code will crash/miswork right away). It doesn't seem an assert is adding much here.
 * We don't have a real story for non-debug builds, especially for disabling NDEBUG in libdragon. One has to recompile everything and know what to do. n64.mk doesn't help at the moment. I wouldn't be surprised if most homebrew is released with NDEBUG on.

So it's a bit on the edge, but the check is a bit slow compared to io_read itself which is a memory op, and I don't think it's helping much.

anacierdem: I agree, except I use NDEBUG for my final builds :)

src/dma.c
anacierdem: looks good as is for consistency 👍

PR #304, opened 2022-09-06
Updated references to m64p to simple64

thekovic:
m64p was rebranded as simple64 so I updated all links and references to it in the comments and the README accordingly.


rasky:
Thanks

PR #306, opened 2022-09-13
build-toolchain.sh: refactoring to support canadian cross

rasky:
This refactor prepares for building a native Windows toolchain, and also an ARM64 toolchain from CI, by optionally allowing for a canadian cross.

For the standard case of a cross-compiler, there are two changes:
 
 * it builds (much) faster than before, by building GCC just once (before, it was being built twice but it's not clear why that was deemed necessary).
 * it removes the hard dependency on GMP/MPFR/MPC, by building them from source. It will still use an installed version if found, but the fallback now still works, rather than erroring out. This also simplifies build instructions.

Co-authored-by: Robin Jones

### DIFF
--- a/tools/build-toolchain.sh
+++ b/tools/build-toolchain.sh
@@ -3,27 +3,24 @@
 # (c) 2012-2021 DragonMinded and libDragon Contributors.
 # See the root folder for license information.
 
-# Before calling this script, make sure you have GMP, MPFR and TexInfo
-# packages installed in your system.  On a Debian-based system this is
-# achieved by typing the following commands:
-#
-# sudo apt-get install libmpfr-dev
-# sudo apt-get install texinfo
-# sudo apt-get install libmpc-dev
-
 # Bash strict mode http://redsymbol.net/articles/unofficial-bash-strict-mode/
 set -euo pipefail
 IFS=$'\n\t'
 
 # Check that N64_INST is defined
 if [ -z "${N64_INST-}" ]; then
-  echo "N64_INST environment variable is not defined."
-  echo "Please define N64_INST and point it to the requested installation directory"
-  exit 1
+    echo "N64_INST environment variable is not defined."
+    echo "Please define N64_INST and point it to the requested installation directory"
+    exit 1
 fi
 
 # Path where the toolchain will be built.
 BUILD_PATH="${BUILD_PATH:-toolchain}"
+
+# Defines the build system variables to allow cross compilation.
+BUILD=${BUILD:-""}
+HOST=${HOST:-""}
+TARGET=${TARGET:-mips64-elf}
 
 # Set N64_INST before calling the script to change the default installation directory path
 INSTALL_PATH="${N64_INST}"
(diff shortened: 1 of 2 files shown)


rasky:
Not sure why it fails on Linux... it works on my Mac. I will try to reproduce.

networkfusion:
From the build output, it looks like the docker container requires an env var...

tools/build-toolchain.sh
```diff
@@ -34,139 +31,273 @@ export PATH="$PATH:$INSTALL_PATH/bin"
...
-  (command -v "$1" >/dev/null 2>&1)
-  return $?
+    (command -v "$1" >/dev/null 2>&1)
+    return $?
 }
 
 # Download the file URL using wget or curl (depending on which is installed)
 download () {
-  if   command_exists wget ; then wget -c  "$1"
-  elif command_exists curl ; then curl -LO "$1"
```
networkfusion, line 54: Perhaps this should be ` elif command_exists curl ; then curl -LJO "$1"`
Extra option `-J` is "--remote-header-name Use the header-provided filename". This can sometimes "fix" compatibility issues (at least with other projects I am working on).

tools/build-toolchain.sh
```diff
@@ -34,139 +31,273 @@ export PATH="$PATH:$INSTALL_PATH/bin"
...
-  else
-    echo "Install wget or curl to download toolchain sources" 1>&2
-    return 1
-  fi
+    if   command_exists wget ; then wget -c  "$1"
+    elif command_exists curl ; then curl -LO "$1"
```
networkfusion, line 55: Why is this not using the "download" function??? as above...

tools/build-toolchain.sh
```diff
@@ -34,139 +31,273 @@ export PATH="$PATH:$INSTALL_PATH/bin"
...
 fi
 
 # Create build path and enter it
 mkdir -p "$BUILD_PATH"
 cd "$BUILD_PATH"
 
-# Dependency source: Download stage
+# Dependency downloads and unpack
 test -f "binutils-$BINUTILS_V.tar.gz" || download "https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILS_V.tar.gz"
+test -d "binutils-$BINUTILS_V"        || tar -xzf "binutils-$BINUTILS_V.tar.gz"
```
networkfusion, line 93: Perhaps use `-C` to unpack to expected directory... (at least for debug purposes...)

tools/build-toolchain.sh
rasky: `-J` breaks the assumption that if I ask a file called `URL/foo.tar.gz`, the downloaded file (if any) is called `foo.tar.gz`. In other words, if there was a case where `-J` is required, we would still not be able to open the file afterwards. So I don't think that's useful here.

tools/build-toolchain.sh
rasky: Not sure what you mean, what you mentioned *is* the implementation of the download function.

PR #307, opened 2022-09-14
rspq: disable INTR_BREAK without race conditions

rasky:
Currently, rspq starts the RSP via rsp_run_async which sets INTR_BREAK (and we can't change that), and then disable INTR_BREAK one split second later.

This can cause race conditions, especially on emulators, given that the RSP will actually hit the break just a few opcodes later, and if an interrupt is generated, it hits an assert because it is unexpected.

Fix it by avoiding the race window and disabling INTR_BREAK at start. We do that with an internal API for now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,9 +118,15 @@ clean:
 	rm -f *.o *.a
 	rm -rf $(CURDIR)/build
 
-clobber: clean doxygen-clean examples-clean tools-clean
+test:
+	$(MAKE) -C tests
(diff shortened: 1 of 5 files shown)


rasky:
I snooped another commit in here which add a test makefile target, so that `make clobber` also clean tests. Otherwise `build.sh` often fails while switching branches.

anacierdem (approved):
LGTM

PR #308, opened 2022-09-17
rsp.inc: improve syntax of single lane vector ops

snacchus:
The syntax of the following opcodes was improved:
* `vnop`
* `vmov`
* `vrcp`
* `vrcph`
* `vrcpl`
* `vrsq`
* `vrsqh`
* `vrsql`

Previously, they were treated the same as normal vector ops, when in reality they work a bit differently. The vs register is interpreted as the element of the destination register where the result will be written to. It works the same as the regular element (the "source" element) and thus now has the same syntax. Another difference is that both the source and the destination element are mandatory in the new syntax of these opcodes, and only lane accessors can be used.
`vnop` is an exception, which now takes no arguments.


rasky:
Tested in the OpenGL branch, works correctly.

PR #309, opened 2022-09-19
Add surface_t and use it in graphics.c and rdp.c

rasky:
Another partial merge coming from the rdpq/opengl branch.

This introduces the new `surface_t` type and use it everywhere. Thanks to a trick, also display.c
will start using it in a backward compatible fashion.

See the descriptions of commit by commit for further information.

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -8,6 +8,32 @@
 
 #include <stdint.h>
 
+/**
+ * @defgroup display Display Subsystem
+ * @ingroup libdragon
+ * @brief Video interface system for configuring video output modes and displaying rendered
+ *        graphics.
+ *
+ * The display subsystem handles interfacing with the video interface (VI)
+ * and the hardware rasterizer (RDP) to allow software and hardware graphics
+ * operations.  It consists of the @ref display, the @ref graphics and the
+ * @ref rdp modules.  A separate module, the @ref console, provides a rudimentary
+ * console for developers.  Only the display subsystem or the console can be
+ * used at the same time.  However, commands to draw console text to the display
+ * subsystem are available.
+ *
+ * The display subsystem module is responsible for initializing the proper video
+ * mode for displaying 2D, 3D and software graphics.  To set up video on the N64,
+ * code should call #display_init with the appropriate options.  Once the display
+ * has been set, a surface can be requested from the display subsystem using
+ * #display_lock.  To draw to the acquired surface, code should use functions
+ * present in the @ref graphics and the @ref rdp modules.  Once drawing to a surface
+ * is complete, the rendered graphic can be displayed to the screen using 
+ * #display_show.  Once code has finished rendering all graphics, #display_close can 
+ * be used to shut down the display subsystem.
+ *
+ */
+
 ///@cond
 struct surface_s;
 typedef struct surface_s surface_t;
@@ -81,14 +107,85 @@ typedef surface_t* display_context_t;
 extern "C" {
 #endif
 
+/**
+ * @brief Initialize the display to a particular resolution and bit depth
+ *
+ * Initialize video system.  This sets up a double, triple, or multiple
+ * buffered drawing surface which can be blitted or rendered to using
+ * software or hardware.
+ *
+ * @param[in] res
+ *            The requested resolution
+ * @param[in] bit
+ *            The requested bit depth
+ * @param[in] num_buffers
+ *            Number of buffers, usually 2 or 3, but can be more.
+ * @param[in] gamma
+ *            The requested gamma setting
+ * @param[in] aa
+ *            The requested anti-aliasing setting
+ */
 void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma_t gamma, antialias_t aa );
-surface_t* display_lock(void);
-void display_show(surface_t* surf);
+
+/**
+ * @brief Close the display
+ *
+ * Close a display and free buffer memory associated with it.
+ */
 void display_close();
 
+/**
+ * @brief Lock a display buffer for rendering
+ *
+ * Grab a surface that is safe for drawing.  If none is available
+ * then this will return 0, without blocking. 
+ * 
+ * When you are done drawing on the buffer, use #display_show to unlock
+ * the surface and schedule the buffer to be displayed on the screen during
+ * next vblank.
+ * 
+ * It is possible to lock more than a display buffer at the same time, for
+ * instance to begin working on a new frame while the previous one is still
+ * being rendered in parallel through RDP. It is important to notice that
+ * surfaces will always be shown on the screen in locking order,
+ * irrespective of the order #display_show is called.
+ *
+ * @return A valid surface to render to or NULL if none is available.
+ */
+surface_t* display_lock(void);
+
+/**
+ * @brief Display a previously locked buffer
+ *
+ * Display a previously-locked surface to the screen on the next vblank. The
+ * surface should be locked via #display_lock.
+ * 
+ * This function does not accept any arbitrary surface, but only those returned
+ * by #display_lock.
+ * 
+ * @param[in] surf
+ *            A surface to show (previously retrieved using #display_lock)
+ */
+void display_show(surface_t* surf);
+
+/**
+ * @brief Get the currently configured width of the display in pixels
+ */
 uint32_t display_get_width(void);
+
+/**
+ * @brief Get the currently configured height of the display in pixels
+ */
 uint32_t display_get_height(void);
+
+/**
+ * @brief Get the currently configured bitdepth of the display (in bytes per pixels)
+ */
 uint32_t display_get_bitdepth(void);
+
+/**
+ * @brief Get the currently configured number of buffers
+ */
 uint32_t display_get_num_buffers(void);
 
 #ifdef __cplusplus
--- a/include/display.h
+++ b/include/display.h
@@ -8,6 +8,11 @@
 
 #include <stdint.h>
 
+///@cond
+struct surface_s;
+typedef struct surface_s surface_t;
+///@endcond
+
 /**
  * @addtogroup display
  * @{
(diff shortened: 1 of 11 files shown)


include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+ * @file surface.h
+ * @brief Surface buffers used to draw images
+ * @ingroup graphics
+ * 
+ * This module implements a structure #surface_t which holds the basic
+ * information for a buffer of memory to be used for graphics rendering.
+ * 
+ * A surface is described by the following properties:
+ * 
+ *  * Size (width. height)
```
snacchus: -> "width, height"

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+ *      surface_t buf = surface_alloc(FMT_RGBA16, 64, 64);
+ *  
+ *      // Draw some text on it (with the CPU)
+ *      graphics_draw_text(&buf, 0, 0, "ABC");
+ * @endcode
+ *
+ * Sometimes, you might have an existing raw pointer to a buffer and need to pass it
+ * to an API that accepts a #surface_t. For those cases, you can use
+ * #surface_make to create a #surface_t instance, that you can throw away
+ * after you called the function.
```
snacchus: It's probably worth mentioning that you can safely call `surface_free` on this kind of surface as well (Same with the sub-surface below)

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+// which are easy to convert back into RDP native fields.
+#define _RDP_FORMAT_CODE(rdp_fmt, rdp_size)        (((rdp_fmt)<<2)|(rdp_size))
+/// @endcond
+
+/** @brief Extract the depth (number of bits per pixel) from a #tex_format_t. (eg: `FMT_RGBA16` => 16) 
+ * 
+ * Note that there are texture format that are 4bpp, so don't divide this by 8 to get the number of bytes
+ * per pixels, but rather use #TEX_FORMAT_BYTES2PIX and #TEX_FORMAT_PIX2BYTES. */
+#define TEX_FORMAT_BITDEPTH(fmt)             (4 << ((fmt) & 0x3))
+/** @brief Convert the specified number of pixels in bytes. */
```
snacchus: -> "to bytes"

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+#define TEX_FORMAT_PIX2BYTES(fmt, pixels)    ((pixels) << (((fmt) & 3) + 2) >> 3)
+/** @brief Convert the specified number of bytes in pixels. */
```
snacchus: -> "to pixels"

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+#define TEX_FORMAT_BYTES2PIX(fmt, bytes)     (((bytes) << 1) >> ((fmt) & 3))
```
snacchus: We should probably mention that both of these macros always round down. Also I think that a version of `TEX_FORMAT_PIX2BYTES` that rounds up is needed (or maybe it should do that by default). For example one might want to allocate a 4 bit per pixel surface with an odd width, in which case it should round up to the next byte. In fact, that should probably be done in `surface_alloc`.

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+ * @param[in] buffer    Pointer to the memory buffer
+ * @param[in] format    Pixel format
+ * @param[in] width     Width in pixels
+ * @param[in] height    Height in pixels
+ * @return              The initialized surface
+ * 
+ * @see #surface_make
+ */
+inline surface_t surface_make_linear(void *buffer, tex_format_t format, uint32_t width, uint32_t height) {
+    return surface_make(buffer, format, width, height, TEX_FORMAT_PIX2BYTES(format, width));
```
snacchus, line 192: What to do here if the format is 4 bits per pixel and the width is odd?

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * @brief Allocate a new surface in memory
+ * 
+ * This function allocates a new surface with the specified pixel format,
+ * width and height. The surface must be freed via #surface_free when it is
+ * not needed anymore.
+ * 
+ * A surface allocated via #surface_alloc can be used as a RDP frame buffer
+ * (passed to #rdp_attach) because it is guarateed to have the required
+ * alignment of 64 bytes.
```
snacchus: Provided it uses one of the supported formats.

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+ * @return              The initialized surface
+ */
+surface_t surface_alloc(tex_format_t format, uint32_t width, uint32_t height);
+
+/**
+ * @brief Initialize a surface_t structure, pointing to a rectangular portion of another
+ *        surface.
+ * 
+ * The surface returned by this function will point to a portion of the buffer of
+ * the parent surface, and will have of course the smae pixel format.
```
snacchus: -> "same"

include/surface.h
```diff
@@ -0,0 +1,245 @@
...
+ * 
+ * @param[in]  parent   Parent surface that will be pointed to
+ * @param[in]  x0       X coordinate of the top-left corner of the parent surface
+ * @param[in]  y0       Y coordinate of the top-left corner of the parent surface
```
snacchus: Technically, they are the coordinates of the new sub-surface within the parent surface. The wording might need to be clarified.

src/display.c
```diff
@@ -162,27 +166,32 @@ static const uint32_t * const reg_values[] = {
...
 /** @brief Pointer to uncached 16-bit aligned version of buffers */
-void *__safe_buffer[NUM_BUFFERS];
+static void *__safe_buffer[NUM_BUFFERS];
 
 /** @brief Currently displayed buffer */
 static int now_showing = -1;
 
-/** @brief Complete drawn buffer to display next */
-static int show_next = -1;
+/** @brief True if the buffer indexed by now_drawing is currently locked */
```
snacchus: Is this doc still correct?

src/display.c
```diff
@@ -479,42 +486,53 @@ void display_close()
     for( int i = 0; i < __buffers; i++ )
     {
         /* Free framebuffer memory */
-        if( buffer[i] )
+        if( __safe_buffer[i] )
         {
-            free( buffer[i]);
+            free_uncached( __safe_buffer[i]);
```
snacchus: Why not use `surface_free` instead?

src/display.c
```diff
@@ -479,42 +486,53 @@ void display_close()
...
+ * display contexts will always be shown on the screen in locking order,
+ * irrespective of the order #display_show is called.
  *
  * @return A valid display context to render to or 0 if none is available.
  */
-display_context_t display_lock()
+display_context_t display_lock(void)
 {
-    display_context_t retval = 0;
+    display_context_t retval = NULL;
```
snacchus: Update to use `surface_t *` as well

include/display.h
```diff
@@ -65,18 +70,27 @@ typedef enum
...
-display_context_t display_lock();
-void display_show(display_context_t disp);
+surface_t* display_lock(void);
+void display_show(surface_t* disp);
 void display_close();
 
+uint32_t display_get_width(void);
+uint32_t display_get_height(void);
+uint32_t display_get_bitdepth(void);
+uint32_t display_get_num_buffers(void);
```
snacchus: This is probably a good opportunity to move the docs to the header file.

src/surface.c
```diff
@@ -0,0 +1,61 @@
...
+
+surface_t surface_make_sub(surface_t *parent, uint32_t x0, uint32_t y0, uint32_t width, uint32_t height)
+{
+    assert(x0 + width <= parent->width);
+    assert(y0 + height <= parent->height);
+
+    tex_format_t fmt = surface_get_format(parent);
+
+    surface_t sub;
+    sub.buffer = parent->buffer + y0 * parent->stride + TEX_FORMAT_PIX2BYTES(fmt, x0);
```
snacchus, line 60: Again, what to do if width is odd and the surface has a 4 bits per pixel format?

src/surface.c
```diff
@@ -0,0 +1,61 @@
...
+}
+
+surface_t surface_alloc(tex_format_t format, uint32_t width, uint32_t height)
+{
+    return (surface_t){ 
+        .flags = format | SURFACE_FLAGS_OWNEDBUFFER,
+        .width = width,
+        .height = height,
+        .stride = TEX_FORMAT_PIX2BYTES(format, width),
+        .buffer = malloc_uncached_aligned(64, height * TEX_FORMAT_PIX2BYTES(format, width)),
```
snacchus, line 37: See my comment in surface.h

src/display.c
```diff
@@ -479,42 +486,53 @@ void display_close()
...
+    free(surfaces);
+    surfaces = NULL;
+
     enable_interrupts();
 }
 
 /**
  * @brief Lock a display buffer for rendering
  *
  * Grab a display context that is safe for drawing.  If none is available
```
snacchus: Should we still use the term "display context" in documentation or just switch to "surface"?

snacchus (commented):
Implementation looks good to me, my only concern is with `TEX_FORMAT_PIX2BYTES` in combination with 4 bit per pixel formats (see comments). Otherwise made some comments on documentation.

include/display.h
rasky: done

include/surface.h
rasky: done

include/surface.h
rasky: done

include/surface.h
rasky: I think it makes sense to always roundup `TEX_FORMAT_PIX2BYTES`. For `TEX_FORMAT_BYTES2PIX`, the current rounding-down behavior should symmetrically be the correct one.

I'ved added the roundup and updated the documentation.

include/surface.h
rasky: In the latest version, the stride is rounded up, which I think makes sense in the case as well.

src/surface.c
rasky: This should be fixed as well with the roundup.

src/surface.c
rasky: It's not possible to make a sub surface starting from an odd offset in a 4bpp format, for obvious reasons. For even offsets, this should work now.

rasky:
@snacchus thanks for the review, please double check the odd/even stuff with 4bpp

src/surface.c
snacchus: So if `x0` is odd, then this will silently set the beginning of the buffer to the next byte. We should probably assert in that case, because this could have unforeseen consequences (like overflowing at the end of the line). And obviously the first pixel of each line is not what the user might expect.

include/display.h
snacchus: Have you committed this yet? It's still in the .c file for me.

rasky: I reverted it and forgot to mention it. I will do this in a final commit before merging otherwise there are too many conflicts in case other changes are needed

src/display.c
rasky: good catch, fixed

src/surface.c
rasky: Done, see if it seems OK to you.

snacchus: Looks good.

include/display.h
rasky: Added now

PR #310, opened 2022-09-24
debug: SC64 version 2 - support for SD card and USB functionality

Polprzewodnikowy:
This change brings SD card functionality for new SC64 version and replaces old unsupported USB code with new implementation. All references to SummerCart64 were replaced with new name - SC64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ via `fprintf(stderr)`, see the debug library for more information.
 
 All cartridges that are able to load custom ROMs should be able to successfully
 load libdragon ROMs via either USB/serial, or from a MMC/SD card. For instance,
-the followig are known to work: 64drive, EverDrive64 (all models), SummerCart 64.
+the followig are known to work: 64drive, EverDrive64 (all models), SC64.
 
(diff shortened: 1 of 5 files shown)


rasky:
Thanks! Are the changes in usb.c exactly those that were submitted upstream to UNFloader?

Polprzewodnikowy:
I've updated `usb.c` with changes from N64-UNFLoader repo. Only differences are lines
```c
/* 591 */ s8 usb_64drive_wait();
/* 627 */ void usb_64drive_setwritable(u8 enable);
/* 1331 */ u32 usb_sc64_execute_cmd(u8 cmd, u32 *args, u32 *result);
```
where `static` modifier has been removed to expose these functions outside for use with the sdfs subsystem.

anacierdem:
Upstream usb.c reverted the timeout fixes we did back in the day so I expect a regression with my ED setup. I want to check if it is still working fine with the latest changes to see how it behaves.

rasky:
The last push-force includes a version of usb.c that has restored the timeout patch. We keep it as a fork from upstream for now.

PR #311, opened 2022-10-15
display: Support arbitrary resolutions

sp1187:
This PR adds support for custom resolutions by changing the replacing the existing `resolution_t` enum type with a struct type of the same name that includes width, height and interlacing as its values, which is then sent as an argument to `display_init` like before. The new implementation includes some asserts to protect against invalid settings that do not work on real hardware.

There is still one bug left to resolve. When changing resolutions frequently, it seems that some resolutions will break the VI output in real hardware and leave it in a corrupt state that remains even after soft reset. This can be reproduced that applying this patch to the `test` example that I made for testing the feature and then press D-pad right, then D-pad down (which should correspond to 322x241). 

```diff
diff --git a/examples/test/test.c b/examples/test/test.c
index b211a01e..b25fd846 100644
--- a/examples/test/test.c
+++ b/examples/test/test.c
@@ -70,10 +70,8 @@ int main(void)
         /*Fill the screen */
         graphics_fill_screen( disp, 0 );
 
-        sprintf(tStr, "Video mode: %lu\n", *((uint32_t *)0x80000300));
+        sprintf(tStr, "Size: %lux%lu\n", res.width, res.height);
         graphics_draw_text( disp, 20, 20, tStr );
-        sprintf(tStr, "Status: %08lX\n", *((uint32_t *)0xa4400000));
-        graphics_draw_text( disp, 20, 30, tStr );
 
         /* Full bright color */
         graphics_draw_box(disp, 20, 40, 20, 20, graphics_make_color(255, 0, 0, 255));
@@ -125,32 +123,28 @@ int main(void)
         if( keys.c[0].up )
         {
             display_close();
-
-            res = RESOLUTION_640x480;
+            res.height--;
             display_init( res, bit, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE );
         }
 
         if( keys.c[0].down )
         {
             display_close();
-
-            res = RESOLUTION_320x240;
+            res.height++;
             display_init( res, bit, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE );
         }
 
         if( keys.c[0].left )
         {
             display_close();
-
-            bit = DEPTH_16_BPP;
+            res.width -= 2;
             display_init( res, bit, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE );
         }
 
         if( keys.c[0].right )
         {
             display_close();
-
-            bit = DEPTH_32_BPP;
+            res.width += 2;
             display_init( res, bit, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE );
         }
     }
```

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -6,6 +6,7 @@
 #ifndef __LIBDRAGON_DISPLAY_H
 #define __LIBDRAGON_DISPLAY_H
 
+#include <stdbool.h>
 #include <stdint.h>
 
 /**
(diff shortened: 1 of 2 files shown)


include/display.h
```diff
@@ -14,24 +15,30 @@
...
+static const resolution_t RESOLUTION_320x240 = {320, 240, false};
+/** @brief 512x240 mode, high-res progressive */
+static const resolution_t RESOLUTION_512x240 = {512, 240, false};
+/** @brief 640x240 mode, high-res progressive */
+static const resolution_t RESOLUTION_640x240 = {640, 240, false};
+/** @brief 512x480 mode, interlaced */
+static const resolution_t RESOLUTION_512x480 = {512, 480, true};
+/** @brief 640x480 mode, interlaced */
+static const resolution_t RESOLUTION_640x480 = {640, 480, true};
+
```
rasky, line 75: `static` members are not exposed in documentation, which would make the API a bit harder to understand. One alternative is to declare them here, and define them in display.c.

include/display.h
```diff
@@ -14,24 +15,30 @@
  */
 
 /**
- * @brief Valid video resolutions
+ * @brief Video resolution structure
  */
```
rasky, line 53: Can you maybe document here that the user can either fill the structure or use one of the "standard" / "common" resolution provided, with a list of them?

include/display.h
```diff
@@ -14,24 +15,30 @@
...
-    RESOLUTION_512x480,
-    /** @brief 512x240 mode, high-res progressive */
-    RESOLUTION_512x240,
-    /** @brief 640x240 mode, high-res progressive */
-    RESOLUTION_640x240,
+typedef struct {
+    /** @brief Screen width */
+    int32_t width;
+    /** @brief Screen height */
+    int32_t height;
```
rasky, line 58: Maybe I'd also mention min/max values for width and height. Later we can improve this if we find more restrictions on allowed values.

rasky (approved):
Thanks @sp1187, I'll play with the sample and see if I can pinpoint the crashing configurations

sp1187:
Seems that waiting for vblank before updating the VI settings was enough to avoid the VI crash, I updated the PR to have waits in `display_init` and `display_close`.

rasky:
Merged manually, thanks

PR #316, opened 2022-11-02
mixer: fix audio statics at the end of non-looping waveforms.

### LINKED ISSUE #302: SFX glitching
Issue being posted after discussion with Rasky:

Playback of sound effects that don't loop causes an audible glitch in some cases at the end of the sound.

If you replace the cannon sound affect in mixertest with this one you can hear the issue.
[sample_sound.zip](https://github.com/DragonMinded/libdragon/files/9490592/sample_sound.zip)

rasky:
rsp_mixer ucode has an "optimization" to avoid an expensive per-sample check: it doesn't check if it's reading out of bounds (within the samples that were DMA'd into DMEM). This means that in general it can read up to 64 bytes after the end of a waveform. In the mixer code, this is referred as "waveform over-read".

To workaround for this, audioconv64 pads non-looping waveforms with 64 bytes of silence, and pads looping waveforms by repeating the first 64 bytes of the loop beginning. This in general works very well and the overhead is basically none.

The only problem is that, at some point during mixer/xm64 development, I got confused and thought that overread was only needed for looping waveforms, so I axed a few places were the code was accounting for over-reads in non-looping waveforms, and even added FIXME to other places to remember myself to remove them in the future.

This is so wrong. The overread is badly needed for non looping waveforms as well. So this commit reinstates the missing calculations, and removes the wrong FIXMEs.

An unfortunate follow-up to this is that the optimal buffer calculation for XM64 was one of the rouine were overread computation was removed, so RAM buffers allocated for XM64 files created until today are not sufficient anymore after this commit. This requires an internal version bump of the XM64 format to avoid random asserts to users. I have also added the code to load the current version and simply increase the buffers manually (it is a conservative increase, but better than just crashing). Reconverting the files via audioconv will of course calculate the new optimal values.

Fixes #302

### DIFF
--- a/src/audio/mixer.c
+++ b/src/audio/mixer.c
@@ -505,6 +505,9 @@ static void mixer_exec(int32_t *out, int num_samples) {
 				// actually present in the waveform.
 				if (wpos+wlen > len)
 					wlen = len-wpos;
+				// FIXME: due to a limit in the RSP ucode, we need to overread
+				// more data, possibly even past the end of the sample
+				wlen += MIXER_LOOP_OVERREAD >> bps;
 				assert(wlen >= 0);
 			} else if (loop_len < sbuf->size) {
 				// If the whole loop fits the sample buffer, we just need to
--- a/src/audio/libxm/context.c
+++ b/src/audio/libxm/context.c
@@ -127,7 +127,7 @@ void xm_context_save(xm_context_t* ctx, FILE* out) {
 	#define WALIGN()  ({ while (ftell(out) % 8) _W8(0); })
 
 
-	const uint8_t version = 5;
+	const uint8_t version = 6;
 	WA("XM64", 4);
 	W8(version);
 	W32(ctx->ctx_size);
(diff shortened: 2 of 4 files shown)


src/audio/mixer.c
```diff
@@ -505,6 +505,9 @@ static void mixer_exec(int32_t *out, int num_samples) {
 				// actually present in the waveform.
 				if (wpos+wlen > len)
 					wlen = len-wpos;
+				// FIXME: due to a limit in the RSP ucode, we need to overread
```
anacierdem, line 508: What needs fixing here? I presumed all converters would zero-pad the samples.

anacierdem (commented):
Looks good to me, I will test this when I have the time as well. I am not sure if I can provide a definitive answer on whether it is fixed as it was happening at random (as it would be expected from an incorrect bound). I only have a question in the comments.

src/audio/mixer.c
rasky: I think the idea is: fixing the ucode so that the RSP doesn't have that limit anymore, assuming it can be done without waiving too much performance.

Issue #110, opened 2020-10-06
Remove "copy code and data" in entrypoint


jago85:
Hi,
I want to discuss the need for the following lines:
https://github.com/DragonMinded/libdragon/blob/d310a6b3c65aac02fdd53ab2e621a54dbd791dfe/src/entrypoint.S#L22-L40

I came across this because, in my project, I had to remove them, as I load the ROM from the PIF instead of the cartridge. The funny thing is that, after compiling Libdragon without this code, the ROMs were still working from the cart. This is because the bootcode (at least 6102) copies 1 MiB of data using DMA from cart to RAM. 

So I see the following problems:
- the copy code is pointless when the ROM is smaller than 1 MiB
    - the code copies the same data that was DMA'ed to RAM recently
    - the code is slow using IO reads
- the code doesn't check memory size
   - the code may be handy as soon as the ROM is bigger than 1 MiB
   - but it should at least check if the ROM fits into the available memory

Most small projects should be in memory after the DMA transfer of the bootcode. If someone needs more code/data then loading should be implemented separately. 

Things I am not aware of:
- Don't know if such a change would break projects using bigger ROMs. 
   - This would not affect the DFS as it is already treated separately and is read directly from the cart.
- Does someone use the 6105 bootcode? I don't know how this works. It will probably do the DMA transfer too.
- Maybe someone can point out other facts why we need this.

DragonMinded:
I actually don't remember if 6105 would break with this. If it works on real HW after removing this then it does indeed seem redundant. I would be okay with a documented caveat that libdragon works with the N64 boot code and thus expects you to remain within 1 MiB and if you want more, look into gcc overlays or dynamic loading of code using DFS. That matches how lots of systems work in practice.

jago85:
Yes, it does work on real HW.
Do you have an idea where to document it? Then I could prepare this.

clbr:
I was wondering why libdragon's pif boot code does not touch the RSP when several commercial roms do (conker, majora). Where does it come from? Git history just says "merge alt_libn64", and I can't find much on that.

clbr:
Its md5sum doesn't seem to match any of 6101, 2, 3, 5, 6 available in mikeryan's repo (last 4032 bytes).

jago85:
It uses the 6102 bootcode. (The header file in the repo's root.)

But this has nothing to do with the entry code that my question was related to.

[...]

PR #318, opened 2022-11-13
entrypoint: use DMA to load text/data section instead of CPU writes

rasky:
This speeds up a bit the boot of libdragon applications. It uses PI DMA to fetch text/data from ROM, and clears BSS using uncached memory (which is faster for this kind of write-once tasks) while the DMA is in progress.

Updates #110


anacierdem (approved):
LGTM, thanks for looking into this!

PR #319, opened 2022-11-19
Fix C++ Exceptions

gamemasterplc:
Breaks linking dfsdemo, mptest, mputest, vrutest, and vtest. All of these demos that fail to link do not use n64.mk

### DIFF
--- a/n64.ld
+++ b/n64.ld
@@ -53,7 +53,10 @@ SECTIONS {
     } > mem
 
    .eh_frame_hdr : { *(.eh_frame_hdr) } > mem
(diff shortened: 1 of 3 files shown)


gamemasterplc:
Added a change to fix demos that don't link in libgcc. This adds a weak definition of __register_frame_info to entrypoint.S.

anacierdem (commented):
This looks good to me but I am not confident enough approving it b/c I don’t know GCC specifics very well. I think it is good for @rasky to have a final look :)

rasky (approved):
Thanks

PR #321, opened 2022-12-09
Small changes from the OpenGL branch

rasky:
This is another set of unrelated changes coming from the OpenGL. These can be merged safely even right now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,7 @@ install-mk: n64.mk
 install: install-mk libdragon
 	install -Cv -m 0644 libdragon.a $(INSTALLDIR)/mips64-elf/lib/libdragon.a
 	install -Cv -m 0644 n64.ld $(INSTALLDIR)/mips64-elf/lib/n64.ld
+	install -Cv -m 0644 rsp.ld $(INSTALLDIR)/mips64-elf/lib/rsp.ld
 	install -Cv -m 0644 header $(INSTALLDIR)/mips64-elf/lib/header
 	install -Cv -m 0644 libdragonsys.a $(INSTALLDIR)/mips64-elf/lib/libdragonsys.a
 	install -Cv -m 0644 include/pputils.h $(INSTALLDIR)/mips64-elf/include/pputils.h
--- a/include/pputils.h
+++ b/include/pputils.h
@@ -63,6 +63,42 @@
 #define __PPCAT2(n,x) n ## x
 #define __PPCAT(n,x) __PPCAT2(n,x)
 
+// __CALL_FOREACH_BIS. Like __CALL_FOREACH, but it allows to be called without the expansion
+// of a __CALL_FOREACH.
+#define __FEB_0(_call, ...)
+#define __FEB_1(_call, x)       _call(x)
+#define __FEB_2(_call, x, ...)  _call(x) __FEB_1(_call, __VA_ARGS__)
+#define __FEB_3(_call, x, ...)  _call(x) __FEB_2(_call, __VA_ARGS__)
+#define __FEB_4(_call, x, ...)  _call(x) __FEB_3(_call, __VA_ARGS__)
+#define __FEB_5(_call, x, ...)  _call(x) __FEB_4(_call, __VA_ARGS__)
+#define __FEB_6(_call, x, ...)  _call(x) __FEB_5(_call, __VA_ARGS__)
+#define __FEB_7(_call, x, ...)  _call(x) __FEB_6(_call, __VA_ARGS__)
+#define __FEB_8(_call, x, ...)  _call(x) __FEB_7(_call, __VA_ARGS__)
+#define __FEB_9(_call, x, ...)  _call(x) __FEB_8(_call, __VA_ARGS__)
+#define __FEB_10(_call, x, ...) _call(x) __FEB_9(_call, __VA_ARGS__)
+#define __FEB_11(_call, x, ...) _call(x) __FEB_10(_call, __VA_ARGS__)
+#define __FEB_12(_call, x, ...) _call(x) __FEB_11(_call, __VA_ARGS__)
+#define __FEB_13(_call, x, ...) _call(x) __FEB_12(_call, __VA_ARGS__)
+#define __FEB_14(_call, x, ...) _call(x) __FEB_13(_call, __VA_ARGS__)
+#define __FEB_15(_call, x, ...) _call(x) __FEB_14(_call, __VA_ARGS__)
+#define __FEB_16(_call, x, ...) _call(x) __FEB_15(_call, __VA_ARGS__)
+#define __FEB_17(_call, x, ...) _call(x) __FEB_16(_call, __VA_ARGS__)
+#define __FEB_18(_call, x, ...) _call(x) __FEB_17(_call, __VA_ARGS__)
+#define __FEB_19(_call, x, ...) _call(x) __FEB_18(_call, __VA_ARGS__)
+#define __FEB_20(_call, x, ...) _call(x) __FEB_19(_call, __VA_ARGS__)
+#define __FEB_21(_call, x, ...) _call(x) __FEB_20(_call, __VA_ARGS__)
+#define __FEB_22(_call, x, ...) _call(x) __FEB_21(_call, __VA_ARGS__)
+#define __FEB_23(_call, x, ...) _call(x) __FEB_22(_call, __VA_ARGS__)
+#define __FEB_24(_call, x, ...) _call(x) __FEB_23(_call, __VA_ARGS__)
+#define __FEB_25(_call, x, ...) _call(x) __FEB_24(_call, __VA_ARGS__)
+#define __FEB_26(_call, x, ...) _call(x) __FEB_25(_call, __VA_ARGS__)
+#define __FEB_27(_call, x, ...) _call(x) __FEB_26(_call, __VA_ARGS__)
+#define __FEB_28(_call, x, ...) _call(x) __FEB_27(_call, __VA_ARGS__)
+#define __FEB_29(_call, x, ...) _call(x) __FEB_28(_call, __VA_ARGS__)
+#define __FEB_30(_call, x, ...) _call(x) __FEB_29(_call, __VA_ARGS__)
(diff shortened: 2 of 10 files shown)


src/utils.h
```diff
@@ -5,8 +5,11 @@
  * Misc utilities functions and macros. Internal header.
  */
 
+#define SWAP(a, b) ({ typeof(a) t = a; a = b; b = t; })
```
TheFakeMontyOnTheRun, line 8: Perhaps wouldn't it be safer to wrap `a` and `b` with parenthesis?

rasky: In this specific case it shouldn’t matter as both variables are also used as LHS of an assignment. I don’t think there’s a case where the parenthesis would matter *and* it would compile

include/pputils.h
```diff
@@ -63,6 +63,42 @@
...
+#define __FEB_23(_call, x, ...) _call(x) __FEB_22(_call, __VA_ARGS__)
+#define __FEB_24(_call, x, ...) _call(x) __FEB_23(_call, __VA_ARGS__)
+#define __FEB_25(_call, x, ...) _call(x) __FEB_24(_call, __VA_ARGS__)
+#define __FEB_26(_call, x, ...) _call(x) __FEB_25(_call, __VA_ARGS__)
+#define __FEB_27(_call, x, ...) _call(x) __FEB_26(_call, __VA_ARGS__)
+#define __FEB_28(_call, x, ...) _call(x) __FEB_27(_call, __VA_ARGS__)
+#define __FEB_29(_call, x, ...) _call(x) __FEB_28(_call, __VA_ARGS__)
+#define __FEB_30(_call, x, ...) _call(x) __FEB_29(_call, __VA_ARGS__)
+#define __FEB_31(_call, x, ...) _call(x) __FEB_30(_call, __VA_ARGS__)
+#define __CALL_FOREACH_BIS(fn, ...)  __GET_33RD_ARG("ignored", ##__VA_ARGS__, __FEB_31, __FEB_30, __FEB_29, __FEB_28, __FEB_27, __FEB_26, __FEB_25, __FEB_24, __FEB_23, __FEB_22, __FEB_21, __FEB_20, __FEB_19, __FEB_18, __FEB_17, __FEB_16, __FEB_15, __FEB_14, __FEB_13, __FEB_12, __FEB_11, __FEB_10, __FEB_9, __FEB_8, __FEB_7, __FEB_6, __FEB_5, __FEB_4, __FEB_3, __FEB_2, __FEB_1, __FEB_0)(fn, ##__VA_ARGS__)
```
anacierdem, line 100: I don't know the meaning of __CALL_FOREACH_BUS but there is either a type here or in the comment above

include/pputils.h
```diff
@@ -63,6 +63,42 @@
 #define __PPCAT2(n,x) n ## x
 #define __PPCAT(n,x) __PPCAT2(n,x)
 
+// __CALL_FOREACH_BUS. Like __CALL_FOREACH, but it allows to be called wihtout the expansion
```
anacierdem: typo: without

include/rsp_queue.inc
```diff
@@ -167,6 +167,26 @@ _RSPQ_SAVED_STATE_END:
...
+# This register is initialized to zero any time a command is called
+#define vzero    $v00
+
+# These registers are initialized with the constant data required to make
+# vector shift macros code works (power of twos).
+#define vshift   $v30
+#define vshift8  $v31
+
+# We also define direct access to small constants as they can be useful in some
+# calculatios.
```
anacierdem: typo: calculations

n64.mk
```diff
@@ -107,7 +107,8 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.S
 		DATASECTION="$(basename $@).data"; \
 		BINARY="$(basename $@).elf"; \
 		echo "    [RSP] $<"; \
-		$(N64_CC) $(RSPASFLAGS) -nostartfiles -Wl,-Ttext=0x1000 -Wl,-Tdata=0x0 -Wl,-e0x1000 -o $$BINARY $<; \
+		$(N64_CC) $(RSPASFLAGS) -nostartfiles -Wl,-Trsp.ld -Wl,--gc-sections -o $@ $<; \
+		mv "$@" $$BINARY; \
```
anacierdem, line 111: Why move later rather than direct output?

rsp.ld
```diff
@@ -0,0 +1,50 @@
...
+    /* This is a workaround to make ld place text symbols at the correct addresses (0x0 - 0x1000).
+       The RSP technically uses a harvard-architecture (https://en.wikipedia.org/wiki/Harvard_architecture)
+       which means that it uses different address spaces for instructions and data accesses.
+       Because ld is not designed for such architectures, we need to place the data section somewhere different,
+       since it would otherwise overlap the text section. As a workaround, we place it at 0x04000000 (which is also
+       the location of DMEM from the VR4300's point of view). Because the RSP only uses the lower 12 bits
+       of any address, this works out fine (as long as we always wrap data addresses in "%lo()").
+       
+       Note that this is not actually required to run the ucode correctly (instruction addresses above 0x1000 are truncated anyway),
+       but it makes debugging with gdb a lot easier (e.g. using this fork of cen64 https://github.com/lambertjamesd/cen64).
```
anacierdem, line 24: hmm, cen64 reports the dmem/imem contents at these VMA locations then?

Makefile
```diff
@@ -70,6 +70,7 @@ install-mk: n64.mk
 install: install-mk libdragon
 	install -Cv -m 0644 libdragon.a $(INSTALLDIR)/mips64-elf/lib/libdragon.a
 	install -Cv -m 0644 n64.ld $(INSTALLDIR)/mips64-elf/lib/n64.ld
+	install -Cv -m 0644 rsp.ld $(INSTALLDIR)/mips64-elf/lib/rsp.ld
```
anacierdem, line 73: We should also add this as an elf dependency to make sure it is rebuilt when working on the linker script in the future:
```%.elf: $(N64_LIBDIR)/libdragon.a $(N64_LIBDIR)/libdragonsys.a $(N64_LIBDIR)/n64.ld $(N64_LIBDIR)/rsp.ld```

File: tests/testrom.c [hunk 8]
```diff
@@ -299,9 +309,12 @@ int main() {
...
 			failures++;
 			printf("FAIL\n\n");
-
 			if (ctx.log != logbuf) {
-				printf("%s\n\n", logbuf);
+				debugf("%s\n", logbuf);
+			}
+			if (ctx.err != errbuf) {
+				printf("%s\n", errbuf);
+				debugf("%s\n", errbuf);
```
anacierdem, line 317: Was there a specific reason why we do `console_set_debug(false);` at start of tests?

anacierdem (changes requested):
Thanks! There is only a single typo on a macro name that I'm not sure if its correct and we can add the linker script as an elf dependency. o/w nothing to worry about.

include/pputils.h
rasky: Right, the correct name is `__CALL_FOREACH_BIS` which means a duplicate of `__CALL_FOREACH`. The macros Is not reentrant which means that if you need to use it twice recursively, you need two of them. rdpq will need that.

include/pputils.h
rasky: Fixed

include/rsp_queue.inc
rasky: Fixed

n64.mk
rasky: It happened often enough to me to have a truncated / zero byte .o file when something went wrong. I'm not sure if it's an artifact of how GCC handles assembly files (maybe when doing .S -> .o, it opens the output file immediately), because it never happens with .o files coming from .c files.

Makefile
rasky: Probably this would have to be done for the RSP compilation rule:

```
$(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.S $(N64_LIBDIR)/rsp.ld
```

but the problem is that `$(N64_LIBDIR)/rsp.ld` only exists after libdragon is installed, and that rule needs to work also for building libdragon itself (since libdragon needs ucode). This is actually similar to n64.mk issues where we ended up installing it before anything else. So if we wanted to go this route, I think we should also install `n64.ld` in the `install-mk` rule (maybe renaming it `install-pre` or something), which also requires sudo etc. So I guess it would require some testing so I'd rather postpone it.

tests/testrom.c
rasky: I think my general idea was that the test output on the screen wouldn't always match the console. On the screen we do something graphical formatting / alignment that is not required in the logs. Also maybe in the future we should probably switch to a slightly more advanced UI (eg: having a fixed header and a scrollable area, etc). So I started from the day one with this idea and kept the two separated. If you look at the code, there are already some differences, albeit not fundamental for now, I agree with that.

rsp.ld
rasky: Yes I think so, but we would have to ask Snacchus about this actually.

snacchus: Yes this is actually somewhat specific to cen64. It will report DMEM/IMEM at `0x04000000`, but only to the debugger. This address is derived from the uncached virtual address `0xA4000000` where DMEM/IMEM is mirrored on hardware. But `0x04000000` doesn't actually make sense in any other context. We could change it to the aforementioned virtual uncached address, but then RSP-debugging with cen64 as-is will break slightly. This might not be a bad thing, because to my knowledge, I'm basically the only person who uses that specific feature, and I could just modify cen64.

anacierdem: If this maps well to the actual memory map and not cen64 specific, let's keep it

Makefile
anacierdem: You are right, just opened an issue in order to not forget about it: #323

rsp.ld
anacierdem: Ah sorry I haven't read it properly. Looks like the correct approach here is using the unchaced address. OTOH I am ok to have this as is right now and open respective issues here and cen64 and resolve later.

rasky: Done

PR #327, opened 2023-01-10
build: export FatFs header files / sdfs: added RTC support

Polprzewodnikowy:
This change will redefine all publicly exported function symbols from FatFs library. It allows for linking apps with user provided FatFs code without namespace clash. This is beneficial for flashcart menu loaders as low level interaction with filesystem is required.

Mechanism can also be used for other external libraries integrated into libdragon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,6 +114,8 @@ install: install-mk libdragon
 	install -Cv -m 0644 include/rspq.h $(INSTALLDIR)/mips64-elf/include/rspq.h
 	install -Cv -m 0644 include/rspq_constants.h $(INSTALLDIR)/mips64-elf/include/rspq_constants.h
 	install -Cv -m 0644 include/rsp_queue.inc $(INSTALLDIR)/mips64-elf/include/rsp_queue.inc
+	install -CDv -m 0644 src/fatfs/diskio.h $(INSTALLDIR)/mips64-elf/include/fatfs/diskio.h
+	install -CDv -m 0644 src/fatfs/ff.h $(INSTALLDIR)/mips64-elf/include/fatfs/ff.h
 
 
 clean:
--- a/src/debug.c
+++ b/src/debug.c
@@ -511,6 +511,7 @@ bool debug_init_sdfs(const char *prefix, int npart)
 	strlcpy(sdfs_prefix, prefix, sizeof(sdfs_prefix));
 	attach_filesystem(sdfs_prefix, &fat_fs);
 	enabled_features |= DEBUG_FEATURE_FILE_SD;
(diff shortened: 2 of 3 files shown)


rasky:
Can we do the opposite? Let’s export the symbols aka install the headers. I don’t think it’s good to have two copies of the same library.

rasky (changes requested):
As indicated by the previous comment, I think we'd prefer a different approach: exporting the current copy of fatfs to users, rather than hiding it.

Polprzewodnikowy:
That approach could work indeed. However, there's still problem with baked in settings in `ffconf.h` file that make sense for game but in menu loader case they are sub optimal. For example RTC support is turned off and `FF_FS_TINY` is enabled. That last setting doesn't even make sense for N64 because we aren't memory limited. Can those settings be changed for every application?

rasky:
I think one can always patch the local copy of ffconf.h and recompile libdragon; in general, I advise to always keep a vendored libdragon copy in each application to simplify builds even in the future, and allow for simple tweaks like this.

If you prefer to have a way to change parameters without touching the file, I guess one could change the defines in ffconf.h to use the syntax:

#ifndef FF_FS_TINY
#define FF_FS_TINY 1
#endif

so that they could be overridden in the Makefile. 

Do any of the above work for you?

Since I'm not a fatfs specialist, can you please describe the negative effects of FF_FS_TINY? Maybe we can reevaluate that default.

Makefile
```diff
@@ -114,6 +114,8 @@ install: install-mk libdragon
 	install -Cv -m 0644 include/rspq.h $(INSTALLDIR)/mips64-elf/include/rspq.h
 	install -Cv -m 0644 include/rspq_constants.h $(INSTALLDIR)/mips64-elf/include/rspq_constants.h
 	install -Cv -m 0644 include/rsp_queue.inc $(INSTALLDIR)/mips64-elf/include/rsp_queue.inc
+	install -Cv -m 0644 src/fatfs/diskio.h $(INSTALLDIR)/mips64-elf/include/diskio.h
+	install -Cv -m 0644 src/fatfs/ff.h $(INSTALLDIR)/mips64-elf/include/ff.h
```
rasky: Maybe we could install those into a fatfs subdir to separate them. Not a big deal but I think it would be a tad neater

rasky (approved):

networkfusion:
Please correct me if I am wrong, but the original solution in this PR allowed release builds, but the new one does not... If so, the compiler directives for fatfs needs changing from `DEBUG` to its own (say `INCLUDE_FATFS`)?!

PR #333, opened 2023-01-23
Create a TOC of files contained in a ROM, and use it for DragonFS

rasky:
These commits do the following:

 * Add to n64tool a feature to create a TOC of the files appended in the ROM
 * Add to libdragon an internal API (rompak_internal.h) to search this TOC. Users should stick to DragonFS of course, but libdragon might find useful to store its own data files in the ROM, without recurring to weird hacks, nor polluting the user-level DragonFS (which might not even exist).
 * Change DragonFS to use the TOC to open the image. This also allows to part with the hardcoded offset, which means that it now works by default even for ROMs bigger than 1Mb. Also the padding to 1Mb is now all at the end of the ROM, ready to be removed as soon as we have an open-source IPL3.

This PR should be 100% backward compatible. TOC generation is optional (though activated by default in n64.mk), and DragonFS will respect hardcoded offsets passed to `dfs_init`: the TOC is inspected only when `DFS_DEFAULT_LOCATION` is specified.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ libdragonsys.a: $(BUILD_DIR)/system.o
 libdragon.a: $(BUILD_DIR)/n64sys.o $(BUILD_DIR)/interrupt.o \
 			 $(BUILD_DIR)/inthandler.o $(BUILD_DIR)/entrypoint.o \
 			 $(BUILD_DIR)/debug.o $(BUILD_DIR)/usb.o $(BUILD_DIR)/fatfs/ff.o \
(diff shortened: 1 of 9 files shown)


rasky:
After reading this PR ten times, it finally clicked to me that this PR is breaking existing applications that do not use `n64.mk`. Those applications will upgrade libdragon but will not pick up automatically the change that adds `--toc` to the `n64tool` invocation. 

Now I think in general the stable branch should assume n64.mk usage to keep sanity, but this change is too wide and important to ignore that use case. I guess the code will have to be changed to fallback to the old default hardcoded address if the rompak is not found.

PR #345, opened 2023-02-03
debug: add support for reading from SD to SDRAM in flashcarts

rasky:
Normally, reading from SD cards involve first issuing a flashcart-specific command to read SD contents into SDRAM (which is mapped in PI space as "ROM"), and then run a PI DMA transfer to copy contents into RDRAM, just like the data was in ROM.

There are use cases in which there is no need for the second half: the data can stay in SDRAM and be consumed from there. A common one is flashcart menus that want to load a ROM to boot, and don't need data to go into RDRAM (it wouldn't even fit).

This commit adds support to this use case by simply checking if the pointer provided to disk_read (and thus, to fread() by the user) is in RDRAM or in a PI-mapped space, and calling a new flashcart hook in the latter case.

This is implemented for now on 64drive and SC64.

### DIFF
--- a/src/debug_sdfs_sc64.c
+++ b/src/debug_sdfs_sc64.c
@@ -67,6 +67,15 @@ static DRESULT fat_disk_read_sc64(BYTE* buff, LBA_t sector, UINT count)
 	return RES_OK;
 }
 
+static DRESULT fat_disk_read_sdram_sc64(BYTE* buff, LBA_t sector, UINT count)
+{
+	_Static_assert(FF_MIN_SS == 512, "this function assumes sector size == 512");
+	_Static_assert(FF_MAX_SS == 512, "this function assumes sector size == 512");
+	if (sc64_sd_read_sectors((uint32_t)buff, sector, count))
+		return FR_DISK_ERR;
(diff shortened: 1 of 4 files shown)


src/debug_sdfs_sc64.c
```diff
@@ -67,6 +67,15 @@ static DRESULT fat_disk_read_sc64(BYTE* buff, LBA_t sector, UINT count)
 	return RES_OK;
 }
 
+static DRESULT fat_disk_read_sdram_sc64(BYTE* buff, LBA_t sector, UINT count)
+{
+	_Static_assert(FF_MIN_SS == 512, "this function assumes sector size == 512");
```
Polprzewodnikowy: This check seems to be the same for every `fat_disk_*flashcart*` function. Can it be moved up to `disk_read`/`disk_write` function?

rasky: they are static assert, I think the idea was to pinpoint functions where the assumption was being made. I'll move them to disk_read/disk_write as you suggest as all implementation seem to somehow assume that anyway

PR #346, opened 2023-02-03
entrypoint: trigger PI DMA transfer only for ROMs bigger than 1MB

Polprzewodnikowy:


anacierdem (approved):
LGTM

anacierdem:
I think we can even have this on trunk, I don't see why not

rasky:
Merged manually to trunk

PR #348, opened 2023-02-07
Batch of fixes for a few regressions in latest backports from unstable

rasky:


src/rspq/rspq.c
```diff
@@ -489,9 +492,9 @@ static void rspq_get_current_ovl(rsp_queue_t *rspq, int *ovl_idx, const char **o
 /** @brief RSPQ crash handler. This shows RSPQ-specific info the in RSP crash screen. */
 static void rspq_crash_handler(rsp_snapshot_t *state)
 {
-    rsp_queue_t *rspq = (rsp_queue_t*)state->dmem;
+    rsp_queue_t *rspq = (rsp_queue_t*)(state->dmem + RSPQ_DATA_ADDRESS);
     uint32_t cur = rspq->rspq_dram_addr + state->gpr[28];
-    uint32_t dmem_buffer = RSPQ_DEBUG ? 0x140 : 0x100;
+    uint32_t dmem_buffer = RSPQ_DEBUG ? 0x1A0 : 0x100;
```
snacchus, line 497: We still don't have a good way to extract ucode symbols, right? Maybe we should at least put these two values into constants (since they are used again further down)

include/rsp_queue.inc
```diff
@@ -242,9 +242,10 @@ RSPQ_DefineCommand RSPQCmd_SwapBuffers,     12    # 0x07
 RSPQ_DefineCommand RSPQCmd_TestWriteStatus, 8     # 0x08 -- must be even (bit 24 must be 0)
 
 #if RSPQ_DEBUG
+    .align 3
 RSPQ_LOG_IDX:                .long 0
 RSPQ_LOG:                    .ds.l 16
-RSPQ_LOG_END:                .long 0xFFFFFFFF
+RSPQ_LOG_END:                .long 0xABCD0123
```
snacchus: Consider defining the marker value as a constant somewhere

snacchus (changes requested):
Looks good to me! I only left two minor comments.

I ran the testrom locally and while the test you adjusted now runs correctly, I noticed that `test_constructors` always causes a FPU exception. The reason is unrelated to your changes and quite simple: It's missing the `TEST_FLAGS_NO_BENCHMARK` in it's definition, which leads to a division-by-zero error (because the expected time is defined as 0). Maybe we should guard against this mistake by checking the expected time if `TEST_FLAGS_NO_BENCHMARK` is not set.

src/rspq/rspq.c
rasky: OK I'll do that on unstable and port it to trunk. rspq on unstable had a refactoring that shuffled things around. I want to merge that first otherwise I always need to do changes twice.

include/rsp_queue.inc
rasky: Done

rasky:
It looks like the problem with `test_constructor` is because the test is too fast: the testsuite divides durations by 1024, and test_constructors' duration becomes 0, causing a later divide by zero. The workaround you suggested still stands though, I've pushed it.

PR #349, opened 2023-02-13
Misc improvements to interrupt/exceptions (merge from unstable)

rasky:
This PR contains multiple separate commits that bring several improvements to interrupt.h / exception.h:

 * Move pre-NMI functions to interrupt.h where they belonged
 * Improve exception names shown in crash screens to make them more user friendly
 * register_exception_handler now returns the previous handler
 * Add API to register syscall handlers

There is also a small unrelated commit to the build system, to allow split installations of toolchain and libdragon (without which, other merges from unstable are harder).

These are prerequisites to merge the backtracing feature of the unstable branch.


include/exception.h
```diff
@@ -105,32 +107,36 @@ typedef struct
...
+ */
+typedef void (*syscall_handler_t)(exception_t *exc, uint32_t code);
 
-void register_exception_handler( void (*cb)(exception_t *) );
+exception_handler_t register_exception_handler( exception_handler_t cb );
 void exception_default_handler( exception_t* ex );
 
-void register_reset_handler( void (*cb)(void) );
-uint32_t exception_reset_time( void );
+void register_syscall_handler( syscall_handler_t cb, uint32_t first_code, uint32_t last_code );
```
anacierdem, line 139: Why donwe need a range here? If we really want a mechanism for representing multiple codes, would a mask be a better option?

include/interrupt.h
```diff
@@ -56,10 +58,33 @@ void set_SI_interrupt( int active );
...
+ * resetting. See #exception_reset_time for more details.
+ * 
+ * @note The general knowledge about this is that the reset time should be
+ *       500 ms. Testing on different consoles show that, while most seem to
+ *       reset after 500 ms, a few EU models reset after 200ms. So we define
+ *       the timer shorter for greater compatibility.
+ */
+#define RESET_TIME_LENGTH      TICKS_FROM_MS(200)
+
+uint32_t exception_reset_time( void );
```
anacierdem, line 77: I think we may need to document this in the wiki as someone might have imported the other header already

n64.mk
```diff
@@ -7,23 +7,26 @@ N64_ROM_SAVETYPE = # Supported savetypes: none eeprom4k eeprom16 sram256k sram76
 N64_ROM_RTC = # Set to true to enable the Joybus Real-Time Clock
 N64_ROM_REGIONFREE = # Set to true to allow booting on any console region
 
+# Override this to use a toolchain installed separately from libdragon
```
anacierdem, line 10: How will this help the later changes from unstable?

src/exception.c
```diff
@@ -289,103 +351,85 @@ void __onCriticalException(volatile reg_block_t* regs)
...
-	for (int i=0;i<MAX_RESET_HANDLERS;i++)
-	{		
-		if (!__prenmi_handlers[i])
+	assertf(first_code <= 0xFFFFF, "The maximum allowed syscall code is 0xFFFFF (requested: %05lx)\n", first_code);
+	assertf(last_code <= 0xFFFFF, "The maximum allowed syscall code is 0xFFFFF (requested: %05lx)\n", first_code);
+	assertf(first_code <= last_code, "Invalid range for syscall handler (first: %05lx, last: %05lx)\n", first_code, last_code);
+
+	for (int i=0;i<MAX_SYSCALL_HANDLERS;i++)
+	{
+		if (!__syscall_handlers[i].handler)
```
anacierdem, line 386: Should we assert if it is registered a second time?

tests/test_exception.c
```diff
@@ -329,4 +329,26 @@ void test_exception(TestContext *ctx) {
...
 #undef ASSERT_REG_GP_HANDLER
-#undef ASSERT_REG
\ No newline at end of file
+#undef ASSERT_REG
+
+
+static volatile bool tsh_called = true;
+static volatile uint32_t tsh_code = 0;
+
+void test_syscall_handler(exception_t *exc, uint32_t code) {
```
anacierdem, line 338: Is there a reason why we don't use the gcc extension here? Is the stack destroyed or something?

anacierdem (commented):
A few suggestions and questions to discuss. Thank you for the improvements and additions @rasky 🎉

include/exception.h
rasky: I actually had implemented with a mask before, but it was harder to read and use. The idea is that if you are going to use a syscall, you might need a few variations around it. For instance, the inspector in the unstable branch currently uses 3 syscalls which are three different entry points (MIPS exceptions, assertions, C++ exceptions). A range is simpler to understand and read in my opinion.

include/interrupt.h
rasky: Actually I think the only public API is including `libdragon.h`. I always considered moving stuff across include files a non-breaking change. We might want to document this better, but all our examples and all user code I have ever seen only includes libdragon.h.

n64.mk
rasky: It doesn't actually, it's just to help me to avoid conflicts when merging n64.mk.

src/exception.c
rasky: Good idea, we should possibly check if the range is in conflict with an existing range. I will add this.

include/exception.h
anacierdem: Might be easier to read but I thought we can save a few ops. From readability standpoint, ranges might be confusing as well b/c of inclusive/exclusive ranges and not possible to represent ranges with holes in it. I don't have a too strong of an opinion though.

tests/test_exception.c
rasky: you mean the nested function? No there isn't a good reason, I could do that. I figured out that, if for any reason (= a bug in libdragon) the syscall handler remains registered to a nested function, it might lead to a hard-to-diagnose crash if it gets invoked, as the stack would have been destroyed.

n64.mk
anacierdem: I don't know how we use it but I can check for myself

include/exception.h
rasky: Speedwise, a few cycles per syscall isn't a problem, we don't have any plan to use syscalls in code sensitive codepaths at the moment. Consider that a range is checked with:

```
  if (code >= min && code <= max)
```

which gets compiled to

```
  if (code-min >= max-min)
```
so it's 3 cycles.

A mask is:

```
  if (code & mask == value)
```

which is 2 cycles. Honestly not a big deal.

I have two other alternatives in mind:
1) Not declare a range, and invoke all syscall handlers always with the code. Each handler will check if they like the code and do something. This is the simplest API-wise but we don't get to detect conflicts.
2) Do like we do in rspq and settle on a fixed split. Since the syscall code is 20 bit, we could decide to split 8+12 for instance, and simply let register on a single 8 bit prefix, and forward to it all codes with that prefix.

n64.mk
rasky: The usecase is coming from the Arch Linux mips64 package which is trying to be a "generic mips64 compiler" that can be used also for romhacks or libultra work, or even non-N64 work. So it is installed under `/usr` like a regular package. Then on top of that there is a different "libdragon" package which gets installed to `/opt/libdragon`.

include/exception.h
anacierdem: Yeah performance-wise a single cycle is not important. I liked 1 the most and maybe we can solve the other issue by returning a bool from the handler which is set if it is handled. If it is handled multiple times we can fail. Though still the return value does not necessarily reflect the reality. OTOH this may open the possibility of multiple handlers if that ever make sense? The only thing that might be useful is some kind of a validator detached from the actual handlers but we could do that in the handler itself anyways. And still performance will degrade based on the # of handlers.

Even though I didn't like the range api, I don't have a better alternative. Accepting a predicate per handler would definitely be overkill. Maybe we can ask discord? (I am still not objecting to the range, just making sure we explore it - which clearly you already did btw)

src/exception.c
```diff
@@ -253,7 +265,57 @@ static const char* __get_exception_name(exception_code_t code)
...
+			return "Floating point underflow";
+		} else if (ex->regs->fc31 & C1_CAUSE_INEXACT_OP) {
+			return "Floating point inexact operation";
+		} else {
+			return "Generic floating point";
+		}
+	case EXCEPTION_CODE_TLB_LOAD_I_MISS:
+		if (epc == badvaddr) {
+			return "Invalid program counter address";
+		} else if (badvaddr < 128) {
```
snacchus, line 292: Why does it check `< 128` for null reference? (same below)

rasky: it's just a heuristic to catch null pointers. One common case is that you have a NULL pointer to struct, and deference one of the fields of the struct. In that case you don't get a TLB miss on exactly 0x00000000, but the address is basically the offset of the field in the struct. If the struct is larger you might instead get `Read from invalid memory address`, which is not so bad too, so I guess that's an acceptable compromise.

include/exception.h
rasky: Honestly this API isn't going to be used so much I think. It could even be kept private for now, I doubt there are many use cases for libraries to use syscalls. 

Basically the main advantage of a syscall handler is that you get a snapshot of the state of all registers at that point, so it must be some advanced / debugging feature. We have two uses for now: one is the inspector (which, being a "debugger", requires a snapshot of the state), and another is the kernel (which does a context switch so it's handy if the thread has been already saved onto the stack).

I can't think of another case right now, but it must be something this low-level. So I don't think we're getting to a point where there are dozens of users of this API and there will be conflicts in managing ranges. So personally I think "anything goes" here. I wouldn't want to switch back to the mask because I tried that and it was worse readability wise, much harder to understand. I don't have a strong opinion on the other options though, I think anything goes for the few uses we will have of this.

anacierdem: I am totally ok 👍🏼

src/exception.c
anacierdem: Maybe we can incorporate "struct" in the message when it is not exactly zero to further help?

rasky: Unfortunately you can get a similar result if you dereference a NULL array (such as `m[2][2]` for a matrix), so explicitly mentioning "struct" could bring more confusion than it solves. Do you think it's confusing for users? I can understand being confusing for us reading the source code, but the user gets a "NULL pointer dereference" on a specific source code line, and I guess they can expect any pointer there to cause that, irrespective if it's an array or a struct.

anacierdem: Haven't thought of the array, right. I don't think it is particularly confusing. Just adding a comment will clear things for readers for sure.

rasky: done

src/exception.c
rasky: done

PR #352, opened 2023-02-23
Retrieve ROM address from file descriptor.

giulianobelinassi:
This commit adds a new function which allows the user to retrieve the ROM address of a file given its file descriptor.  This is useful if the user is using some virtual filesystem and wants to extract some extra performance by bypassing newlib read functions at all, or even dfs_read at all if it is known that things are well behaved.

Note: One can use
```
dfs_rom_addr_of_file(fp->_handle - 2);
```
to retrieve the rom address of a FILE* object.

### DIFF
--- a/include/dragonfs.h
+++ b/include/dragonfs.h
@@ -94,6 +94,7 @@ int dfs_tell(uint32_t handle);
 int dfs_close(uint32_t handle);
 int dfs_eof(uint32_t handle);
 int dfs_size(uint32_t handle);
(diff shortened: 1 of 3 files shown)


rasky:
Superseded by https://github.com/DragonMinded/libdragon/pull/589.

We have now implemented `ioctl` in libdragon and started with a `ioctl` to find out the ROM base address of a file. This not only covers the functionality requested in this PR, but also allows to dynamically check whether a `FILE*` (or a file descriptor) is in ROM or not, which is required in many situations.

PR #354, opened 2023-02-24
Fix some typos, some clarifications, and questions

Dragorn421:
I read through some of the lowest-level parts of libdragon out of curiosity. I'm very familiar with libultra and wanted to see how it compared
I must say it was a pleasant read, the comments are very helpful without overdoing it. (and in general it's way cleaner than libultra)

I found a few typos and took the opportunity to also clarify some things I thought could use more details, ofc feel free to suggest reverting

There are some (many?) questions sprinkled in (most with a "TODO" alongside), I figured it would be less chaotic than Discord to use Github reviews to discuss them

(I'm in the N64brew discord btw, `Dragorn421#0750`, if it's more convenient)

### DIFF
--- a/include/system.h
+++ b/include/system.h
@@ -44,7 +44,7 @@ typedef struct
      * @return A pointer to an arbitrary file handle assigned by the filesystem code 
      *         or NULL on error.
      */
-    void *(*open)( char *name, int flags );
+    void *(*open)( const char *name, int flags );
     /** 
      * @brief Function to call when performing a fstat command
      *
--- a/include/system.h
+++ b/include/system.h
@@ -44,7 +44,7 @@ typedef struct
      * @return A pointer to an arbitrary file handle assigned by the filesystem code 
      *         or NULL on error.
      */
-    void *(*open)( const char *name, int flags );
+    void *(*open)( char *name, int flags );
     /** 
      * @brief Function to call when performing a fstat command
      *
--- a/src/entrypoint.S
+++ b/src/entrypoint.S
@@ -43,7 +43,7 @@ set_sp:
 	mtc0 $0,C0_CAUSE
 
 	/* Check if PI DMA transfer is required,
-	   knowing that IPL3 loads 1MB of ROM to RAM,
+	   knowing that IPL3 loads 1 MiB of ROM to RAM,
 	   and __libdragon_text_start is located
 	   right after the ROM header where this 1 MiB starts. */
 	la a0, __libdragon_text_start
@@ -69,7 +69,7 @@ set_sp:
 .Lskip_dma:
 	/* fill .bss with 0s */
 	la a0, __bss_start
-	or a0, 0x20000000           /* with __bss_start being KSEG0 this makes the address KSEG1 (uncached) ? */
+	or a0, 0x20000000           /* convert address to KSEG1 (uncached) */
 	la a1, __bss_end
 	or a1, 0x20000000
 bss_init:
--- a/src/entrypoint.S
+++ b/src/entrypoint.S
@@ -45,10 +45,10 @@ set_sp:
 	/* Check if PI DMA transfer is required,
 	   knowing that IPL3 loads 1MB of ROM to RAM,
 	   and __libdragon_text_start is located
-	   right after the ROM header where this 1MB starts. */
+	   right after the ROM header where this 1 MiB starts. */
 	la a0, __libdragon_text_start
 	la a1, __data_end
-	li t0, 0x100000				/* stock IPL3 load size (1MB) */
+	li t0, 0x100000				/* stock IPL3 load size (1 MiB) */
 	subu a2, a1, a0				/* calculate data size */
 	sub a2, a2, t0				/* calculate remaining data size */
 	blez a2, .Lskip_dma			/* skip PI DMA if data is already loaded */
--- a/src/entrypoint.S
+++ b/src/entrypoint.S
@@ -42,10 +42,13 @@ set_sp:
 	mtc0 v0,C0_SR
 	mtc0 $0,C0_CAUSE
 
-	/* Check if PI DMA transfer is required */
+	/* Check if PI DMA transfer is required,
+	   knowing that IPL3 loads 1MB of ROM to RAM,
+	   and __libdragon_text_start is located
+	   right after the ROM header where this 1MB starts. */
 	la a0, __libdragon_text_start
 	la a1, __data_end
-	li t0, 0x100000				/* stock IPL3 load size */
+	li t0, 0x100000				/* stock IPL3 load size (1MB) */
 	subu a2, a1, a0				/* calculate data size */
 	sub a2, a2, t0				/* calculate remaining data size */
 	blez a2, .Lskip_dma			/* skip PI DMA if data is already loaded */
@@ -58,15 +61,15 @@ set_sp:
 
 	/* Start PI DMA transfer */
 	lui t0, 0xA460
-	sw a0, 0x00(t0)
-	sw a1, 0x04(t0)
+	sw a0, 0x00(t0)             /* PI_DRAM_ADDR */
+	sw a1, 0x04(t0)             /* PI_CART_ADDR */
 	addi a2, -1
-	sw a2, 0x0C(t0)
+	sw a2, 0x0C(t0)             /* PI_WR_LEN */
 
 .Lskip_dma:
 	/* fill .bss with 0s */
 	la a0, __bss_start
-	or a0, 0x20000000
+	or a0, 0x20000000           /* with __bss_start being KSEG0 this makes the address KSEG1 (uncached) ? */
 	la a1, __bss_end
 	or a1, 0x20000000
 bss_init:
@@ -78,8 +81,8 @@ bss_init:
 	/* Wait for DMA transfer to be finished */
 	lui t0, 0xA460
 wait_dma_end:
-	lw t1, 0x10(t0)
-	andi t1, 3
+	lw t1, 0x10(t0)             /* PI_STATUS */
+	andi t1, 3                  /* PI_STATUS_DMA_BUSY | PI_STATUS_IO_BUSY */
 	bnez t1, wait_dma_end
 	nop
 
--- a/src/exception.c
+++ b/src/exception.c
@@ -364,6 +364,7 @@ uint32_t exception_reset_time( void )
 }
 
 
+// TODO: why "exception" ?
 /**
  * @brief Respond to a reset exception.
  * 
@@ -371,7 +372,7 @@ uint32_t exception_reset_time( void )
  */
 void __onResetException( volatile reg_block_t* regs )
 {
-	/* This function will be called many times becuase there is no way
+	/* This function will be called many times because there is no way
 	   to acknowledge the pre-NMI interrupt. So make sure it does nothing
 	   after the first call. */
 	if (__prenmi_tick) return;
--- a/src/exception.c
+++ b/src/exception.c
@@ -364,7 +364,6 @@ uint32_t exception_reset_time( void )
 }
 
 
-// TODO: why "exception" ?
 /**
  * @brief Respond to a reset exception.
  * 
--- a/src/interrupt.c
+++ b/src/interrupt.c
@@ -771,7 +771,7 @@ void disable_interrupts()
  */
 void enable_interrupts()
 {
-    /* Don't do anything if we've hosed up (TODO: what does this mean?) or aren't initialized */
+    /* Don't do anything if we aren't initialized */
     if( __interrupt_depth < 0 ) { return; }
 
     /* Check that we're not calling enable_interrupts() more than expected */
--- a/src/interrupt.c
+++ b/src/interrupt.c
@@ -572,8 +572,8 @@ void set_AI_interrupt(int active)
  *
  * The VI interrupt is generated when the VI begins displaying a specific line
  * of the display output. The line number configured always refers to the
- * final TV output, so it should be either in the range 0..479 (NTSC) or
- * 0..575 (PAL).
+ * final TV output, so it should be either in the range 0..524 (NTSC) or
+ * 0..624 (PAL).
  * The vblank happens at the beginning of the display period, in range
  * 0..33 (NTSC) or 0..43 (PAL). A common value used to trigger the interrupt
  * at the beginning of the vblank is 2.
--- a/src/interrupt.c
+++ b/src/interrupt.c
@@ -32,10 +32,9 @@
  * #disable_interrupts is considered a violation of this assumption
  * and should be avoided.  Calling #disable_interrupts when interrupts
  * are already disabled will have no effect interrupts-wise
- * (but should be paired with a #enable_interrupts regardless).
- * Calling #enable_interrupts again to restore from a critical section
- * will not enable interrupts if interrupts were not enabled when
- * calling #disable_interrupts.
+ * (but should be paired with a #enable_interrupts regardless),
+ * and in that case the paired #enable_interrupts will not enable
+ * interrupts either.
  * In this manner, it is safe to nest calls to disable and enable
  * interrupts.
  *
--- a/src/interrupt.c
+++ b/src/interrupt.c
@@ -31,9 +31,11 @@
  * in pairs.  Calling #enable_interrupts without first calling
  * #disable_interrupts is considered a violation of this assumption
  * and should be avoided.  Calling #disable_interrupts when interrupts
- * are already disabled will have no effect.  Calling #enable_interrupts
- * again to restore from a critical section will not enable interrupts
- * if interrupts were not enabled when calling #disable_interrupts.
+ * are already disabled will have no effect interrupts-wise
+ * (but should be paired with a #enable_interrupts regardless).
+ * Calling #enable_interrupts again to restore from a critical section
+ * will not enable interrupts if interrupts were not enabled when
+ * calling #disable_interrupts.
  * In this manner, it is safe to nest calls to disable and enable
  * interrupts.
  *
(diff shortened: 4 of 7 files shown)


src/entrypoint.S
```diff
@@ -42,10 +42,13 @@ set_sp:
...
 
-	/* Check if PI DMA transfer is required */
+	/* Check if PI DMA transfer is required,
+	   knowing that IPL3 loads 1MB of ROM to RAM,
+	   and __libdragon_text_start is located
+	   right after the ROM header where this 1MB starts. */
 	la a0, __libdragon_text_start
 	la a1, __data_end
-	li t0, 0x100000				/* stock IPL3 load size */
+	li t0, 0x100000				/* stock IPL3 load size (1MB) */
```
sp1187: We try to use IEC binary prefixes consistently in the documentation to avoid ambiguity, so this should be (1 MiB) instead. It is also recommended to put a space between the number and the unit. Same with the description comment above.

sp1187 (changes requested):

src/entrypoint.S
```diff
@@ -58,15 +61,15 @@ set_sp:
...
+	sw a1, 0x04(t0)             /* PI_CART_ADDR */
 	addi a2, -1
-	sw a2, 0x0C(t0)
+	sw a2, 0x0C(t0)             /* PI_WR_LEN */
 
 .Lskip_dma:
 	/* fill .bss with 0s */
 	la a0, __bss_start
-	or a0, 0x20000000
+	or a0, 0x20000000           /* with __bss_start being KSEG0 this makes the address KSEG1 (uncached) ? */
```
rasky: Correct. I would just say:

```suggestion
	or a0, 0x20000000           /* convert address to KSEG1 (uncached) */
```

src/exception.c
```diff
@@ -364,14 +364,15 @@ uint32_t exception_reset_time( void )
 }
 
 
+// TODO: why "exception" ?
```
rasky: It's been a mistake. I'm about to merge https://github.com/DragonMinded/libdragon/pull/349 that rectifies this.

src/interrupt.c
```diff
@@ -31,9 +31,11 @@
...
  * #disable_interrupts is considered a violation of this assumption
  * and should be avoided.  Calling #disable_interrupts when interrupts
- * are already disabled will have no effect.  Calling #enable_interrupts
- * again to restore from a critical section will not enable interrupts
- * if interrupts were not enabled when calling #disable_interrupts.
+ * are already disabled will have no effect interrupts-wise
+ * (but should be paired with a #enable_interrupts regardless).
+ * Calling #enable_interrupts again to restore from a critical section
+ * will not enable interrupts if interrupts were not enabled when
+ * calling #disable_interrupts.
```
rasky: I don't mind your suggestion (is fine) but the whole comment sounds a bit mouthful. I would say 99% of programmers reading this would be familiar with the concept of "nested locks" or "reentrant locks", so maybe we could find a simplified wording? Anyway if you prefer to keep it like this, it's fine.

src/interrupt.c
```diff
@@ -574,6 +576,7 @@ void set_AI_interrupt(int active)
  * @param[in] line
  *            The vertical line that causes this interrupt to fire.  Ignored
  *            when setting the interrupt inactive
+ *            (TODO: https://n64brew.dev/wiki/Video_Interface#0x0440_000C_-_VI_V_INTR says "When VI_V_CURRENT reaches this half-line number", so line or half-line?)
```
rasky: It's the so-called half-line, which is a term that I've always found confusing. If we want to explain it, then we should attempt something like this (in the function doc body):

```
The VI interrupt is generated when the VI begins displaying a specific line
of the display output. The line number configured it always refers to the
final TV output, so it should be either in the range 0..479 (NTSC) or
0..575 (PAL). The vblank happens at the beginning of the display period, in
range 0..33 (NTSC) or 0..43 (PAL). A common value used to trigger the
interrupt at the beginning of the vblank is 2.

In non-interlaced modes, the VI only draws on even lines, so configuring the
interrupt on an odd line causes the interrupt to never trigger. In interlace
modes, instead, the VI alternates between even lines and odd lines, so any
specific line will trigger an interrupt only every other frame. If you need an
interrupt every frame in interlaced mode, you will need to reconfigure the
interrupt every frame, alternating between an odd and an even number.
```

src/interrupt.c
```diff
@@ -763,12 +766,12 @@ void disable_interrupts()
...
  * @note If this is called inside a nested disable call, it will have no effect on the
- *       system.  Therefore it is safe to nest disable/enable calls.  After the last
- *       nested interrupt is enabled, systemwide interrupts will be reenabled.
+ *       system.  Therefore it is safe to nest disable/enable calls.  After the least
+ *       nested enable call, systemwide interrupts will be reenabled.
  */
 void enable_interrupts()
 {
-    /* Don't do anything if we've hosed up or aren't initialized */
+    /* Don't do anything if we've hosed up (TODO: what does this mean?) or aren't initialized */
```
rasky: I think the idea is "don't do anything if the interrupt depth is negative as all bets are off".

I think the comment and the if line is a relic from the past when libdragon didn't use assertions. Back at a time, if one called `enable_interrupts` at beginning of main by mistake (assuming it had to be done), it would forever run with interrupts disabled because of this and other similar lines. Now, it triggers an assertion instead, so this checks are not useful anymore (I think, but it should be easy to re-review this code and make sure what I wrote is correct).

In general, we switched approach and instead of silently doing nothing in bad situations, we now complain loudly. In fact, we consider a bug whenever libdragon freezes with no error messages.

src/inthandler.S
```diff
@@ -81,9 +81,13 @@ inthandler:
 	mtc0 k1, C0_SR
 
 	# WARNING: it is now possible to trigger reentrant exceptions (and not only
-	# crashing one. Avoid using k0/k1 from now on, as they would get corrupted
+	# crashing ones). Avoid using k0/k1 from now on, as they would get corrupted
 	# by a reentrant exception.
-	#define cause t8
+	# TODO: would like more details, because a bunch of other stuff mentions no reentrant interrupts
+	#       (also, prenmi?)
+	#       -> this is probably about cp1 (fpu) unusable exception, if the fpu is first used in a interrupt handler
```
rasky: In this case, the comment says and means "reentrant exceptions". After we write to `C0_SR` to disable the `EXL` bit, the CPU can retrigger an exception at any point. This can either be the FPU exception, a syscall or even just a bug in the code causing a crash (eg: NULL pointer dereference). In the latter case, we're going to halt so it doesn't really matter if we corrupt k0/k1 on the way to the crash screen, but the idea of this exception handler is to be resistant to nested exceptions for use cases where you actually need to recover from them. Using k0/k1 would prevent nested exceptions from working correctly.

Feel free to suggest a way to reword the comment if you feel it's not clear.

src/inthandler.S
```diff
@@ -190,10 +197,10 @@ interrupt:
 	and t1, t0
 	sw t1, STACK_SR(sp)
 
-	# Reload cause register (might be reused by C code) and test for other interrupts
+	# Reload cause register (might be reused by C code (TODO: what?)) and test for other interrupts
```
rasky: We just called into C (`jal __onResetException`) and C code can modify `t8` as it pleases. So we need to reload `cause` (which is `t8`) from the stack.

src/inthandler.S
```diff
@@ -358,6 +365,6 @@ save_fpu_regs:
 
 
 	.section .bss
- 	.align 8
+ 	.align 8 # TODO: align to 256 bytes, is that intended?
```
rasky: No I'd say it's a bug, I guess we could even remove the line but I'm not sure about the semantics of `.lcomm` so maybe let's keep at least a 4-byte alignment

```suggestion
        .p2align 2
```

src/system.c
```diff
@@ -9,6 +9,7 @@
 #include <errno.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/time.h> // TODO: for #include <sys/_timeval.h>
```
rasky: I don't understand this TODO?

src/system.c
```diff
@@ -641,7 +646,7 @@ int execve( char *name, char **argv, char **env )
 void _exit( int rc )
 {
     /* Default stub just causes a divide by 0 exception.  */
-    int x = rc / INT_MAX;
+    int x = rc / INT_MAX; // TODO: what if rc == INT_MAX
```
rasky: Uhm I'm not sure we ever get here. Surely a division by zero is not the correct thing to do, as it would cause an exception screen to show up. I think it would be better to just remove this code here and loop forever like we do on `abort()`.

There has been some discussion in the emulation community to standardize a way for homebrew code to declare the "end" of a program, so that the emulator could in theory exit; this would be useful when using emulators to run testsuites for instance, even more so in headless mode. But we haven't reached a consensus.

src/system.c
```diff
@@ -1038,7 +1043,7 @@ void *sbrk( int incr )
 int stat( const char *file, struct stat *st )
 {
     /* Dirty hack, open read only */
-    int fd = open( (char *)file, 0, 777 );
+    int fd = open( (char *)file, 0, /* octal: */ 0777 ); // TODO: mode is ignored in libdragon and anyway isn't used in the standard for read only, so just pass 0?
```
rasky: Yeah, maybe just:

```suggestion
    int fd = open( (char *)file, O_RDONLY, 0 );
```

rasky (commented):
Thank you very much for this PR! It's great to have people review so carefully the code and even providing improvements to the docs. I've left a few comments on your TODOs, but I'd say most of the changes are ok.

src/inthandler.S
sp1187: You could explicitly choose `.balign` (bytes) or `.p2align` (powers of two) to make this more explicit.

src/entrypoint.S
```diff
@@ -42,10 +42,13 @@ set_sp:
 	mtc0 v0,C0_SR
 	mtc0 $0,C0_CAUSE
 
-	/* Check if PI DMA transfer is required */
+	/* Check if PI DMA transfer is required,
+	   knowing that IPL3 loads 1MB of ROM to RAM,
```
sp1187: Missed this instance of "1MB".

rasky:
```suggestion
	   knowing that IPL3 loads 1 MiB of ROM to RAM,
```

src/inthandler.S
rasky: Edited suggestion to use p2align

Dragorn421: https://sourceware.org/binutils/docs-2.40/as.html#Align
> GAS also provides .balign and .p2align directives, described later, which have a consistent behavior across all architectures (**but are specific to GAS**). 

Is it fine to only support / rely on GNU as?

rasky: libdragon only compiles with its own toolchain, which is a modern GCC. So yes.

src/system.c
Dragorn421: Sorry it was basically not explained

My IDE is unhappy about `struct timeval` being used in this file, saying it's not defined

I don't know why compilation still works

the man page on my system says (`man timeval`):
```
       timeval
              Include: <sys/time.h>.  Alternatively, <sys/resource.h>, <sys/select.h>, or <utmpx.h>.

              struct timeval {
                  time_t      tv_sec;  /* Seconds */
                  suseconds_t tv_usec; /* Microseconds */
              };
```

system.c has none of these includes

again idk why compilation doesn't fail

src/system.c
Dragorn421: isn't `_exit`, the "implementation" of abort?
https://sourceware.org/newlib/libc.html#Stubs
https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/stdlib/abort.c;hb=main#l59

(my initial point with "what if rc == INT_MAX" is it wouldn't cause a `/0` but it seems it's overshadowed by the above :thinking: )

src/interrupt.c
Dragorn421: Yeah it's weird but just

> Calling #disable_interrupts when interrupts are already disabled will have no effect

may be misleading in that it still pushes the "depth" down by one, so it does have a side effect

--------

I haven't changed this paragraph, do you mean it could be trimmed down?
Maybe continue the sentence like,

>  \* , and in that case the paired #enable_interrupts will not enable interrupts either.

src/interrupt.c
Dragorn421: thanks for the info (if that's not on the wiki it should probably be added somewhere)

However I don't see where "half line" comes in?

src/interrupt.c
Dragorn421: (dang this comment was written in 2012)

would
```suggestion
    /* Don't do anything if we aren't initialized */
```
be fine like the one in `disable_interrupts`, then
other comments indeed just refer to normal operation, not taking bugs into account

src/inthandler.S
Dragorn421: oh I didn't understand it that way at all
would "clobbered" instead of "reused" be fine?
```suggestion
	# Reload cause register (might be clobbered by C code) and test for other interrupts
```
or even
```suggestion
	# Reload cause register (might be clobbered by external handlers) and test for other interrupts
```

src/inthandler.S
Dragorn421: I changed the other `.align`s in inthandler.S to `.p2align`, there are others elsewhere I left alone (just to try to be consistent in separate files)

src/system.c
Dragorn421: lol here is the assembly of `_exit`:
```
Disassembly of section .text._exit:

00000000 <_exit>:
   0:   1000ffff        b       0 <_exit>
   4:   00000000        nop
        ...
```

(
```
cd $N64_INST/mips64-elf/lib
$N64_INST/bin/mips64-elf-objdump libdragonsys.a -d -j.text._exit
```
)

big brain gcc

src/interrupt.c
rasky: I think it's coming from the fact that, given a standard resolution of 320x240, the value to be put in the register is two times the screen line (plus some offset); so it's like the VI registers refers to "half-lines" of the 240 resolution.

But the reality is that the VI register counts the actual display lines, and 320x240 is upscaled to that.

src/system.c
rasky: Ok so I'd remove the division and just keep the infinite loop

src/system.c
```diff
@@ -848,11 +850,11 @@ int lseek( int file, int ptr, int dir )
  * @param[in] flags
  *            Flags specifying open flags, such as binary, append.
  * @param[in] mode
- *            Mode of the file.
+ *            Mode of the file. (TODO: currently ignored)
```
Dragorn421: idk what to say about this, leave it as a TODO ?

rasky: I guess you can keep it without TODO:

```suggestion
 *            Mode of the file (currently ignored).
```

src/system.c
rasky: I'd remove this then, it seems more a problem of the IDE than a libdragon issue

src/interrupt.c
```diff
@@ -569,11 +570,29 @@ void set_AI_interrupt(int active)
 /**
  * @brief Enable or disable the VI interrupt
  *
+ * The VI interrupt is generated when the VI begins displaying a specific line
+ * of the display output. The line number configured always refers to the
+ * final TV output, so it should be either in the range 0..479 (NTSC) or
+ * 0..575 (PAL).
```
rasky:
```suggestion
 * final TV output, so it should be either in the range 0..524 (NTSC) or
 * 0..624 (PAL).
```

I got them wrong the first time, these are the correct ranges (including vblank).

Dragorn421: I'm having a hard time finding a neat primary source but I've seen in various places (e.g. https://www.sony.com/electronics/support/articles/00006681 )
that there are 525 (NTSC) and 625 (PAL) lines
0-indexed that would be 0..524 and 0..624 ?

(if you have a proper spec of NTSC/PAL handy I'd like to see it)

include/system.h
```diff
@@ -44,7 +44,7 @@ typedef struct
      * @return A pointer to an arbitrary file handle assigned by the filesystem code 
      *         or NULL on error.
      */
-    void *(*open)( char *name, int flags );
+    void *(*open)( const char *name, int flags );
```
Dragorn421: because I changed `open` from
`int open( char *file, int flags, int mode )`
to
`int open( const char *file, int flags, int mode )`
to match the standard
(including `fcntl.h` (for `O_RDONLY`) revealed that mistake)
and this is called by `open` using that `name`

src/interrupt.c
sp1187: Yes, it should be 0..524 and 0..624. I tried to poll the VI current line register in my PAL console, and the highest value I got was 624 like you would expect.

include/system.h
rasky: We can't merge this as this breaks backward compatibility with existing filesystem implementations. You will have to cast your way through this.

Dragorn421: Would copying the string be fine or too much of a cost?

rasky: I wouldn't do that for const correctness honestly

src/interrupt.c
rasky: Fixed

src/system.c
```diff
@@ -848,11 +850,11 @@ int lseek( int file, int ptr, int dir )
  * @param[in] flags
  *            Flags specifying open flags, such as binary, append.
  * @param[in] mode
- *            Mode of the file.
+ *            Mode of the file (currently ignored).
  *
  * @return File handle to refer to this file on success, or a negative value on error.
  */
-int open( char *file, int flags, int mode )
+int open( const char *file, int flags, ... )
```
Dragorn421, line 857: For reference:
https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/libc/syscalls/sysopen.c;hb=main
`man 2 open`
`/opt/libdragon/mips64-elf/include/sys/_default_fcntl.h`

PR #355, opened 2023-02-28
Implement stack walking (backtrace)

rasky:
Quoting the docs:

```
This module implements two POSIX/GNU standard functions to help walking
the stack and providing the current execution context: backtrace() and
backtrace_symbols().
 
The functions have an API fully compatible with the standard ones. The
implementation is however optimized for the MIPS/N64 case, and with
standard compilation settings. See the documentation in backtrace.c
for implementation details.
 
You can call the functions to inspect the current call stack. For
a higher level function that just prints the current call stack
on the debug channels, see #debug_backtrace.
```

An optimized symbol table is generated by a new tool (n64sym), which is then hooked up to the build system (n64.mk). The symbol table is then embedded in the ROM as rompak.

n64sym analyzes standard DWARF debugging symbol in the ELF file using tools from the toolchain, so that it doesn't have to embed a DWARF parser and is thus hopefully very solid to feature evolutions of these (ever changing) standards. It is quite fast (~100ms for a standard C ROM, ~1 sec for huge C++ ROMs that take forever to link). The symbol table is reasonably compact (100-200K up to 1M for huge ROMs) and is never loaded into ROM, but fully queried directly from ROM. Obviously it can be removed from production builds (backtracing still works without symbol table, but of course only hexadecimal addresses will be printed).

This feature has been extensively tested in the unstable channel and is quite stable and incredibly useful during development. This PR brings in the core functionality, it doesn't hook it up yet to assertions or exceptions.

### DIFF
--- a/n64.mk
+++ b/n64.mk
@@ -36,7 +36,8 @@ N64_SYM = $(N64_BINDIR)/n64sym
 N64_AUDIOCONV = $(N64_BINDIR)/audioconv64
 
 N64_CFLAGS =  -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
-N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections -g
+N64_CFLAGS += -falign-functions=32   # NOTE: if you change this, also change backtrace() in backtrace.c
+N64_CFLAGS += -ffunction-sections -fdata-sections -g -ffile-prefix-map=$(CURDIR)=
 N64_CFLAGS += -DN64 -O2 -Wall -Werror -Wno-error=deprecated-declarations -fdiagnostics-color=always
 N64_ASFLAGS = -mtune=vr4300 -march=vr4300 -Wa,--fatal-warnings -I$(N64_INCLUDEDIR)
 N64_RSPASFLAGS = -march=mips1 -mabi=32 -Wa,--fatal-warnings -I$(N64_INCLUDEDIR)
--- a/n64.mk
+++ b/n64.mk
@@ -32,14 +32,15 @@ N64_CHKSUM = $(N64_BINDIR)/chksum64
 N64_ED64ROMCONFIG = $(N64_BINDIR)/ed64romconfig
 N64_MKDFS = $(N64_BINDIR)/mkdfs
 N64_TOOL = $(N64_BINDIR)/n64tool
+N64_SYM = $(N64_BINDIR)/n64sym
 N64_AUDIOCONV = $(N64_BINDIR)/audioconv64
 
 N64_CFLAGS =  -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
-N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
+N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections -g
 N64_CFLAGS += -DN64 -O2 -Wall -Werror -Wno-error=deprecated-declarations -fdiagnostics-color=always
 N64_ASFLAGS = -mtune=vr4300 -march=vr4300 -Wa,--fatal-warnings -I$(N64_INCLUDEDIR)
 N64_RSPASFLAGS = -march=mips1 -mabi=32 -Wa,--fatal-warnings -I$(N64_INCLUDEDIR)
-N64_LDFLAGS = -L$(N64_LIBDIR) -ldragon -lm -ldragonsys -Tn64.ld --gc-sections --wrap __do_global_ctors
+N64_LDFLAGS = -g -L$(N64_LIBDIR) -ldragon -lm -ldragonsys -Tn64.ld --gc-sections --wrap __do_global_ctors
 
 N64_TOOLFLAGS = --header $(N64_HEADERPATH) --title $(N64_ROM_TITLE)
 N64_ED64ROMCONFIGFLAGS =  $(if $(N64_ROM_SAVETYPE),--savetype $(N64_ROM_SAVETYPE))
@@ -75,13 +76,14 @@ N64_CFLAGS += -std=gnu99
 %.z64: LDFLAGS+=$(N64_LDFLAGS)
 %.z64: $(BUILD_DIR)/%.elf
 	@echo "    [Z64] $@"
+	$(N64_SYM) $< $<.sym
 	$(N64_OBJCOPY) -O binary $< $<.bin
 	@rm -f $@
 	DFS_FILE="$(filter %.dfs, $^)"; \
 	if [ -z "$$DFS_FILE" ]; then \
-		$(N64_TOOL) $(N64_TOOLFLAGS) --toc --output $@ $<.bin; \
+		$(N64_TOOL) $(N64_TOOLFLAGS) --toc --output $@ $<.bin --align 8 $<.sym; \
 	else \
-		$(N64_TOOL) $(N64_TOOLFLAGS) --toc --output $@ $<.bin --align 16 "$$DFS_FILE"; \
+		$(N64_TOOL) $(N64_TOOLFLAGS) --toc --output $@ $<.bin --align 8 $<.sym --align 16 "$$DFS_FILE"; \
 	fi
 	if [ ! -z "$(strip $(N64_ED64ROMCONFIGFLAGS))" ]; then \
 		$(N64_ED64ROMCONFIG) $(N64_ED64ROMCONFIGFLAGS) $@; \
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -207,6 +207,7 @@ static symtable_header_t symt_open(void) {
  */
 static addrtable_entry_t symt_addrtab_entry(symtable_header_t *symt, int idx)
 {
+    assert(idx >= 0 && idx < symt->addrtab_size);
     return io_read(SYMT_ROM + symt->addrtab_off + idx * 4);
 }
 
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -78,9 +78,6 @@
  */
 #define FUNCTION_ALIGNMENT      32
 
-#define MAX_FILE_LEN 120        ///< Maximum length of a file name in a backtrace entry
-#define MAX_FUNC_LEN 120        ///< Maximum length of a function name in a backtrace entry
-
 /** 
  * @brief Symbol table file header
  * 
@@ -183,10 +180,9 @@ static symtable_header_t symt_open(void) {
         return (symtable_header_t){0};
     }
 
-    symtable_header_t symt_header;
+    symtable_header_t symt_header alignas(8);
     data_cache_hit_writeback_invalidate(&symt_header, sizeof(symt_header));
-    dma_read_raw_async(&symt_header, SYMT_ROM, sizeof(symtable_header_t));
-    dma_wait();
+    dma_read(&symt_header, SYMT_ROM, sizeof(symtable_header_t));
 
     if (symt_header.head[0] != 'S' || symt_header.head[1] != 'Y' || symt_header.head[2] != 'M' || symt_header.head[3] != 'T') {
         debugf("backtrace: invalid symbol table found at 0x%08lx\n", SYMT_ROM);
@@ -573,8 +569,8 @@ static void format_entry(void (*cb)(void *, backtrace_frame_t *), void *cb_arg,
     symtable_entry_t entry alignas(8);
     symt_entry_fetch(symt, &entry, idx);
 
-    char file_buf[MAX_FILE_LEN+2] alignas(8);
-    char func_buf[MAX_FUNC_LEN+2] alignas(8);
+    char file_buf[entry.file_len + 2] alignas(8);
+    char func_buf[MAX(entry.func_len + 2, 32)] alignas(8);
 
     cb(cb_arg, &(backtrace_frame_t){
         .addr = addr,
@@ -638,6 +634,8 @@ bool backtrace_symbols_cb(void **buffer, int size, uint32_t flags,
 
 char** backtrace_symbols(void **buffer, int size)
 {
+    const int MAX_FILE_LEN = 120;
+    const int MAX_FUNC_LEN = 120;
     const int MAX_SYM_LEN = MAX_FILE_LEN + MAX_FUNC_LEN + 24;
     char **syms = malloc(2 * size * (sizeof(char*) + MAX_SYM_LEN));
     char *out = (char*)syms + size*sizeof(char*);
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -330,7 +330,38 @@ char* __symbolize(void *vaddr, char *buf, int size)
  * uses a stack frame or not, whether it uses a frame pointer, and where the return address is stored.
  * 
  * Since we do not have DWARF informations or similar metadata, we can just do educated guesses. A
- * mistake in the heuristic will result probably in a wrong backtrace from this point on. 
+ * mistake in the heuristic will result probably in a wrong backtrace from this point on.
+ * 
+ * The heuristic works as follows:
+ * 
+ *  * Most functions do have a stack frame. In fact, 99.99% of the functions you can find in a call stack
+ *    must have a stack frame, because the only functions without a stack frame are leaf functions (functions
+ *    that do not call other functions), which in turns can never be part of a stack trace.
+ *  * The heuristic walks the function code backwards, looking for the stack frame. Specifically, it looks
+ *    for an instruction saving the RA register to the stack (eg: `sd $ra, nn($sp)`), and an instruction
+ *    creating the stack frame (eg: `addiu $sp, $sp, -nn`). Once both are found, the heuristic knows how to
+ *    fill in `.stack_size` and `.ra_offset` fields of the function description structure, and it can stop.
+ *  * Some functions also modify $fp (the frame pointer register): sometimes, they just use it as one additional
+ *    free register, and other times they really use it as frame pointer. If the heuristic finds the
+ *    instruction `move $fp, $sp`, it knows that the function uses $fp as frame pointer, and will mark
+ *    the function as BT_FUNCTION_FRAMEPOINTER. In any case, the field `.fp_offset` will be filled in
+ *    with the offset in the stack where $fp is stored, so that the backtrace engine can track the
+ *    current value of the register in any case.
+ *  * The 0.01% of the functions that do not have a stack frame but appear in the call stack are leaf
+ *    functions interrupted by exceptions. Leaf functions pose two important problems: first, $ra is
+ *    not saved into the stack so there is no way to know where to go back. Second, there is no clear
+ *    indication where the function begins (as we normally stops analysis when we see the stack frame
+ *    creation). So in this case the heuristic would fail. We rely thus on two hints coming from the caller:
+ *    * First, we expect the caller to set from_exception=true, so that we know that we might potentially
+ *      deal with a leaf function. 
+ *    * Second, the caller should provide the function start address, so that we stop the analysis when
+ *      we reach it, and mark the function as BT_LEAF.
+ *    * If the function start address is not provided (because e.g. the symbol table was not found and
+ *      thus we have no information about function starts), the last ditch heuristic is to look for
+ *      the nops that are normally used to align the function start to the FUNCTION_ALIGNMENT boundary.
+ *      Obviously this is a very fragile heuristic (it will fail if the function required no nops to be
+ *      properly aligned), but it is the best we can do. Worst case, in this specific case of a leaf
+ *      function interrupted by the exception, the stack trace will be wrong from this point on.
  * 
  * @param func                        Output function description structure
  * @param ptr                         Pointer to the function code at the point where the backtrace starts.
@@ -339,12 +370,13 @@ char* __symbolize(void *vaddr, char *buf, int size)
  * @param func_start                  Start of the function being analyzed. This is optional: the heuristic can work
  *                                    without this hint, but it is useful in certain situations (eg: to better
  *                                    walk up after an exception).
- * @param exception_ra                If != NULL, this function was interrupted by an exception. This variable
- *                                    stores the $ra register value as saved in the exception frame, that might be useful.
+ * @param from_exception              If true, this function was interrupted by an exception. This is a hint that
+ *                                    the function *might* even be a leaf function without a stack frame, and that
+ *                                    we must use special heuristics for it.
  * 
  * @return true if the backtrace can continue, false if must be aborted (eg: we are within invalid memory)
  */
-bool __bt_analyze_func(bt_func_t *func, uint32_t *ptr, uint32_t func_start, void *exception_ra)
+bool __bt_analyze_func(bt_func_t *func, uint32_t *ptr, uint32_t func_start, bool from_exception)
 {
     *func = (bt_func_t){
         .type = (ptr >= inthandler && ptr < inthandler_end) ? BT_EXCEPTION : BT_FUNCTION,
@@ -390,20 +422,27 @@ bool __bt_analyze_func(bt_func_t *func, uint32_t *ptr, uint32_t func_start, void
         // We can stop looking and process the frame
         if (func->stack_size != 0 && func->ra_offset != 0)
             break;
-        if (exception_ra && addr == func_start) {
-            // The frame that was interrupted by an interrupt handler is a special case: the
-            // function could be a leaf function with no stack. If we were able to identify
-            // the function start (via the symbol table) and we reach it, it means that
-            // we are in a real leaf function.
-            func->type = BT_LEAF;
-            break;
-        } else if (exception_ra && !func_start && MIPS_OP_NOP(op) && (addr + 4) % FUNCTION_ALIGNMENT == 0) {
-            // If we are in the frame interrupted by an interrupt handler, and we does not know
-            // the start of the function (eg: no symbol table), then try to stop by looking for
-            // a NOP that pads between functions. Obviously the NOP we find can be either a false
-            // positive or a false negative, but we can't do any better without symbols.
-            func->type = BT_LEAF;
-            break;
+        if (from_exception) {
+            // The function we are analyzing was interrupted by an exception, so it might
+            // potentially be a leaf function (no stack frame). We need to make sure to stop
+            // at the beginning of the function and mark it as leaf function. Use
+            // func_start if specified, or try to guess using the nops used to align the function
+            // (crossing fingers that they're there).
+            if (addr == func_start) {
+                // The frame that was interrupted by an interrupt handler is a special case: the
+                // function could be a leaf function with no stack. If we were able to identify
+                // the function start (via the symbol table) and we reach it, it means that
+                // we are in a real leaf function.
+                func->type = BT_LEAF;
+                break;
+            } else if (!func_start && MIPS_OP_NOP(op) && (addr + 4) % FUNCTION_ALIGNMENT == 0) {
+                // If we are in the frame interrupted by an interrupt handler, and we does not know
+                // the start of the function (eg: no symbol table), then try to stop by looking for
+                // a NOP that pads between functions. Obviously the NOP we find can be either a false
+                // positive or a false negative, but we can't do any better without symbols.
+                func->type = BT_LEAF;
+                break;
+            }
         }
         addr -= 4;
     }
--- a/src/backtrace.c
+++ b/src/backtrace.c
@@ -99,8 +99,8 @@
  *   split externally to allow for efficiency reasons. Each entry stores the function name, 
  *   the source file name and line number, and the binary offset of the symbol within the containing
  *   function.
- * * String table: This tables can be thought as a large buffer holding all the strings needed by all
- *   symbol entries (function names and file names). Each symbol entry stores a string as an index
+ * * String table: this table can be thought as a large buffer holding all the strings needed by all
+ *   symbol entries (function names and file names). Each symbol entry stores a string as an offset
  *   within the symbol table and a length. This allows to reuse the same string (or prefix thereof)
  *   multiple times. Notice that strings are not null terminated in the string table.
  * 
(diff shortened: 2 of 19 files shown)


n64.mk
```diff
@@ -32,14 +32,15 @@ N64_CHKSUM = $(N64_BINDIR)/chksum64
 N64_ED64ROMCONFIG = $(N64_BINDIR)/ed64romconfig
 N64_MKDFS = $(N64_BINDIR)/mkdfs
 N64_TOOL = $(N64_BINDIR)/n64tool
+N64_SYM = $(N64_BINDIR)/n64sym
 N64_AUDIOCONV = $(N64_BINDIR)/audioconv64
 
 N64_CFLAGS =  -march=vr4300 -mtune=vr4300 -I$(N64_INCLUDEDIR)
-N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections
+N64_CFLAGS += -falign-functions=32 -ffunction-sections -fdata-sections -g
```
anacierdem: Is it ok to have this enabled by default? Would it be overkill to only emit debug information and symbol file only if `NDEBUG` is not set? I don't see any reason not to but not sure if someone may object. Though, we can always change this later so not important ATM...

src/backtrace_internal.h
```diff
@@ -0,0 +1,43 @@
+#ifndef __LIBDRAGON_BACKTRACE_INTERNAL_H
+#define __LIBDRAGON_BACKTRACE_INTERNAL_H
+
+/** @brief The "type" of funciton as categorized by the backtrace heuristic (__bt_analyze_func) */
+typedef enum {
+    BT_FUNCTION,                ///< Regular function with a stack frame
+    BT_FUNCTION_FRAMEPOINTER,   ///< The function uses the register fp as frame pointer (normally, this happens only when the function uses alloca)
+    BT_EXCEPTION,               ///< This is an exception handler (inthandler.S)
+    BT_LEAF                     ///< Leaf function (no calls), no stack frame allocated, sp/ra not modified
```
anacierdem, line 9: I suppose the tail calls will get into this category as well?

src/backtrace_internal.h
```diff
@@ -0,0 +1,43 @@
...
+} bt_func_type;
+
+/** @brief Description of a function for the purpose of backtracing (filled by __bt_analyze_func) */
+typedef struct {
+    bt_func_type type;       ///< Type of the function
+    int stack_size;          ///< Size of the stack frame
+    int ra_offset;           ///< Offset of the return address in the stack frame
```
anacierdem: I think it is the offset from the top of the frame. Might be nice to include that information as well.

src/backtrace_internal.h
```diff
@@ -0,0 +1,43 @@
...
+ * 
+ * If the symbol table is not found in the rompack or the address is not found,
+ * the return string is "???".
+ * 
+ * @param vaddr         Address to symbolize 
+ * @param buf           Buffer where to store the result
+ * @param size          Size of the buffer
+ * @return char*        Pointer to the return string. This is within the provided
+ *                      buffer, but not necessarily at the beginning because of DMA
+ *                      alignment constraints.
```
anacierdem, line 39: So the size must be at least some amount larger than the potential output, is there a rule of thumb here?

src/debug.c
```diff
@@ -70,6 +70,8 @@ static char sdfs_logic_drive[3] = { 0 };
 /** @brief debug writer functions (USB, SD, IS64) */
 static void (*debug_writer[3])(const uint8_t *buf, int size) = { 0 };
 
+/** @brief internal backtrace printing function */
+void __debug_backtrace(FILE *out, bool skip_exception);
```
anacierdem, line 79: Shouldn't this be static?

tools/common/polyfill.h
```diff
@@ -0,0 +1,74 @@
+#ifndef LIBDRAGON_TOOLS_POLYFILL_H
+#define LIBDRAGON_TOOLS_POLYFILL_H
+
+#ifdef __MINGW32__
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+size_t getline(char **lineptr, size_t *n, FILE *stream) {
```
anacierdem: Can we have a reference if we are using these from somewhere else?

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+    // Write the symbol table to file
+    verbose("Writing %s\n", outfn);
+    FILE *out = fopen(outfn, "wb");
+    if (!out) {
+        fprintf(stderr, "Cannot create file: symtable.bin\n");
+        exit(1);
+    }
+
+    fwrite("SYMT", 4, 1, out);
+    w32(out, 2); // Version
```
anacierdem, line 324: What happened to version 1 :)

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+            return 1;
+        }
+    }
+
+    if (i == argc) {
+        fprintf(stderr, "missing input filename\n");
+        return 1;
+    }
+
+    if (!n64_inst) {
```
anacierdem, line 407: Is the check necessary? Seems like it is only set once?

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+            uint32_t addr = strtoul(line, NULL, 16);
+            symbol_add(elf, addr, true);
+        }
+        // Find the callsites
+        if (strstr(line, "\tjal\t") || strstr(line, "\tjalr\t")) {
+            uint32_t addr = strtoul(line, NULL, 16);
+            symbol_add(elf, addr, false);
+        }
+    }
+    free(line);
```
anacierdem, line 235: If we are free ing line, it might be also nice to do so for cmd?

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+        if (!n64_inst) {
+            // Do not mention N64_GCCPREFIX in the error message, since it is
+            // a seldom used configuration.
+            fprintf(stderr, "Error: N64_INST environment variable not set.\n");
+            return 1;
+        }
+        // Remove the trailing backslash if any. On some system, running
+        // popen with a path containing double backslashes will fail, so
+        // we normalize it here.
+        n64_inst = strdup(n64_inst);
```
anacierdem: Might be nice to free things to be a good citizen.

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+    walign(out, 16);
+    w32_at(out, addrtable_off, ftell(out));
+    for (int i=0; i < stbds_arrlen(symtable); i++) {
+        struct symtable_s *sym = &symtable[i];
+        w32(out, sym->addr | (sym->is_func ? 0x1 : 0) | (sym->is_inline ? 0x2 : 0));
+    }
+
+    walign(out, 16);
+    w32_at(out, symtable_off, ftell(out));
+    for (int i=0; i < stbds_arrlen(symtable); i++) {
```
anacierdem, line 343: Can we have a reference for the layout of the file? I think at least pointing to the structs at backtrace.c would be nice

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+        fprintf(stderr, "Cannot create file: symtable.bin\n");
+        exit(1);
+    }
+
+    fwrite("SYMT", 4, 1, out);
+    w32(out, 2); // Version
+    int addrtable_off = w32_placeholder(out);
+    w32(out, stbds_arrlen(symtable));
+    int symtable_off = w32_placeholder(out);
+    w32(out, stbds_arrlen(symtable));
```
anacierdem, line 328: This is symtable length written the second time?
Ok, I later found out why in `backtrace.c` but still I'm not sure why

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+ * Backtrace 
+ * =========
+ * MIPS ABIs do not generally provide a way to walk the stack, as the frame
+ * pointer is not guaranteed to be present. It is possible to force its presence
+ * via "-fno-omit-frame-pointer", but we tried to provide a solution that works
+ * with standard compilation settings.
+ * 
+ * To perform backtracing, we scan the code backward starting from the return address
+ * of each frame. While scanning, we note some special instructions that we look
+ * for. The two main instructions that we look for are `sd ra, offset(sp)` which is
```
anacierdem, line 19: `sd ra, offset(sp)` does not exist in a leaf node right? So do we simply omit leaf stacks?
I think the later part:
> so that it's always possible to infer the function a certain address belongs to

also explains why we don't need them.

tests/backtrace.S
```diff
@@ -0,0 +1,147 @@
+#include "../src/regs.S"
+
+	.set noreorder
+    
+    .text
+
+    # This file contains functions used by test_backtrace.c to test
+    # the backtrace analysis code. The code of this functions is not
+    # run, but just scanned to extract the stack frame information.
```
anacierdem, line 9: Is this generated? If so, it would be great to include instructions/source code

tools/n64sym.c
```diff
@@ -0,0 +1,487 @@
...
+        if (strstr(line, "\tjal\t") || strstr(line, "\tjalr\t")) {
+            uint32_t addr = strtoul(line, NULL, 16);
+            symbol_add(elf, addr, false);
+        }
+    }
+    free(line);
+    pclose(disasm);
+}
+
+void compact_filenames(void)
```
anacierdem: This is currently unused

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+/** @brief Enable to debug why a backtrace is wrong */
+#define BACKTRACE_DEBUG 0
+
+/** @brief Function alignment enfored by the compiler (-falign-functions). 
+ * 
+ * @note This must be kept in sync with n64.mk.
+ */
+#define FUNCTION_ALIGNMENT      32
+
+#define MAX_FILE_LEN 120        ///< Maximum length of a file name in a backtrace entry
```
anacierdem: There is no limit on n64sym, why we have limits here?

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+ *   for each entry; all of them will have the same address, and all but the last one will have bit
+ *   1 set to 1.
+ * * Symbol table: this is a sequence of symbol table entries, each representing a symbol. The size
+ *   of this table (in number of entries) is exactly the same as the address table. In fact, each
+ *   address of the address table can be thought of as an external member of this structure; it's
+ *   split externally to allow for efficiency reasons. Each entry stores the function name, 
+ *   the source file name and line number, and the binary offset of the symbol within the containing
+ *   function.
+ * * String table: This tables can be thought as a large buffer holding all the strings needed by all
+ *   symbol entries (function names and file names). Each symbol entry stores a string as an index
```
anacierdem: >Each symbol entry stores a string...

is this a typo? I thing should be "an offset" instead of "a string" IINM

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+ *   within the symbol table and a length. This allows to reuse the same string (or prefix thereof)
+ *   multiple times. Notice that strings are not null terminated in the string table.
```
anacierdem, line 102: Do we also optimize for substrings?

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+ * 
+ * The SYMT file is generated by the n64sym tool during the build process.
+ */
+typedef struct alignas(8) {
+    char head[4];           ///< Magic ID "SYMT"
+    uint32_t version;       ///< Version of the symbol table
+    uint32_t addrtab_off;   ///< Offset of the address table in the file
+    uint32_t addrtab_size;  ///< Size of the address table in the file (number of entries)
+    uint32_t symtab_off;    ///< Offset of the symbol table in the file
+    uint32_t symtab_size;   ///< Size of the symbol table in the file (number of entries); always equal to addrtab_size.
```
anacierdem, line 112: Is this for "just in case"?

src/backtrace.c
```diff
@@ -0,0 +1,682 @@
...
+    data_cache_hit_writeback_invalidate(entry, sizeof(symtable_entry_t));
+    dma_read(entry, SYMT_ROM + symt->symtab_off + idx * sizeof(symtable_entry_t), sizeof(symtable_entry_t));
+}
+
+// Fetch the function name of an entry
+static char* symt_entry_func(symtable_header_t *symt, symtable_entry_t *entry, uint32_t addr, char *buf, int size)
+{
+    if (addr >= (uint32_t)inthandler && addr < (uint32_t)inthandler_end) {
+        // Special case exception handlers. This is just to show something slightly
+        // more readable instead of "notcart+0x0" or similar assembly symbols
```
anacierdem, line 290: hmm, maybe we can include the subsymbols as well?

anacierdem (commented):
This looks awesome! And I know it works on really weird C++ codebases :) Thank you!
Dropped a few comments. Didn't finish `backtrace.c` yet, but wanted to send this so that you can go over them in the meantime.

n64.mk
rasky: It can be done, but we can't tie it to NDEBUG (which is a C preprocessor symbol but not a Makefile symbol). We should introduce some Makefile-level variable to control this.

rasky: The important part I think is that the code is always considering symbols an optional part, and is ready to work without it. Obviously, symbolization can't work in that case, but you would still be able to get hexadecimal backtraces.

src/backtrace_internal.h
rasky: Tail-calling is a property of how you call a function, it's not a property of the function itself. Any function of the above types might tailcall another function of any of the above types.

src/backtrace_internal.h
rasky: Done.

src/backtrace_internal.h
anacierdem: Actually what I meant was a recursive tailcall. In that case it will just call itself without pushing anything to the stack. So shouldn't be any different than a leaf node IINM?

src/backtrace_internal.h
rasky: __symbolize will truncate the symbol to fit the buffer anyway, so any number would do. Consider that n64sym has a maximum size of symbols that can be specified from the command line (default 64), and I think you want to have space for a \0 plus some alignment at the beginning (I think 1 byte should suffice).

I think that if you have 64-ish characters available, you can simply pass a buffer of 128 chars to be on the safe side anyway. These numbers are small enough that it shouldn't be a problem.

src/debug.c
rasky: True, but it's going to be used by the new exception handler, in a followup.

n64.mk
anacierdem: Yeah we may introduce it in the future this is ok for now.

tools/common/polyfill.h
rasky: Done. I've also took the chance to get the improved version I found as this one was buggy.

tools/n64sym.c
rasky: The data layout was different and only allowed for 16bit offsets into string table, which is not enough for large C++ codebases. It's in the history of the unstable branch, but we're directly bringing version 2 to trunk.

tools/n64sym.c
rasky: No you're right, it's a leftover. Removed

tools/n64sym.c
rasky: Done.

tools/n64sym.c
rasky: in this specific case, I resist the urge as this is a single allocation in main, and the free would be done at the end of main so it's not really important, as the process is exiting anyway.

tools/n64sym.c
rasky: Add references in comments.

tools/n64sym.c
anacierdem: I agree the only thing concerning me was the probability of someone copying it in the future but doesn't seem likely either. Maybe the best opton is just commenting this intent?

tools/n64sym.c
anacierdem: As the one on the trunk is the actual release I think we can simply reset but that's ok if you want them to be compatible

tools/n64sym.c
rasky: So basically the address table contains the address for each symbol in the symbol table.

Looking at the code right now, there's no big win anymore in keeping the address separated from the symbol itself. It used to be more useful for alignment / lookups / dmas reasons but I don't see such a strong benefit anymore. We might evaluate merging the address table and the symbol table to simplify a bit readability of the code.

src/backtrace.c
rasky: well in general a leaf function can never be part of a stack trace by definition (as they don't call anything), with the only exception if an exception/interrupt happens in the middle of it. In that case, we can find a leaf function in the stack trace which is indeed a bit of a problem as it's not easy to understand where the function begins. So we handle that with some heuristics and through the symbol map (if present). See the code that handles BT_LEAF.

tests/backtrace.S
rasky: No it's just random copy and paste from disassembly of libdragon code. The code isn't even supposed to make sense, I have just copied and pasted together different snippets of assembly to recreate "normal looking" code for the several testcases.

tools/n64sym.c
rasky: Correct. I wanted to keep it around as I wanted to improve it sooner or later (as explained in the TODO), but I can remove it from trunk if you want.

anacierdem: If we are going to use it soon enough, we can leave. Let's remove o/w

src/backtrace.c
rasky: Right, this is not needed anymore, we can use VLAs to workaround this at least for the backtrace_symbols_cb case, so that it can handle any size.

For `backtrace_symbols()` (which is the standard API we implement for completeness but it's not the best one to use), since we want to avoid too many allocations, it's better to allocate once with a large enough size, so I moved the constants there.

src/backtrace.c
rasky: it was:
```Each symbol entry stores a string as an index within the symbol table```
I changed to:
```Each symbol entry stores a string as an offset within the symbol table```

Is this OK?

Comment (anacierdem) [reply src/backtrace.c]: Ah, yes I think this resolves the confusion for me

File: src/backtrace.c (hunk 18 above)
Comment (rasky) [reply src/backtrace.c:102]: The file format would allow that, true. 

The limit is on n64sym, because it becomes quadratic to search for a matching substring in an existing table. So in n64ysm we create a hash map with all prefixes (not all possible substrings), and then look that up. 

I think we could improve it adding to the hashmap some special substrings that we feel are more important than others. Assuming a C symbol like `symbol_add`, we could add to the hash map the word `add` (so any word separated by `_`, or even camel case) as it's more likely for a symbol `add` to exist compared to a symbol called `bol_a` (a random substring).

Anyway, I measured that, even using a full substring search, the gain in size is really minimal, as it rarely happens, so I didn't spend too much time on it.

File: src/backtrace.c (hunk 19 above)
Comment (rasky) [reply src/backtrace.c:112]: Yeah, I figured it out that keeping the header more generic wasn't going to be a problem.

File: src/backtrace.c (hunk 20 above)
Comment (rasky) [reply src/backtrace.c:290]: Honestly they are just local symbols, they're not really expected to be exported. It's more of a side effect of how gas works. If you check unstable, we already changed entrypoint.S to use the syntax `.Lskip_dma` (which is how local labels should be called), so they wouldn't appear anymore in the stacktraces.

I think even if we remove all labels from stacktraces, we would end up with a single symbol called `inthandler`; as much as we could rename it, I think it does provide more clear context to the user using a special symbol like `<EXCEPTION HANDLER>`.

File: src/backtrace.c (hunk 18 above)
Comment (anacierdem) [reply src/backtrace.c:102]: Yeah, I don't expect much difference. Just was curious b/c gcc seem to do it for sreing constants. In that case they all anchor from the trailing `\0` so probably it is a much smaller search space

File: src/backtrace.c (hunk 19 above)
Comment (anacierdem) [reply src/backtrace.c:112]: On second thought what about "reserving" a few bytes in the header as well? OTOH it is a matter of a rebuild but that extra length made me think about this as well

File: src/backtrace_internal.h (hunk 2 above)
Comment (rasky) [reply src/backtrace_internal.h:9]: Yes. In general a recursive tail call is handled by the compiler transforming the code in a loop, so at the assembly level you can't really tell it was written as recursive tail call. That's because a real tail call means popping the whole stack frame and then jumping, but in the recursive case it would be a waste of time to pop and push the same stack frame.

File: src/backtrace.c (hunk 19 above)
Comment (rasky) [reply src/backtrace.c:112]: It's actually not necessary because anything after the header is accessed via offsets specified in the header itself. That's actually by design.

Comment (anacierdem) [reply src/backtrace.c:112]: I mean in case we need to store anything else in the header in the future. I didn't get your point?

Comment (anacierdem) [reply src/backtrace.c:112]: Hmm I think I get it now, we will simply shift everything if we need any space.

File: tools/n64sym.c (hunk 12 above)
Comment (anacierdem) [reply tools/n64sym.c:328]: Is it because the symbol entries are constant length now?

File: src/backtrace.c [hunk 21]
```diff
@@ -0,0 +1,680 @@
...
+
+char* __symbolize(void *vaddr, char *buf, int size)
+{
+    symtable_header_t symt = symt_open();
+    if (symt.head[0]) {
+        uint32_t addr = (uint32_t)vaddr;
+        int idx = 0;
+        addrtable_entry_t a = symt_addrtab_search(&symt, addr, &idx);
+        while (!ADDRENTRY_IS_FUNC(a))
+            a = symt_addrtab_entry(&symt, --idx);
```
anacierdem, line 312: Maybe we can assert if idx goes down to 0, just in case.

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+ * mistake in the heuristic will result probably in a wrong backtrace from this point on. 
+ * 
+ * @param func                        Output function description structure
+ * @param ptr                         Pointer to the function code at the point where the backtrace starts.
+ *                                    This is normally the point where a JAL opcode is found, as we are walking
+ *                                    up the call stack.
+ * @param func_start                  Start of the function being analyzed. This is optional: the heuristic can work
+ *                                    without this hint, but it is useful in certain situations (eg: to better
+ *                                    walk up after an exception).
+ * @param exception_ra                If != NULL, this function was interrupted by an exception. This variable
```
anacierdem: I am not sure how the exception ra relates to parsing a random function? Shouldn't we be able to analyze the function irrespective of the current execution context?
Also see my later comment, instead of introducing this like a flag, we can just store the trace and merge them later.

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+    #if BACKTRACE_DEBUG
+    debugf("backtrace: start\n"); 
+    #endif
+
+    uint32_t* exception_ra = NULL;      // If != NULL, 
+    uint32_t func_start = 0;            // Start of the current function (when known)
+
+    // Start from the backtrace function itself. Put the start pointer somewhere after the initial
+    // prolog (eg: 64 instructions after start), so that we parse the prolog itself to find sp/fp/ra offsets.
+    ra = (uint32_t*)backtrace_foreach + 64;
```
anacierdem, line 485: Is there a way to use the actual length of this function somehow 🤔

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+            func.type == BT_FUNCTION ? "BT_FUNCTION" : (func.type == BT_EXCEPTION ? "BT_EXCEPTION" : (func.type == BT_FUNCTION_FRAMEPOINTER ? "BT_FRAMEPOINTER" : "BT_LEAF")),
+            ra, sp, fp, func.ra_offset, func.fp_offset, func.stack_size);
+        #endif
+
+        switch (func.type) {
+            case BT_FUNCTION_FRAMEPOINTER:
+                if (!func.fp_offset) {
+                    debugf("backtrace: framepointer used but not saved onto stack at %p\n", ra);
+                } else {
+                    // Use the frame pointer to refer to the current frame.
```
anacierdem, line 505: ...to the current stack?

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+            case BT_FUNCTION:
+                if (func.fp_offset)
+                    fp = *(uint32_t**)((uint32_t)sp + func.fp_offset);
+                ra = *(uint32_t**)((uint32_t)sp + func.ra_offset) - 2;
+                sp = (uint32_t*)((uint32_t)sp + func.stack_size);
+                exception_ra = NULL;
+                func_start = 0;
+                break;
+            case BT_EXCEPTION: {
+                // Exception frame. We must return back to EPC, but let's keep the
```
anacierdem, line 522: Would it be easier if we trace back from the last ra instead of EPC as if the exception has never occurred and then "append" the exception trace? Then we wouldn't need `exception_ra`. That would look something like:
- create stack trace A
- create stack trace B
- oops, this was an exception, temporarily store [A, B] and start walking back from EPC
- create stack trace C
- create stack trace D
- No more traces, append A, B to C, D

Current `exception_ra` confused me a lot. OTOH, we know this is working so that's ok if you don't think this would be useful.
Also please let me know if I am way of in my theory, you are the one who knows the code inside out.

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+
+                    // The function that jumped into an invalid PC was not interrupted by the exception: it
+                    // is a regular function
+                    // call now.
+                    exception_ra = NULL;
+                    break;
+                }
+
+                // The next frame might be a leaf function, for which we will not be able
+                // to find a stack frame. It is useful to try finding the function start.
```
anacierdem, line 564: I didn't exactly get why this is useful only when we are in an exception frame? Considering it is only being used in tandem with `exception_ra`, maybe we won't need it? This part really confused me. I really feel like the logic around the two should be simpler, but I don't know how exactly they relate.

src/backtrace.c
```diff
@@ -0,0 +1,680 @@
...
+                    while (!ADDRENTRY_IS_FUNC(entry))
+                        entry = symt_addrtab_entry(&symt, --idx);
+                    func_start = ADDRENTRY_ADDR(entry);
+                    #if BACKTRACE_DEBUG
+                    debugf("Found interrupted function start address: %08lx\n", func_start);
+                    #endif
+                }
+            }   break;
+            case BT_LEAF:
+                ra = exception_ra - 2;
```
anacierdem, line 580: This won't be necessary if we get rid of `exception_ra` as well

anacierdem (commented):
I have concerns around whether we need `exception_ra` or not. All looks good otherwise. Thank you again for the immense work you put into this, just to make people happy :)

src/backtrace.c
```diff
@@ -330,7 +330,38 @@ char* __symbolize(void *vaddr, char *buf, int size)
...
+ *  * Some functions also modify $fp (the frame pointer register): sometimes, they just use it as one additional
+ *    free register, and other times they really use it as frame pointer. If the heuristic finds the
+ *    instruction `move $fp, $sp`, it knows that the function uses $fp as frame pointer, and will mark
+ *    the function as BT_FUNCTION_FRAMEPOINTER. In any case, the field `.fp_offset` will be filled in
+ *    with the offset in the stack where $fp is stored, so that the backtrace engine can track the
+ *    current value of the register in any case.
+ *  * The 0.01% of the functions that do not have a stack frame but appear in the call stack are leaf
+ *    functions interrupted by exceptions. Leaf functions pose two important problems: first, $ra is
+ *    not saved into the stack so there is no way to know where to go back. Second, there is no clear
+ *    indication where the function begins (as we normally stops analysis when the see the stack frame
```
anacierdem: typo: `when the see`

src/backtrace.c
```diff
@@ -0,0 +1,720 @@
...
+ *      Obviously this is a very fragile heuristic (it will fail if the function required no nops to be
+ *      properly aligned), but it is the best we can do. Worst case, in this specific case of a leaf
+ *      function interrupted by the exception, the stack trace will be wrong from this point on.
+ * 
+ * @param func                        Output function description structure
+ * @param ptr                         Pointer to the function code at the point where the backtrace starts.
+ *                                    This is normally the point where a JAL opcode is found, as we are walking
+ *                                    up the call stack.
+ * @param func_start                  Start of the function being analyzed. This is optional: the heuristic can work
+ *                                    without this hint, but it is useful in certain situations (eg: to better
```
anacierdem, line 372: It is actually not used anywhere else for now, but considering it may be used in the future and it is not the same thing as `from_exception` I think it is ok to have it as a separate parameter. Just thinking out loud.

anacierdem (commented):
Looks much better now. I have resolved all points that I understand after the changes :)

tools/n64sym.c
rasky: I extracted the code into a `n64_toolchain_dir()` function in common. This allows to easily reuse it in other tools and make sure we don't duplicate the memory leak.

tools/n64sym.c
rasky: I don't remember exactly, the code underwent two or three refactorings. Anyway, it could be further improved now but I don't have a rush to do it, as it works quite well.

tools/n64sym.c
rasky: I have removed this and cherry-picked another follow up change from unstable to remove absolute paths via `-ffile-prefix-map`.

src/backtrace.c
rasky: Done

src/backtrace.c
rasky: Done

src/backtrace.c
rasky: Yeah I thought the same, for now I'd just leave it separate

src/backtrace.c
rasky: Normally, the stack pointer is used to access the frame. When there is a framepointer, though, we must use that one instead as the stack pointer has probably been moved further down. So I think the sentence is correct.

src/backtrace.c
rasky: I guess we could find a way, but at the end of the day we only need to walk through the prologue of the function, we don't need to go to the end. As long as the function is at least 64 opcode, this should work. Otherwise, it becomes more complicated and requires hooking up with the build system.

PR #361, opened 2023-03-24
timer: allow timers with very short intervals

### LINKED ISSUE #359: Timers are not called if the ticks value is "too small"
This was tested on hardware

Consider the simple rom:
```c
#include <libdragon.h>

// set to 1 when the timer callback runs
volatile int timer_i = 0;

void timer_callback(int ovfl)
{
    timer_i = 1;
}

int main(void)
{
    display_init(RESOLUTION_320x240, DEPTH_16_BPP, 2, GAMMA_NONE, ANTIALIAS_RESAMPLE);
    timer_init();

    // have timer_callback called every 1000 ticks
    new_timer(1000, TF_CONTINUOUS, timer_callback);

    int i = 0; // frame counter, mostly to make sure something is happening

    // increment i, and display i and timer_i on screen, every frame
    while (1)
    {
        i++;

        surface_t *surf;
        while ((surf = display_lock()) == NULL)
            ;

        graphics_fill_screen(surf, graphics_make_color(0, 0, 0, 0));

        char buf[100];
        snprintf(buf, sizeof(buf), "i=%d  timer_i=%d", i, timer_i);
        graphics_draw_text(surf, 10, 50, buf);

        display_show(surf);
    }
}
```
as expected this results in `timer_i=1` since `timer_callback` is run 1000 ticks or so after boot:
image

However if 1000 is changed to 1 (for example) (i.e. `new_timer(1, TF_CONTINUOUS, timer_callback);`) then it stays `timer_i=0`, indicating the timer callback never runs (note: `i` keeps incrementing)

image

Same thing with `TF_ONE_SHOT` instead of `TF_CONTINUOUS`.

Some limited testing shows 100 (`new_timer(100, TF_CONTINUOUS, timer_callback);`) doesn't run the callback but 200 does

rasky:
The timer module currently happens to ignore timers scheduled with very small interval. This is not a deliberate choice but rather an implementation issue: if the expiration timer is shorter than the time it takes to the timer library to setup the new timer (including intervening other interrupts that might happen meanwhile), the timer basically is scheduled on the next timer wraparound (~90 secs). This behavior is obviously buggy.

This commit changes the timer code to account for a timer that expires too soon by basically polling the timer queue immediately after the timer is scheduled.

Continuous timers with very short interval would now effectively cause an infinite loop. To avoid a hard freeze without notifying the user, we add an assert in case a timer expires more than 1000 times in a row within the same interrupt. Moreover, we also adjust a bit the code to allow to call stop_timer from within a timer callback, in case a timer wants to stop itself; this also allows us to write a test for this whole issue.

Fixes #359

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -182,12 +182,13 @@ static int __proc_timers(timer_link_t * thead)
 }
 
 /**
- * @brief Timer interrupt callback function
+ * @brief Poll the timer list and run callbacks for expired timers
  *
- * This function is called by the interrupt controller whenever 
- * compare == count.
+ * This function is called by the interrupt handler whenever 
+ * compare == count, and also when inserting into or removing
+ * from the timers list to improve handling timers with tiny delays
  */
-static void timer_interrupt_callback(void)
+static void timer_poll(void)
 {
 	uint32_t loop_count = 0;
 	while (__proc_timers(TI_timers)) {
@@ -248,7 +249,7 @@ void timer_init(void)
 	C0_WRITE_COUNT(1);
 	C0_WRITE_COMPARE(0);
 	set_TI_interrupt(1);
-	register_TI_handler(timer_interrupt_callback);
+	register_TI_handler(timer_poll);
 	enable_interrupts();
 }
 
@@ -287,7 +288,7 @@ timer_link_t *new_timer(int ticks, int flags, timer_callback1_t callback)
 			timer->next = TI_timers;
 			TI_timers = timer;
 			timer_update_compare(TI_timers, now);
-			timer_interrupt_callback();
+			timer_poll();
 		}
 
 		enable_interrupts();
@@ -331,7 +332,7 @@ timer_link_t *new_timer_context(int ticks, int flags, timer_callback2_t callback
 			timer->next = TI_timers;
 			TI_timers = timer;
 			timer_update_compare(TI_timers, now);
-			timer_interrupt_callback();
+			timer_poll();
 		}
 
 		enable_interrupts();
@@ -373,7 +374,7 @@ void start_timer(timer_link_t *timer, int ticks, int flags, timer_callback1_t ca
 			timer->next = TI_timers;
 			TI_timers = timer;
 			timer_update_compare(TI_timers, now);
-			timer_interrupt_callback();
+			timer_poll();
 		}
 
 		enable_interrupts();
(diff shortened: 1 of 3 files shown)


src/timer.c
```diff
@@ -92,6 +96,7 @@ static void timer_update_compare(timer_link_t *head) {
  */
 static int __proc_timers(timer_link_t * thead)
 {
+	// static int loop_count = 0;
```
Dragorn421:
```suggestion
```

src/timer.c
```diff
@@ -408,14 +433,16 @@ void restart_timer(timer_link_t *timer)
 {
 	if (timer)
 	{
+		disable_interrupts();
+
+		uint32_t now = TICKS_READ();
 		timer->left = TICKS_READ() + (int32_t)timer->set;
```
Dragorn421:
```suggestion
		timer->left = now + (int32_t)timer->set;
```
not that it matters I think but while you're at it

src/timer.c
```diff
@@ -177,11 +190,15 @@ static int __proc_timers(timer_link_t * thead)
  */
 static void timer_interrupt_callback(void)
```
Dragorn421: the comment on this function says
```
 *
 This function is called by the interrupt controller whenever 
 * compare == count.
```
I'd suggest either amending the comment, for example add
"also called when inserting into or removing from the timers list to improve handling timers with tiny delays" (if that's right?)
or moving the function body to a new "proc all timers and update next timer interrupt" function, and calling that instead of `timer_interrupt_callback`?

Dragorn421 (commented):
(in passing:) idk if being able to `stop_timer` in callbacks is so useful, compared to `delete_timer` (which can't be used in callbacks or it would cause libdragon to use-after-free), which one would want to naturally use if they used `new_timer` in the first place

src/timer.c
rasky: I've amended the comment and also renamed the function to `timer_poll` to avoid confusion.

rasky:
> (in passing:) idk if being able to `stop_timer` in callbacks is so useful, compared to `delete_timer` (which can't be used in callbacks or it would cause libdragon to use-after-free), which one would want to naturally use if they used `new_timer` in the first place

I know it's not extremely useful. We could make `delete_timer` work from callbacks as well, by detecting that it's being called from a timer callback (eg: via a new `TF_IN_CALLBACK` flag to be set and reset across the callback call) and in that case let `__proc_timers` doing the free (notifying it via eg. `TF_PENDING_DELETE`).

PR #362, opened 2023-03-25
Implement Overlays

gamemasterplc:
Implemented overlays in libdragon including 2 examples in examples/overlays. One example shows off three actors that can be spawned on the screen with the A button. The other shows two scenes that can be switched with the start button.


src/backtrace.c
```diff
@@ -169,13 +174,24 @@ static bool is_valid_address(uint32_t addr)
  * 
  * If not found, return a null header.
  */
-static symtable_header_t symt_open(void) {
-    if (SYMT_ROM == 0xFFFFFFFF) {
-        SYMT_ROM = rompak_search_ext(".sym");
-        if (!SYMT_ROM)
-            debugf("backtrace: no symbol table found in the rompak\n");
+static symtable_header_t symt_open(void *addr) {
+    dl_module_t *module = __dl_get_module(addr);
```
rasky: Is it possible to reverse the logic here? First have a check if the address falls within the main app text segment, and only call __dl_get_module if it doesn't?

rasky: Also if possible I'd break the dependency here with a trick similar to what we used for the demangling function. Have an internal global function pointer called __symt_check_dl_addr or something like that, and have the DL initializes it at the first dlopen. Otherwise we are now bringing part of the dynamic loader code into all applications just because of this reference.

PR #364, opened 2023-03-28
Add Dreamdrive64 Support

khill25:
This adds support for Dreamdrive64 rom cart. It's using a different address space from Picocart64, so I took a different abbreviation.


rasky:
Hi, thanks for the PR. First thing before review: is it possible to relicense the code using the Unlicense license that libdragon uses? I wouldn't want to bring in code that adds attribution constraints to our users like the BSD license does.

khill25:
> Hi, thanks for the PR. First thing before review: is it possible to relicense the code using the Unlicense license that libdragon uses? I wouldn't want to bring in code that adds attribution constraints to our users like the BSD license does.

No problem. I'm used to submitting PRs internally at my previous companies 😅

src/debug.c
```diff
@@ -226,7 +227,7 @@ DWORD get_fattime(void)
 static DSTATUS fat_disk_status_default(void) { return 0; }
 
 static DRESULT fat_disk_ioctl_default(BYTE cmd, void* buff)
-{
+{
```
Polprzewodnikowy: What's up with those random whitespaces added throughout the code?

src/debug.c
```diff
@@ -264,6 +265,16 @@ static fat_disk_t fat_disk_sc64 =
 	fat_disk_ioctl_default
 };
 
+static fat_disk_t fat_disk_ddr64 =
+{
+    fat_disk_initialize_ddr64,
+    fat_disk_status_default,
+    fat_disk_read_ddr64,
+	NULL,
```
Polprzewodnikowy: Inconsistent whitespace used, rest of the file use tab indentation and I suggest sticking with one option.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+// [WRITE] Register to define the cic type and save type.
+// 0xFF00 == Cic
+// 0x00FF == save
+#define DDR64_REGISTER_SELECTED_ROM_META (DDR64_REGISTER_SD_SELECT_ROM + 0x4)
+
+#endif
+
+static uint8_t ddr64_sd_wait() {
+    uint32_t timeout = 0;
+    uint32_t isBusy __attribute__((aligned(8)));
```
Polprzewodnikowy: You seem to have copied 64drive wait function from `usb.c`. Alignment on this variable is actually not needed as it is used only for return value from `io_read` function. USB lib has been rewritten recently and those alignment attributes were removed where possible.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+	isBusy = 1;
```
Polprzewodnikowy: Unnecessary assignment, value is overwritten by `io_read` function below.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+
+    // Wait until the cartridge interface is ready
+    do {
+		
+        // returns 1 while sd card is busy
+		isBusy = io_read(DDR64_CIBASE_ADDRESS_START + DDR64_REGISTER_SD_BUSY);
+        
+        // Took too long, abort
+        if((timeout++) > 10000000) {
+			fprintf(stdout, "SD_WAIT timed out. isBusy: %ld\n", isBusy);
```
Polprzewodnikowy: Do we really want to have error printing in function that can return error code?

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+			return -1;
+		}
+    }
+    while(isBusy != 0);
+    (void) timeout; // Needed to stop unused variable warning
```
Polprzewodnikowy, line 63: Unnecessary, `timeout` variable is used in the code above.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+static DRESULT fat_disk_read_ddr64(BYTE* buff, LBA_t sector, UINT count) {
+	_Static_assert(FF_MIN_SS == 512, "this function assumes sector size == 512");
+	_Static_assert(FF_MAX_SS == 512, "this function assumes sector size == 512");
+
+	int retryCount = 2;
+	int i = 0;
+	do {
+		uint64_t current_sector = sector + i;
+		
+		uint32_t part0 __attribute__((aligned(8)));
```
Polprzewodnikowy: Again, alignment unnecessary because you're just passing value to the `io_write` function that doesn't have any alignment requirements.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+		part0 = (current_sector & 0xFFFFFFFF00000000LL) >> 32;
```
Polprzewodnikowy: You can shift value first then mask it off like `(current_sector >> 32) & 0xFFFFFFFFUL`, I think it looks cleaner but it's personal preference.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+
+		uint32_t part1 __attribute__((aligned(8)));
```
Polprzewodnikowy: Same as above.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+		part1 = (current_sector & 0x00000000FFFFFFFFLL);
```
Polprzewodnikowy: You can ommit upper zeros, I think it looks cleaner but it's personal preference.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+
+        // wait for the sd card to finish
+        if(ddr64_sd_wait() == 0) {
+			data_cache_hit_writeback_invalidate(buff, 512);
+			dma_read(buff, DDR64_BASE_ADDRESS_START, 512);
+			buff += 512;
+			i++;
+		} else {
+			retryCount--;
+			fprintf(stderr, "wait timeout\n");
```
Polprzewodnikowy: Again, why print debug info here.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+		}
+
+		if (retryCount <= 0) {
+			break;
```
Polprzewodnikowy: No error code returned when read operation failed?

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+		}
+		
+	} while (i < count);
+
+	return RES_OK;
+}
+
+static DRESULT fat_disk_write_ddr64(const BYTE* buff, LBA_t sector, UINT count)
+{
+	fprintf(stderr, "Picocart64 does not currently support SD card writes.\n");
```
Polprzewodnikowy: Change this to `assertf` instead of printing to the stderr.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+    return RES_OK;
+}
+
+static DSTATUS fat_disk_initialize_ddr64(void) { return RES_OK; }
```
Polprzewodnikowy: I would move this function before read/write functions as it's logically first to be called before doing actual R/W operations.

src/debug_sdfs_ddr64.c
```diff
@@ -0,0 +1,136 @@
...
+#define DDR64_CIBASE_ADDRESS_START   (DDR64_BASE_ADDRESS_END + 1)
+#define DDR64_CIBASE_ADDRESS_LENGTH  (0x00000800)
+#define DDR64_CIBASE_ADDRESS_END     (DDR64_CIBASE_ADDRESS_START + DDR64_CIBASE_ADDRESS_LENGTH - 1)
+
+// [Read]: Returns DDR64_MAGIC
+#define DDR64_REGISTER_MAGIC         (0x00000000)
+#define DDR64_MAGIC                  (0xD9D96400)
+
+// [WRITE]: Write number of bytes to print from TX buffer
+#define DDR64_REGISTER_UART_TX       (0x00000004)
```
Polprzewodnikowy: Don't really understand why unused defines are added here.

khill25:
Wow thanks for the though review! It's been a while and I let my code standards lapse 😅 I'll set about getting this fixed up. Not sure what is up with the whitespace, I'll run a formatter and see if I can sort that out.

src/debug.c
anacierdem: I think we need to use a tool for this at some point, this happens a lot unfortunately. Clang format does not seem to play well with some gcc extensions though

khill25: Whitespace is driving me nuts! It looks totally fine in visual studio code and in my git client, I don't notice it's messed up until I push to the PR.

anacierdem: You can choose to display them if you haven't already: https://stackoverflow.com/questions/30140595/show-whitespace-characters-in-visual-studio-code

khill25: So appears that the rest of the codebase uses spaces and I'm using tabs. I'll have to convert all my tabs to spaces

khill25:
Weep for the tab to space changes.

Anyway I think I addressed all the feedback. Again, I'm really sorry about the whitespace, I know it's annoying to review code that's basically all whitespace changes.

rasky:
Hi @khill25, let me update you on this PR.

Your code was sent as two major changes were being worked on. One (just landed) is #366 that is a big improvement to usb.c (coming from upstream). Another one, waiting for last touches, is #371 that basically removes the current SD code in favor of vendoring [libcart](https://github.com/devwizard64/libcart), which is a newly released library which is more comprehensive that our builtin SD support. Our goal is not having to maintain the SD code anymore in here, and offload that to libcart.

Your PR touches the same areas at the same time, so it is unfortunately due to some rework once also #371 lands. My suggestion would be to wait until #371 lands and then I can point you to the better way of reworking your code with the new overall architecture. That will probably be adding support for Dreamdrive64 to libcart rather than libdragon proper, but let me see where we end with #371 first.

Thanks for the patience!

rasky:
Hi @khill25, the libcart PR (#371) was merged. This means that SD flashcart code is now maintained at upstream [libcart](https://github.com/devwizard64/libcart) which is where you should send your PR now. I will close this PR now as we won't be able to merge it as-is, as we don't carry SD flashcard code anymore.

PR #365, opened 2023-04-05
Allow RDPQ material and texturing commands integration inside GL's context

SpookyIluha:
These changes allow libdragon's OpenGL to fully utilize the N64's custom RDP material and texturing capabilities via RDPQ.

There are now 2 new GL enums that can be enabled or disabled with glEnable/glDisable: GL_RDPQ_MATERIAL_N64 and GL_RDPQ_TEXTURING_N64.

The GL_RDPQ_MATERIAL_N64 enum allows to setup a custom Color Combiner and/or Color Blender via rdpq_mode_combiner/rdpq_mode_blender and also allows to setup the RDP's internal color registers such as PRIM/ENV/BLEND etc.

The GL_RDPQ_TEXTURING_N64 enum allows to load custom textures and surfaces (for example with rdpq_tex_load/sub), allows to configure the tile and tilesize parameters of each tile, allows to setup custom mipmapping and to override rdpq_mode_filter filtering.


src/GL/rsp_gl.S
```diff
@@ -453,8 +454,19 @@ GLCmd_PreInitPipe:
...
+    lw state_flags2, %lo(GL_STATE_FLAGS2)
+    # li t2, %lo(GL_STATE_FLAGS2)
+
+    #define modes0          t8
+    #define modes1          t9
+
+    # Load the current OTHERMODE state from RDPQ
+
+    lw modes0, %lo(RDPQ_OTHER_MODES) + 0x0
+    lw modes1, %lo(RDPQ_OTHER_MODES) + 0x4
```
snacchus: Loading `modes0` and `modes1` here does not do anything since they are overwritten later on anyway. I would completely drop that part and just do this:
```suggestion
    #define state_flags  k1
    #define state_flags2 k0

    lw state_flags, %lo(GL_STATE_FLAGS)
    lw state_flags2, %lo(GL_STATE_FLAGS2)
```

src/GL/rsp_gl.S
```diff
@@ -518,6 +530,11 @@ GLCmd_PreInitPipe:
     #define active_tex      s7
 
     # Get Active texture
+    # * Skip if FLAG2_USE_RDPQ_TEXTURING is active
+
+    andi t1, state_flags2, FLAG2_USE_RDPQ_TEXTURING
+    bgtz t1, rdpq_skiptexturing
+    nop
```
snacchus: The `active_tex` register is actually needed in some cases even if `GL_RDPQ_TEXTURING_N64` is enabled. More precisely, it's needed to compute the combiner state further down (given that `GL_RDPQ_MATERIAL_N64` is _not_ enabled at the same time)
For this reason I would not skip over the code that sets `active_tex` and just drop this altogether (See my next comment for an alternative solution):
```suggestion
```

src/GL/rsp_gl.S
```diff
@@ -648,6 +672,43 @@ GLCmd_PreInitPipe:
...
+    sw t1, %lo(RDPQ_COMBINER) + 0x4
+    sw t2, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x0
+    sw t3, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x4
+
+rdpq_skipcombinerblender:
+
+    # * Skip if FLAG2_USE_RDPQ_TEXTURING is active
+
+    andi t1, state_flags2, FLAG2_USE_RDPQ_TEXTURING
+    bgtz t1, rdpq_mergemask
```
snacchus: Like above, I'd prefer to use `bnez`. Also adjusting the label name.
```suggestion
    bnez t1, gl_mergemask
```

src/GL/rsp_gl.S
```diff
@@ -631,6 +650,11 @@ GLCmd_PreInitPipe:
 1:
     or modes1, t3
 
+    # * Skip if FLAG2_USE_RDPQ_MATERIAL is active
+    andi t1, state_flags2, FLAG2_USE_RDPQ_MATERIAL
+    bgtz t1, rdpq_skipcombinerblender
+    nop
```
snacchus: To stay consistent with the rest of the code I'd use `bnez` instead of `bgtz`, which is guaranteed to ignore the sign bit (just to be safe). Also, the `nop` is not necessary here and I adjusted the label name according to my suggestion below.
```suggestion
    bnez t1, gl_skipcombinerblender
```

src/GL/rsp_gl.S
```diff
@@ -531,6 +548,8 @@ GLCmd_PreInitPipe:
     jal GL_UpdateTextureUpload
     nop
 
+rdpq_skiptexturing:
```
snacchus: (Continuing from above comment)
Instead, I would skip calling `GL_UpdateTextureUpload` depending on whether the texturing flag is set, like this:
```suggestion
    # * Skip if FLAG2_USE_RDPQ_TEXTURING is active

    andi t1, state_flags2, FLAG2_USE_RDPQ_TEXTURING
    beqz t1, GL_UpdateTextureUpload
    li ra, %lo(1f)
1:
```

src/GL/rsp_gl.S
```diff
@@ -648,6 +672,43 @@ GLCmd_PreInitPipe:
...
+    lw t3, %lo(COMBINER_MIPMAPMASK_TABLE) + 0x4(t5)
+
+    # TODO: The following is sort of equivalent to RDPQCmd_ResetMode. Maybe make that callable from ucode?
+
+    sw t0, %lo(RDPQ_COMBINER) + 0x0
+    sw t1, %lo(RDPQ_COMBINER) + 0x4
+    sw t2, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x0
+    sw t3, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x4
+
+rdpq_skipcombinerblender:
```
snacchus: I don't think we really have a good naming convention for labels so far, but for now I think we should stay consistent with the rest of the code and use the `gl_` prefix.
```suggestion
gl_skipcombinerblender:
```

src/GL/rsp_gl.S
```diff
@@ -685,34 +746,17 @@ GLCmd_PreInitPipe:
...
-    move t5, is_points
-    lw t3, %lo(GL_STATE_TEX_ENV_MODE)
-
-    li t4, GL_REPLACE
-    bne t3, t4, 1f
+    j rdpq_skiptexture
     nop
-    ori t5, 0x2
-1:
+rdpq_mergemask:
```
snacchus: Same as above, let's use the `gl_` prefix.
```suggestion
gl_mergemask:
```

src/GL/rsp_gl.S
```diff
@@ -685,34 +746,17 @@ GLCmd_PreInitPipe:
...
+    lw t1, %lo(RDPQ_OTHER_MODES) + 0x0
+    and t1, RDPQ_TEXTURING_MASK
+    and t2, modes0, ~(RDPQ_TEXTURING_MASK)
+    or  modes0, t1, t2
 
-    sw t0, %lo(RDPQ_COMBINER) + 0x0
-    sw t1, %lo(RDPQ_COMBINER) + 0x4
-    sw t2, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x0
-    sw t3, %lo(RDPQ_COMBINER_MIPMAPMASK) + 0x4
+rdpq_skiptexture:
```
snacchus: Same as above, this label should also use the `gl_` prefix. Additionally, I think the name is somewhat misleading altogether. Maybe it should be `gl_save_modes` or something?
```suggestion
gl_save_modes:
```

src/GL/rsp_gl.S
```diff
@@ -685,34 +746,17 @@ GLCmd_PreInitPipe:
     or modes0, t7
 2:
 
-    # Combiner
-    move t5, is_points
-    lw t3, %lo(GL_STATE_TEX_ENV_MODE)
-
-    li t4, GL_REPLACE
-    bne t3, t4, 1f
+    j rdpq_skiptexture
```
snacchus: Adjusting the label name to match my suggestion below
```suggestion
    j gl_save_modes
```

snacchus (changes requested):
I've re-reviewed rsp_gl.S and spotted a couple more issues. Some are bugs, some are more stylistic in nature. I've added change suggestions to all my comments so if you agree with them, you can just apply them as-is. I did test them myself first, but just to be safe I'd appreciate if you could test them as well.

Once these are resolved the PR is good to be merged in my view. Technically, to make this a "proper" extension, there are a couple of things missing, but I can take care of that separately as they don't affect normal use cases in any way.

PR #366, opened 2023-04-09
Updated USB library to match latest

buu342:
See title.

The main changes are as follows:
* The code regarding PI read/writes and DMA transfers has been simplified to make it more legible than tons of `#ifdef`'s everywhere.
* GCC's aligned macro has been replaced with a different approach to add support for people using IDO. (https://github.com/buu342/N64-UNFLoader/issues/86)
* All flashcart code has been fixed to be async. USB can be unplugged and the game will continue as normal, being able to perform USB IO once the USB is plugged back in (https://github.com/buu342/N64-UNFLoader/issues/89) (https://github.com/buu342/N64-UNFLoader/issues/85)
* Added emulation detection. It is off by default, users will need to explicitly enable the macro and recompile libdragon to enable it. (https://github.com/buu342/N64-UNFLoader/issues/60)
* Other miscellaneous cleanups/tweaks to the code. Nothing worthy pointing out.

Flashcart Specific Changes:
* 64Drive polling has been fixed so that it no longer misses data, eliminating the need for the POLLTIME macro (https://github.com/buu342/N64-UNFLoader/issues/51)
* EverDrive code was changed to remove all unnecessary leftovers and oddities from KRIKzz's original USB code.
* EverDrive code has been changed to read/write aligned to 2 bytes. It was originally believed (due to the USB source + lack of documentation) that it had 16 byte requirements.
* Fixed another cock-up wrt EverDrive alignment (https://github.com/buu342/N64-UNFLoader/issues/98)
* SC64 code has been updated to match the current version.

### DIFF
--- a/include/usb.h
+++ b/include/usb.h
@@ -24,6 +24,7 @@
     #define DATATYPE_RAWBINARY  0x02
     #define DATATYPE_HEADER     0x03
     #define DATATYPE_SCREENSHOT 0x04
+    #define DATATYPE_HEARTBEAT  0x05
     
     extern int usb_datatype;
     extern int usb_datasize;
@@ -118,5 +119,17 @@
     ==============================*/
     
     extern void usb_purge();
+
+
+    /*==============================
+        usb_sendheartbeat
+        Sends a heartbeat packet to the PC
+        This is done once automatically at initialization,
+        but can be called manually to ensure that the
+        host side tool is aware of the current USB protocol
+        version.
+    ==============================*/
+
+    extern void usb_sendheartbeat();
 
 #endif
--- a/include/usb.h
+++ b/include/usb.h
@@ -6,12 +6,13 @@
     *********************************/
     
     // UNCOMMENT THE #DEFINE IF USING LIBDRAGON
-    #define LIBDRAGON                    
-
+    #define LIBDRAGON
+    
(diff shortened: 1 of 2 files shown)


rasky:
@buu342 thanks! Can you please confirm the backward compatibility status of the USB library after this change? Is it 100% compatible with the previous version? I'm speaking of compatibility on both the API side (usb.h) and protocol side for host programs.

anacierdem:
Thank you! The long waited update 😊 I will test it on an everdrive v3 as well.

buu342:
usb.h is pretty much unchanged, however host-side tools will need to be updated due to the EverDrive changes wrt byte alignment. Backwards compatibility can be focused on from here on out, as this is pretty much a V2 update to the library. It just wasn't a priority when the library was originally written (I just expected folk to stay updated).

buu342:
I have added in the heartbeat function as discussed on Discord :slightly_smiling_face:

rasky:
Given that this breaks compatibility for Everdrives and requires a new (unreleased) UNFLoader, we'll keep this PR on hold until the new version of UNFLoader is out and stabilized, so that we can later advise our users to upgrade when necessary.

buu342:
Updated one again with the timeout check function

PR #371, opened 2023-05-01
Add libcart

devwizard64:


src/cart.c
```diff
@@ -0,0 +1,1814 @@
...
+    if (io_read(SC_VERSION_REG) != SC_VERSION) CART_ABORT();
+    __sc_sync();
+    /* SC64 maps 64DD IPL to SDRAM */
+    io_write(SC_DATA0_REG, SC_CFG_DD_MODE);
+    io_write(SC_COMMAND_REG, SC_CONFIG_GET);
+    __sc_sync();
+    /* Have IPL */
+    if (io_read(SC_DATA1_REG) & SC_DD_MODE_IPL)
+    {
+        cart_size = 0x3BC0000; /* 59.75 MiB */
```
Polprzewodnikowy: It's not documented anywhere but if 64DD regs are enabled flashcart firmware can use anything above 32 MiB in SDRAM memory for buffers and SD card sector tables. I would add another check if 64DD regs are used and limit usable memory size to 32 MiB.

src/cart.c
```diff
@@ -0,0 +1,1814 @@
...
+}
+
+#define SC_BASE_REG             0x1FFF0000
+#define SC_BUFFER_REG           0x1FFE0000
+
+#define SC_STATUS_REG           (SC_BASE_REG+0x00)
+#define SC_COMMAND_REG          (SC_BASE_REG+0x00)
+#define SC_DATA0_REG            (SC_BASE_REG+0x04)
+#define SC_DATA1_REG            (SC_BASE_REG+0x08)
+#define SC_VERSION_REG          (SC_BASE_REG+0x0C)
```
Polprzewodnikowy: This register and corresponding magic value was renamed to `IDENTIFIER` recently as it didn't carry any version information. Firmware version can be checked via `V` command. DATA0 register contains major and minor firmware version in format `[31:16]: major | [15:0]: minor` and DATA1 contains firmware revision as 32 bit number.
Major value increments when breaking changes are introduced.
Minor value increments when new commands or registers are added in non breaking fashion.
Revision value increments when no API changes were made (e.g. bug fixes)

PR #374, opened 2023-05-13
Add libcart

devwizard64:


rasky (approved):
Fine by me, waiting for @buu342 to comment the usb.c changes as I'd rather those to go upstream as well.

buu342:
Looks good to me. Always worth asking: have the changes been tested in both extended address and not?

rasky:
Merged manually into trunk, thanks!

PR #377, opened 2023-05-19
Add autodetect fmt from name in mksprite

SpookyIluha:


rasky:
Thanks, reworked as https://github.com/DragonMinded/libdragon/commit/8a48bd55fac97bc909e771ecdaf75e8cbd8adbe4

PR #380, opened 2023-06-14
Add detail texture support to rdpq_sprite_upload

SpookyIluha:
Basically the title.

The detail texture can now be created using mksprite and uploaded alongside the main texture and/or its mipmaps.
This change also introduces a few new modes and functions
- new mipmap modes to accomodate detail texture or sharpen (accessible without much API changes through rdpq_mode_mipmap and rdpq_mipmap_t)
- added support to changing min lod / prim lod / prim color interchangeably or all together and whilst saving the prim state
- added support to reuse textures when doing a multi-texture upload through rdpq_tex_reuse/_sub ~~(not yet ready in PR, hopefully fixed soon)~~
- additional sprite functions to retreive detail texture data


include/rdpq.h
```diff
@@ -952,19 +953,108 @@ inline void rdpq_set_blend_color(color_t color)
...
     extern void __rdpq_write8(uint32_t cmd_id, uint32_t arg0, uint32_t arg1);
-    __rdpq_write8(RDPQ_CMD_SET_PRIM_COLOR, 0, color_to_packed32(color));
+    __rdpq_write8(RDPQ_CMD_SET_PRIM_COLOR_COMPONENT, (0<<16), color_to_packed32(color));
+}
+
+/**
+ * @brief Set the RDP MIN LOD combiner register (RDP command: SET_PRIM_COLOR (partial))
+ * 
+ * This function sets the internal RDP PRIM MIN LOD register, that is used for 
+ * determining the interpolation blend factor of a detail texture.
```
rasky: Should we maybe rename it in another way? I don't know the story behind "MIN LOD", but if we use it as blend factor for detail texture, maybe it should be `rdpq_set_blend_detail_factor()` or something like that.

include/sprite.h
```diff
@@ -170,6 +186,42 @@ uint16_t* sprite_get_palette(sprite_t *sprite);
...
+/**
+ * @brief Get the factor of a detail texture in sprite. Range 0..1
+ * 
+ * This function returns the blend factor used in detail texture min lod. 0 means fully
+ * invisible, while 1 means fully visible.
+ * 
+ * @param sprite        The sprite to access
+ * @return              Blend factor if sprite has detail texture, 0 otherwise
+ */
+float sprite_detail_get_factor(sprite_t *sprite);
```
rasky: this is a bit wide as an API surface for a single feature. My suggestion is to add a sprite_detail_t structure, and then use a single API like this:

```C
surface_t sprite_get_detail_pixels(sprite_t *sprite, sprite_detail_t *info);
```

so that you return the surface and the info (maybe the info can even be optional, if NULL).

rasky (commented):
first round, still need to check the texloader stuff

include/sprite.h
```diff
@@ -59,6 +59,30 @@ typedef struct sprite_s
...
+ * to increase the perceived resolution of the main sprite when rendering
+ * with little to no additional TMEM usage.
+ * 
+ * If the sprite uses a detail texture, its information can be retreived
+ * using the #sprite_get_detail_pixels function.
+ * 
+ * To include a detail texture to libdragon's sprite format, use
+ * the mksprite tool with --detail argument. 
+ * 
+ * #rdpq_sprite_upload automatically uploads detail textures accosiated with
```
rasky:
```suggestion
 * #rdpq_sprite_upload automatically uploads detail textures associated with
```

src/rdpq/rdpq_sprite.c
```diff
@@ -39,12 +39,38 @@ int rdpq_sprite_upload(rdpq_tile_t tile, sprite_t *sprite, const rdpq_texparms_t
 
     // If no texparms were provided but the sprite contains some, use them
     rdpq_texparms_t parms_builtin;
+    rdpq_texparms_t detailtexparms;
     if (!parms && sprite_get_texparms(sprite, &parms_builtin))
         parms = &parms_builtin;
 
+    // Check for detail texture
+    sprite_detail_t detailinfo;
+    sprite_get_detail_texparms(sprite, &detailtexparms);
```
rasky: I'd move this call in the `if(use_detail)` below. Moreover, you can use the boolean return value: if the value is false, you can call `rdpq_tex_upload` passing NULL instead of `&detailtexparms`. Passing NULL to `rdpq_tex_upload` is always faster as it skips the whole tile parms calculation.

src/rdpq/rdpq_tex.c
```diff
@@ -15,9 +15,14 @@
...
 /** @brief Non-zero if we are doing a multi-texture upload */
-static int multi_upload = 0;
-static int multi_upload_bytes = 0;
-static int multi_upload_limit = 0;
+typedef struct rdpq_multi_upload_s {
+    int  used;
+    int  bytes;
+    int  limit;
+
+    tex_loader_t last_tload;
```
rasky: since we're using this also outside of the multi sequence, maybe we should also move this outside of rdpq_multi_upload_t, for clarity. I'm not super excited about always using a global instance of texloader rather than a stack instance like before, but I guess we can live with this for a while.

src/rdpq/rdpq_tex.c
```diff
@@ -293,6 +298,34 @@ static void texload_tile(tex_loader_t *tload, int s0, int t0, int s1, int t1)
...
+    t1 = t1*4 + tload->rect.t1fx;
+    rdpq_set_tile_size_fx(tload->tile, s0, t0, s1, t1);
+}
+
+static void texload_settile_offset(tex_loader_t *tload, int s0, int t0, int s1, int t1, int tmem_offset)
+{
+    assertf(tmem_offset % 8 == 0, "Offset %i must be in multiples of 8", tmem_offset);
+    tex_format_t fmt = surface_get_format(tload->tex);
+
+    rdpq_set_tile(tload->tile, fmt, tmem_offset, tload->rect.tmem_pitch, &(tload->tileparms));
```
rasky: shouldn't this be `tload->tmem_addr + tmem_offset`? Also how does this work if `tmem_addr` is already precomputed as part of the autotmem allocator?

src/rdpq/rdpq_tex.c
```diff
@@ -619,17 +698,18 @@ void rdpq_tex_multi_begin(void)
 {
     // Initialize autotmem engine
     rdpq_set_tile_autotmem(0);
-    if (multi_upload++ == 0) {
-        multi_upload = true;
-        multi_upload_bytes = 0;
-        multi_upload_limit = 4096;
+    if (multi_upload.used++ == 0) {
+        multi_upload.used = true;
```
rasky: I think this line is spurious? `used` is a counter

PR #382, opened 2023-06-27
libdragon build Makefile fixes

Dragorn421:
# Remove installed headers from the include path when building libdragon

```c
    [CC] src/debug.c
In file included from src/debug.c:21:
/home/dragorn421/Documents/libragon/install/mips64-elf/include/usb.h:80:12: error: unknown type name 'u32'
   80 |     extern u32 usb_poll(void);
      |            ^~~
make: *** [n64.mk:151: build/debug.o] Error 1
```
I'm not inquiring about the error, I caused it and can fix it and understand why it's there,
but
why is it including `/home/dragorn421/Documents/libragon/install/mips64-elf/include/usb.h` instead of `/home/dragorn421/Documents/libragon/include/usb.h`=`./include/usb.h`

this is fixed e.g. by
- moving the "local" include path in front of the other args which include the include path under `$(N64_INST)`
```diff
$ git diff Makefile
diff --git a/Makefile b/Makefile
index 0acfd4b8..04dbb1c1 100755
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ libdragon: CC=$(N64_CC)
 libdragon: CXX=$(N64_CXX)
 libdragon: AS=$(N64_AS)
 libdragon: LD=$(N64_LD)
-libdragon: CFLAGS+=$(N64_CFLAGS) $(LIBDRAGON_CFLAGS)
+libdragon: CFLAGS+=$(LIBDRAGON_CFLAGS) $(N64_CFLAGS)
 libdragon: CXXFLAGS+=$(N64_CXXFLAGS) $(LIBDRAGON_CFLAGS)
 libdragon: ASFLAGS+=$(N64_ASFLAGS) $(LIBDRAGON_CFLAGS)
 libdragon: RSPASFLAGS+=$(N64_RSPASFLAGS) $(LIBDRAGON_CFLAGS)
```
- or no-op-ifying the `-I$(N64_INCLUDEDIR)` in `N64_CFLAGS`:
```diff
$ git diff Makefile
diff --git a/Makefile b/Makefile
index 0acfd4b8..f17e4ef4 100755
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ BUILD_DIR = build
 include n64.mk
 INSTALLDIR = $(N64_INST)
 
+N64_INCLUDEDIR = _somenamethatdoesntexist_
 LIBDRAGON_CFLAGS = -I$(CURDIR)/src -I$(CURDIR)/include -ffile-prefix-map=$(CURDIR)=libdragon
 
 # Activate N64 toolchain for libdragon build
```

TLDR should libdragon have `$(N64_INST)/mips64-elf/include` on its include path (it currently does and it's earlier than the rest of the include path)

Rasky said probably not and I agree

I went with another way to fix it in the PR, have N64_INCLUDEDIR point to the source includes (whereas typically it points to the installed includes)

# Add mkdir-ing install folders

This is an issue when N64_INST != N64_GCCPREFIX and installing libdragon to a N64_INST where gcc/binutils hasn't been installed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,9 +78,11 @@ install-mk: $(INSTALLDIR)/include/n64.mk
 $(INSTALLDIR)/include/n64.mk: n64.mk
 # Always update timestamp of n64.mk. This make sure that further targets
 # depending on install-mk won't always try to re-install it.
+	mkdir -p $(INSTALLDIR)/include
 	install -cv -m 0644 n64.mk $(INSTALLDIR)/include/n64.mk
 
 install: install-mk libdragon
+	mkdir -p $(INSTALLDIR)/mips64-elf/lib
 	install -Cv -m 0644 libdragon.a $(INSTALLDIR)/mips64-elf/lib/libdragon.a
 	install -Cv -m 0644 n64.ld $(INSTALLDIR)/mips64-elf/lib/n64.ld
 	install -Cv -m 0644 rsp.ld $(INSTALLDIR)/mips64-elf/lib/rsp.ld
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,11 @@ BUILD_DIR = build
 include n64.mk
 INSTALLDIR = $(N64_INST)
 
-LIBDRAGON_CFLAGS = -I$(CURDIR)/src -I$(CURDIR)/include -ffile-prefix-map=$(CURDIR)=libdragon
+# Don't use the installed include files (e.g. /opt/libdragon/mips64-elf/include) for building libdragon,
+# use the source ones instead (./include)
+N64_INCLUDEDIR = $(CURDIR)/include
+
+LIBDRAGON_CFLAGS = -I$(CURDIR)/src -ffile-prefix-map=$(CURDIR)=libdragon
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,7 +6,11 @@ BUILD_DIR = build
 include n64.mk
 INSTALLDIR = $(N64_INST)
 
-LIBDRAGON_CFLAGS = -I$(CURDIR)/src -I$(CURDIR)/include -ffile-prefix-map=$(CURDIR)=libdragon
+# Don't use the installed include files (e.g. /opt/libdragon/mips64-elf/include) for building libdragon,
+# use the source ones instead (./include)
+N64_INCLUDEDIR = $(CURDIR)/include
```
rasky, line 12: OK can you please amend this comment to make it explicit that you are overriding a variable defining in n64.mk? It's not immediately clear from the comment, and the fact that `N64_INCLUDEDIR` isn't cited within this Makefile doesn't help.

Dragorn421: lmk about this next iteration

PR #383, opened 2023-06-27
Dust off doxygen build make

Dragorn421:


Makefile
```diff
@@ -53,14 +53,8 @@ examples:
...
 	$(MAKE) -C examples clean
 
-doxygen: doxygen.conf
-	mkdir -p doxygen/
-	doxygen doxygen.conf
 doxygen-api: doxygen-public.conf
-	mkdir -p doxygen/
 	doxygen doxygen-public.conf
-doxygen-clean:
-	rm -rf $(CURDIR)/doxygen
```
rasky, line 63: Any reason why we should remove doxygen-clean? It doesn't seem wrong to keep it and also have `clobber` depend on it. We use `clobber` as a `clean all` target.

Dragorn421: the doxygen is generated to website/, afaik the "doxygen" folder is unused besides in this makefile
I could update doxygen-clean to rm "website" though.

Additionally note build.sh runs `make clobber` at the end, but deleting doxygen docs at that point sounds undesirable

rasky: Ok let's keep it like this

PR #385, opened 2023-07-01
inspector: make the RESET button work

### LINKED ISSUE #384: `controller_read`ing constantly in the inspector (crash handler) prevents resetting
# What's happening

With the libdragon inspector ("crash handler") open, pressing the reset button does not make the console reset.
It only resets when after doing some inputs (press a button or move the stick)

# Reason

Querying the controller is done in an infinite loop without waiting in between queries, only breaking out when inputs changed.
https://github.com/DragonMinded/libdragon/blob/f8143d4f42e0fbb0fa9e749c64a26db58fd0cae4/src/inspector.c#L523-L538

from @Polprzewodnikowy ( https://discord.com/channels/205520502922543113/205522877343072266/1124689736309673984 )
> my guess is that inspector is constantly issuing commands to PIF so it's unable to run console reset code
> ...
> it's just quirk of PIF that controller communication command takes higher priority than handling reset button stuff

# Suggested fix

Just put a sleep in the loop. `wait_ms(1);` seems to work for me

rasky:
Fixes #384

### DIFF
--- a/src/inspector.c
+++ b/src/inspector.c
@@ -7,6 +7,7 @@
 #include "backtrace.h"
 #include "backtrace_internal.h"
 #include "cop0.h"
+#include "n64sys.h"
 #include <stdio.h>
 #include <stdarg.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


rasky:
The suggestion by @Dragorn421 and @Polprzewodnikowy seems to work, merging.

PR #396, opened 2023-07-08
display: Increase maximum vertical resolution to 720

sp1187:


src/display.c
```diff
@@ -282,8 +282,10 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
     registers[0] = control;
 
     /* Calculate width and scale registers */
-    assertf(res.width > 0 && res.width <= 800, "invalid width");
-    assertf(res.height > 0 && res.height <= 600, "invalid height");
+    assertf(res.width > 0, "nonpositive width");
+    assertf(res.height > 0, "nonpositive height");
+    assertf(res.width < 1024, "widths >= 1024 are not supported by RDP");
```
rasky: I don't think we should limit display size because of RDP constraints. There can be various tricks to workaround this restriction.

I would keep here the physical limit (whatever that is), and then add an assert to `rdp_attach` on trunk and `rdpq_set_color_image_raw` on unstable that verifies that the width / stride isn't bigger than 1024.

rasky (changes requested):
Thanks! See the comments below.

sp1187:
Finding the maximum non-buggy horizontal resolution turned out to be more complex as it depends on bandwidth usage on real hardware, which is affected by the color depth and anti-aliasing mode used. Therefore I removed the horizontal resolution change as more investigation is needed there.

rasky:
Thanks

PR #407, opened 2023-08-06
fix inverted NULL check in n64dso

spazzylemons:
In the n64dso tool, there is a line of code that checks if a file is null before closing it. Instead of closing if the file is not null, it only closes if the file *is* null, which should not be the case, and causes this tool to fail to compile for me. This small fix removes this error.


rasky:
Thanks!

PR #408, opened 2023-08-10
Revamped display antialiasing options into filters

SpookyIluha:
- The current display options include "anti-aliasing" settings that are cryptic and don't match expectations, this PR fixes any confusion about what is actually happening to the output with selected display filtering options.
- This PR changes and deprecates antialias_t (used only in display_init) into filter_options_t with similar options (deprecated AA options are mapped to them directly) and verbose comments, explaining each filter.
- filter_options_t is an enum with only valid and useful filtering options, such as resampling, AA and de-dithering.
- Also updated gamma comments to better illustrate their use-cases.

### DIFF
--- a/include/display.h
+++ b/include/display.h
@@ -98,7 +98,16 @@ typedef enum
     GAMMA_CORRECT_DITHER
 } gamma_t;
 
-/** @brief Valid display filter options */
+/** @brief Valid display filter options.
+ * 
+ * Libdragon uses preconfigured options for enabling certain 
+ * combinations of Video Interface filters due to a large number of wrong/invalid configurations 
+ * with very strict conditions, and to simplify the options for the user.
+ * 
+ * Like for example antialiasing requiring resampling; dedithering not working with 
+ * resampling, unless always fetching; always enabling divot filter under AA etc.
+ * 
+ * The options below provide all possible configurations that are deemed useful in development. */
 typedef enum
 {
     /** @brief All display filters are disabled */
@@ -108,6 +117,8 @@ typedef enum
      * (which is always NTSC 640x480 or PAL 640x512). 
      * This option enables a bilinear interpolation that can be used during this resize. */
     FILTERS_RESAMPLE,
+    /** @brief Reconstruct a 32-bit output from dithered 16-bit framebuffer. */
+    FILTERS_DEDITHER,
     /** @brief Resize the output image with a bilinear filter (see #FILTERS_RESAMPLE). 
      * Add a video interface anti-aliasing pass with a divot filter. 
      * To be able to see correct anti-aliased output, this display filter must be enabled,
--- a/include/display.h
+++ b/include/display.h
@@ -119,6 +119,7 @@ typedef enum
     FILTERS_RESAMPLE_ANTIALIAS_DEDITHER
 } filter_options_t;
 
+///@cond
 /** 
  * @brief Display anti-aliasing options (DEPRECATED: Use #filter_options_t instead)
  * 
@@ -145,6 +146,8 @@ typedef filter_options_t antialias_t;
  * @see #filter_options_t
  */
 #define ANTIALIAS_RESAMPLE_FETCH_ALWAYS FILTERS_RESAMPLE_ANTIALIAS_DEDITHER
+
+///@endcond
 
 /** 
  * @brief Display context (DEPRECATED: Use #surface_t instead)
--- a/include/display.h
+++ b/include/display.h
@@ -89,26 +89,62 @@ typedef enum
 /** @brief Valid gamma correction settings */
 typedef enum
 {
-    /** @brief Uncorrected gamma */
+    /** @brief Uncorrected gamma, should be used by default and with assets built by libdragon tools */
     GAMMA_NONE,
-    /** @brief Corrected gamma */
+    /** @brief Corrected gamma, should be used on a 32-bit framebuffer
+     * only if linear color space and accurate blending is important */
     GAMMA_CORRECT,
-    /** @brief Corrected gamma with hardware dither */
+    /** @brief Corrected gamma with hardware dithered output */
     GAMMA_CORRECT_DITHER
 } gamma_t;
 
-/** @brief Valid antialiasing settings */
+/** @brief Valid display filter options */
 typedef enum
 {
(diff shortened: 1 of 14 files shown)


include/vi.h
```diff
@@ -0,0 +1,240 @@
+/**
+ * @file vi.h
+ * @brief Video Interface Subsystem
+ * @ingroup display
+ */
```
rasky: If this is private for now, can you move it into src?

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+#include <stdbool.h>
+#include "regsinternal.h"
+#include "n64sys.h"
+
+/**
+ * @addtogroup display
+ * @{
+ */
+
+typedef uint32_t vi_register_t;
```
rasky: Honestly I don't think this brings much, it seems to just hide the real type and it doesn't provide specific benefits I think. We use uint32_t* almost everywhere for registers anyway.

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+ *
+ * Whenever trying to configure VI registers, 
+ * this struct and its index definitions below can be very useful 
+ * in writing comprehensive and verbose code.
+ */
+typedef struct vi_config_s{
+    vi_register_t regs[VI_REGISTERS_COUNT];
+} vi_config_t;
+
+volatile vi_register_t* VI_REGISTERS    =  ((volatile vi_register_t*)VI_REGISTERS_ADDR);
```
rasky: This definition of a variable at a global scope in a header file will cause duplicate variable definitions once we start including this file in different source files. I suggest we stick to the standard approach of using macros instead.

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+volatile vi_register_t* VI_V_VIDEO      =  &((volatile vi_register_t*)VI_REGISTERS_ADDR)[10];
+/** @brief VI Index register of start/end of the color burst enable, in half-lines. */
+volatile vi_register_t* VI_V_BURST      =  &((volatile vi_register_t*)VI_REGISTERS_ADDR)[11];
+/** @brief VI Index register of horizontal subpixel offset and 1/horizontal scale up factor. */
+volatile vi_register_t* VI_X_SCALE      =  &((volatile vi_register_t*)VI_REGISTERS_ADDR)[12];
+/** @brief VI Index register of vertical subpixel offset and 1/vertical scale up factor. */
+volatile vi_register_t* VI_Y_SCALE      =  &((volatile vi_register_t*)VI_REGISTERS_ADDR)[13];
+
+/** @brief VI register by index (0-13)*/
+#define VI_TO_REGISTER(index) ((index >= 0 && index <= VI_REGISTERS_COUNT)? &VI_REGISTERS[index] : NULL)
```
rasky: Please use parentesis around macro argument usages when used as part of an expression:
```suggestion
#define VI_TO_REGISTER(index) (((index) >= 0 && (index) <= VI_REGISTERS_COUNT)? &VI_REGISTERS[index] : NULL)
```

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+
+/** @brief VI index from register */
+#define VI_TO_INDEX(reg) ((reg - VI_REGISTERS));
```
rasky:
```suggestion
#define VI_TO_INDEX(reg) ((reg) - VI_REGISTERS)
```

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * @name Video Mode Register Presets
+ * @brief Presets to begin with when setting a particular video mode
+ * @{
+ */
+static const vi_config_t vi_ntsc_p = {.regs = {
+    0x00000000, 0x00000000, 0x00000000, 0x00000002,
+    0x00000000, 0x03e52239, 0x0000020d, 0x00000c15,
+    0x0c150c15, 0x006c02ec, 0x002501ff, 0x000e0204,
+    0x00000000, 0x00000000 }};
```
rasky: In this case there wouldn't be duplicate definitions but the contents might end up being duplicated. Anyway, for now you can keep them here, we will move to a `vi.c` once we have one.

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+#define VI_BURST_WIDTH_NTSC                 34
+#define VI_HSYNC_WIDTH_NTSC                 57
+
+#define VI_BURST_START_PAL                  64
+#define VI_VSYNC_WIDTH_PAL                  4
+#define VI_BURST_WIDTH_PAL                  35
+#define VI_HSYNC_WIDTH_PAL                  58
+
+/**  Under VI_X_SCALE   */
+#define VI_X_SCALE_SET(value)               (( 1024*value + 320 ) / 640)
```
rasky:
```suggestion
#define VI_X_SCALE_SET(value)               (( 1024*(value) + 320 ) / 640)
```

include/vi.h
```diff
@@ -0,0 +1,240 @@
...
+
+/**  Under VI_Y_SCALE   */
+#define VI_Y_SCALE_SET(value)               (( 1024*value + 120 ) / 240)
```
rasky:
```suggestion
#define VI_Y_SCALE_SET(value)               (( 1024*(value) + 120 ) / 240)
```

include/display.h
```diff
@@ -89,26 +89,76 @@ typedef enum
 /** @brief Valid gamma correction settings */
 typedef enum
 {
-    /** @brief Uncorrected gamma */
+    /** @brief Uncorrected gamma, should be used by default and with assets built by libdragon tools */
     GAMMA_NONE,
-    /** @brief Corrected gamma */
+    /** @brief Corrected gamma, should be used on a 32-bit framebuffer
+     * only if linear color space and accurate blending is important */
```
rasky:
```suggestion
     * only when assets have been produced in linear color space and accurate blending is important */
```

rasky (changes requested):
Thanks Spooky, this looks like a fantastic improvement and the new vi.h is a very good step in the direction of having a real VI library at some point.

Note that doxygen is failing, please have a look at the CI failures. We need to fix those before merging.

Also please rebase the branch on latest trunk, as I want to use the rebase-merge strategy.

rasky (approved):

PR #409, opened 2023-08-15
Enable FatFS chmod

networkfusion:
Enable FatFS to allow file attribute changes as required by https://github.com/Polprzewodnikowy/N64FlashcartMenu/issues/26

(targets unstable branch as that is what is currently required).


rasky:
Merged also to trunk

PR #411, opened 2023-08-15
Update FatFS to R0.15 Patch3

networkfusion:
We currently use R0.14 and there have been a few updates since then.

See http://elm-chan.org/fsw/ff/updates.html and http://elm-chan.org/fsw/ff/patches.html

Also incoroprates #409


networkfusion:
Tested against N64Flashcart Menu with no issues encountered.

src/fatfs/ffconf.h
```diff
@@ -66,6 +59,33 @@
 /* This option switches f_forward() function. (0:Disable or 1:Enable) */
 
 
+// FIXME: remove this, do not merge to stable without further discussion
```
rasky: Why are you adding this FIXME here?

networkfusion: I am not. That is your comment in unstable.

networkfusion: https://github.com/DragonMinded/libdragon/blob/352679922bad5403f8ea4b9123dfe0403d030bc2/src/fatfs/ffconf.h#L27

rasky: oh i had forgotten about that one. Do you still need that #ifdef?

networkfusion: Feel free to remove it. If needed I can workaround using SED before build.

networkfusion: I have removed it.

rasky:
Also merged to trunk

PR #414, opened 2023-08-19
display: Restore VI active check based on set color depth mode

sp1187:
Earlier H_VIDEO-based check introduced a regression that sometimes caused real hardware to hang when changing video resolution. While the VI will not send a video signal at all when color depth mode field is 0, it will still send a signal when H_VIDEO is 0, albeit with an entirely black picture. This could still cause the VI to be confused and crash if register changes were done outside VBLANK.

### DIFF
--- a/src/display.c
+++ b/src/display.c
@@ -149,12 +149,20 @@ static void __wait_for_vblank()
     while( reg_base[4] != 2 ) {  }
 }
 
-/** @brief Return true if VI is active (H_VIDEO != 0) */
+/** @brief Return true if VI is sending a video signal (16-bit or 32-bit color set) */
 static inline bool __is_vi_active()
 {
     volatile uint32_t *reg_base = (uint32_t *)REGISTER_BASE;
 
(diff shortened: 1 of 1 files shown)


src/display.c
```diff
@@ -149,12 +149,20 @@ static void __wait_for_vblank()
...
+/** @brief Return true if VI is sending a video signal (16-bit or 32-bit color set) */
 static inline bool __is_vi_active()
 {
     volatile uint32_t *reg_base = (uint32_t *)REGISTER_BASE;
 
-    return (reg_base[9] != 0);
+    return (reg_base[0] & 0x03) >= 2;
+}
+
+/** @brief Set active image height to 0, which keeps VI signal active but only sending a blank image */
```
rasky: I think H_VIDEO is "horizontal"?

rasky (approved):

src/display.c
sp1187: Yes, you're right. Updated comment.

PR #418, opened 2023-08-22
rdpq_rect: implement rdpq_texture_rectangle_flip

ghost:
Still missing a doc comment.


rasky:
@9nova I'm a bit more dubious on this as this primitive is a bit complex to use. I'd rather rdpq manage to lower this barrier if possible.

What about adding a rdpq_rectangle_rotated instead, with a 4-way rotation argument, and let it expand to the right primitive, adjusting the arguments accordingly? I think this would make things easier on the user.

ghost:
The map editor I used handles that at build time and just spits out a set of flags (flip dsdx, flip dtdy, swap dy/dx). So my game actually needs the low-level call: https://github.com/N64brew-Game-Jam-2023/alien-sun/blob/main/src/render.c#L470-L473

I agree though that there should be a higher level way to access it if someone is not using this kind of build step, so I'll try to add it to the PR

PR #422, opened 2023-08-26
Fix Restore VI active check based on set color depth mode

SpookyIluha:
Earlier H_VIDEO-based check introduced a regression that sometimes caused real hardware to hang when changing video resolution. A filter PR reverted this commit.

### DIFF
--- a/src/vi.h
+++ b/src/vi.h
@@ -259,7 +259,7 @@ static inline void vi_wait_for_vblank()
 /** @brief Return true if VI is sending a video signal (16-bit or 32-bit color set) */
 static inline bool vi_is_active()
 {
-    return ((*VI_CTRL & VI_CTRL_TYPE) > VI_CTRL_TYPE_BLANK? true : false);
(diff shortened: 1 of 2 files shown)


src/vi.h
```diff
@@ -254,12 +256,19 @@ static inline void vi_wait_for_vblank()
     while(*VI_V_CURRENT != VI_V_CURRENT_VBLANK ) {  }
 }
 
-/** @brief Return true if VI is active (VI_H_VIDEO != 0) */
+/** @brief Return true if VI is sending a video signal (16-bit or 32-bit color set) */
 static inline bool vi_is_active()
 {
-    return (*VI_H_VIDEO? true : false);
+    return ((*VI_CTRL & VI_CTRL_TYPE) > VI_CTRL_TYPE_BLANK? true : false);
```
sp1187: Hmm, realized it's a bit unnecessary to have a ternary expression when the returned values are just true and false. Guess ` return (*VI_CTRL & VI_CTRL_TYPE) > VI_CTRL_TYPE_BLANK;` should be enough.

sp1187 (approved):

PR #427, opened 2023-09-01
Refactor examples Makefiles

meeq:
More groundwork for my upcoming Joypad subystem PR: 

* The `examples` Makefile now uses an evaluated template for each example directory. This eliminates the tedious and error-prone process of making sure the `all`, `clean`, and `.PHONY` targets are kept in sync.
* I had to clean up the old examples that were still using bespoke Makefiles because deprecation warnings were being treated as errors. All examples now use `n64.mk`
* Removed obsolete `dfsdemo` example and references to libmikmod in build.sh


meeq:
~GitHub Actions are failing because of Mikmod 😞 
Should I just delete the `dfsdemo` test in this PR?~

Fixed in https://github.com/DragonMinded/libdragon/pull/427/commits/9d316b0570a19ef4b8f618eb452cb6cef5458f02

PR #428, opened 2023-09-01
Use refcount for timer_init/close

meeq:
More groundwork for my Joypad subsystem:

* The Joypad subsystem depends on the Timer subsystem
* The Controller subsystem is going to be converted into a shim for the Joypad subsystem
* The `ctest` example (and probably many other places in the wild) call `controller_init` before `timer_init`, so the Joypad subsystem needs to be able to initialize the Timer subsystem or things break.
* Rasky suggested in Discord that a refcount approach would be ideal for init/close so that multiple calls will be idempotent, and the subsystem will only close after a corresponding number of calls.

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -39,8 +39,11 @@
  * @{
  */
 
+/** @brief Refcount of #timer_init vs #timer_close calls. */
+static int timer_init_refcount = 0;
+
 /** @brief Internal linked list of timers */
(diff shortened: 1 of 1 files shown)


rasky:
@meeq can you please rebase this to unstable? I'd like to merge this there first

rasky:
Oh I saw it was included in the big PR to unstable, so I'm closing this for the time being.

Issue #218, opened 2021-10-30
Rework GameCube controller support


meeq:
*Extracted from https://github.com/DragonMinded/libdragon/pull/215#issuecomment-955559468*

While reviewing the controller subsystem routines for GameCube controllers to update the Doxygen comments, I have collected some notes concerning the usefulness and ergonomics of the current implementation:

1. Players that want to use a GameCube controller on N64 will need to make a passive GCN->N64 controller adapter for each controller in order to use it.
    * GameCube controllers do not identify as N64 controllers, use separate Joybus commands, and have a completely different data structure, so commercially-available GCN->N64 adapters actively convert the Joybus commands for compatibility with existing Nintendo 64 software.
    *  The use-case for using GameCube Joybus commands on a Nintendo 64 is extremely esoteric and was never supported by Nintendo or any third-party.
2. `controller_read_gc` and `controller_read_gc_origin` are not integrated with the rest of the controller subsystem.
    * A developer would need completely separate controller-handling code-paths to support GameCube controllers.
3. `struct controller_data` is wasteful, confusing, and unnecessary:
    * You will never get N64 and GCN controller data in the same result, so half of the struct is always padding.

I don't necessarily think we should remove the GC controller commands completely, but I think they need to be removed from the rest of the controller subsystem and flagged as highly-experimental if it even makes sense to keep it at all.

I strongly doubt that any developer except the original author is actually using these GC controller functions, so it might make sense to deprecate them and rework the developer experience (or just remove them entirely). Right now the usage is a lot of pain for very little gain.

meeq:
Linking this with #48 where GameCube controller support was introduced

meeq:
https://github.com/clbr/n64gctest is using the GameCube controller support functions

meeq:
Pinging @clbr to get some more context on the intentions of this functionality and what the impact of deprecation/rework would be for their projects.

meeq:
Linking to the n64-dev Discord discussion about GC controllers on N64: https://discord.com/channels/205520502922543113/205520502922543113/903630132294611044

clbr:
I cannot see that discord discussion, please post it on pastebin or here. Further, I'm appalled at your approach in general; "this homebrew is doing new, exciting things that no released software did - therefore we must remove it". Seriously, WTF?

[...]

PR #430, opened 2023-09-02
Joypad subsystem

meeq:
Stacked on top of #429 which incorporates #426, #427, #428
Finally resolves https://github.com/DragonMinded/libdragon/issues/218

## Highlights
* Introduces a new Joypad subsystem that is intended to replace the Controller subsystem.
  * Based on my working draft repository: https://github.com/meeq/JoypadTest-N64/
  * Adopts a new abstraction to unify N64 and GameCube joypad input styles into a more-common interface.
  * Intends to replace the usage of synchronous Joybus commands with an asynchronous model so that games aren't burning CPU cycles waiting on a slow serial interface.
* Deprecates the Controller subsystem and replaces it with a compatibility shim
* Refactors the inspector to use `joypad_read_n64_inputs_sync` instead of `controller_read`
* Includes a new joypadtest example ROM to demonstrate usage
* Intentionally does not update existing examples in an attempt to ensure compatibility
  * I intend on updating the examples once the Joypad subsystem is merged and stable

#### Nintendo 64

The standard Nintendo 64 controller has fewer (and different) inputs than a GameCube controller:

* Missing X & Y buttons
* Missing a C analog stick; instead there are 4 digital C directional buttons
* Does not have any pressure sensitive buttons or triggers

#### GameCube

GameCube controllers must be connected using a "passive adapter" that can be created by splicing the Joybus data pin on a GameCube controller plug into the Joybus data pin on the N64 console.

The GameCube controller has more (and different) inputs than a Nintendo 64 controller:

* Has X & Y buttons
* Has pressure-sensitive L & R triggers
* Missing C directional buttons; instead there is a C analog stick
* Some have pressure-sensitive A & B buttons
  * Analog A & B is non-functional on most GameCube controllers (official *and* third party)

### Compatibility compromises

* Nintendo 64 controllers cannot press X or Y buttons
* Nintendo 64 controllers (attempt to) emulate analog L & R triggers using digital inputs
* Nintendo 64 controllers (attempt to) emulate analog C stick using digital inputs
* GameCube controllers (attempt to) emulate C buttons based on analog C stick position
  * Players will not be able to press C-Left & C-Right or C-Up & C-Down simultaneously
* GameCube controllers always use analog mode 3 and will not read analog A & B inputs

### Rumble

Rumble motor control is supported for both N64 and GameCube controllers with a common API.

For Nintendo 64 controllers, this requires a Rumble Pak accessory.

For GameCube controllers, the 5-volt VCC pin on the controller plug must be connected to an external power source.

## Still to do

- [x] Resolve deprecation warnings in `mempak.c`
- [x] Resolve deprecation warnings in `tpak.c`
- [ ] Improve async APIs for Joypad accessories
  * Currently only the sync APIs are exposed
- [ ] Improve documentation
- [ ] Lots more testing

## Under consideration

- [ ] Deadzone handling for analog sticks
- [ ] Deprecate the mempak subsystem in favor of an async version
- [ ] Deprecate the tpak subsystem in favor of an async version
- [ ] Refactor the eeprom and eepromfs subsystems to be async
- [ ] Refactor the rtc subsystem to be async
- [ ] Port over the Bio Sensor subsystem from my working draft repository
- [ ] Implement a pointer subsystem to properly handle mouse input
- [ ] Figure out what's going on with VRU/VRS support
  * There is still one lingering function in `controller.c` that is only used by the `vrutest` example.
  * I would strongly prefer to deprecate `execute_raw_command` and offer a proper VRU/VRS API.
  * The code that exists today is completely opaque and appears to be incomplete.


include/joypad.h
```diff
@@ -0,0 +1,427 @@
+/**
+ * @file joypad.h
+ * @author Christopher Bonhage (me@christopherbonhage.com)
+ * @brief Joypad Subsystem
+ * @ingroup joypad
+ */
+
+#ifndef __JOYPAD_H
```
rasky: We're standardizing on using this convention:
```suggestion
#ifndef LIBDRAGON_JOYPAD_H
```

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+    /** @brief Joypad Port 4 */
+    JOYPAD_PORT_4     = 3,
+} joypad_port_t;
+
+/** @brief Count of Joypad ports */
+#define JOYPAD_PORT_COUNT 4
+
+/** @brief Convenience macro to validate a Joypad port number */
+#define ASSERT_JOYPAD_PORT_VALID(port) \
+    assert((port) >= 0 && (port) < JOYPAD_PORT_COUNT)
```
rasky: This should probably go into joypad_internal.h? Any reason why it's here?

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+
+/** @brief Convenience macro to iterate through all Joypad ports */
+#define JOYPAD_PORT_FOREACH(iterator_token) \
+    for (\
+        joypad_port_t iterator_token = JOYPAD_PORT_1; \
+        iterator_token < JOYPAD_PORT_COUNT; \
+        iterator_token += 1 \
+    )
```
rasky, line 94: Same

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+{
+    /** @brief Unsupported Joypad Style */
+    JOYPAD_STYLE_NONE = 0,
+    /**
+     * @brief Nintendo 64 Joypad Style.
+     *
+     * The N64 controller has fewer and different buttons than
+     * a GameCube controller: no X or Y buttons, no analog triggers,
+     * and four C-directional buttons instead of an analog C-stick.
+     */
```
rasky, line 113: Maybe I wouldn't just describe it as the set difference of features missing from the Gamecube controller. OTOH, it seems better to describe the differences in JOYPAD_STYLE_GCN, and here simply specifying "A standard N64 controller" or something like this.

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+    JOYPAD_STYLE_N64,
+    /**
+     * @brief GameCube Joypad Style.
+     *
+     * The GameCube controller has more and different buttons
+     * than a Nintendo 64 controller: X & Y buttons, analog
+     * L & R triggers, and an analog C-stick instead of buttons.
```
rasky, line 123: I would repeat here that we support this through a passive adapter.

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+{
+    /// @cond
+    union
+    {
+    /// @endcond
+        /** @brief Structure containing the state of all buttons. */
+        joypad_buttons_t buttons;
+        /// @cond
+        struct __attribute__((packed))
+        {
```
rasky, line 242: What is the rationale of this union here, and repeating the whole button definitions? Is it just for the syntactic sugar of doing `input.a` instead of `input.btn.a`? Is there something that you feel it's important?

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+     * digital C-Up and C-Down button values (-76=C-Down, +76=C-Up).
+     */
+    int8_t cstick_y;
+    /**
+     * @brief Position of the analog L trigger. (0, 255)
+     * 
+     * This value will be close to zero when no pressure is applied,
+     * and close to 200 when full pressure is applied.
+     * 
+     * For GameCube controllers, this value will be relative to its origin.
```
rasky: What is the origin in this case?

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+joypad_buttons_t joypad_get_buttons_released(joypad_port_t port);
+joypad_buttons_t joypad_get_buttons_held(joypad_port_t port);
+
+joypad_8way_t joypad_get_stick_direction(joypad_port_t port);
+joypad_8way_t joypad_get_dpad_direction(joypad_port_t port);
+joypad_8way_t joypad_get_c_direction(joypad_port_t port);
+
+int joypad_get_axis_pressed(joypad_port_t port, joypad_axis_t axis);
+int joypad_get_axis_released(joypad_port_t port, joypad_axis_t axis);
+int joypad_get_axis_held(joypad_port_t port, joypad_axis_t axis);
```
rasky, line 697: Can you please move the doxygen description of these functions here? All new modules have public APIs documented and described in their .h, leaving .c comments mostly relative to implementation details.

include/joypad.h
rasky: Same for all other include files

include/joybus_n64_accessory.h
rasky, line 1: I'm a bit fought on the name of this file. I understand that the joybus module can be used to talk also to N64-devices. but on the other hand it seems redundant to mark something as "N64-specific" in libdragon.

This is just for discussion.

include/controller.h
```diff
@@ -8,15 +8,39 @@
...
 /**
  * @addtogroup controller
  * @{
  */
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void execute_raw_command( int controller, int command, int bytesout, int bytesin, unsigned char *out, unsigned char *in );
```
rasky: Are we going to deprecate this as well?

src/joypad.c
```diff
@@ -0,0 +1,1217 @@
...
+#include "debug.h"
+#include "interrupt.h"
+#include "joypad_internal.h"
+
+/**
+ * @defgroup joypad Joypad Subsystem
+ * @ingroup peripherals
+ * @brief Joypad abstraction interface.
+ *
+ * The Joypad subsystem is the successor to the LibDragon Controller subsystem.
```
rasky: Maybe I'd remove this first sentence since in X months/years/centuries nobody will know about the controller subsystem.

src/joypad.c
```diff
@@ -0,0 +1,1217 @@
...
+ * The Joypad subsystem will automatically identify and read all connected input
+ * devices once per frame.
+ *
+ * To refer to individual ports, use the #joypad_port_t enumeration values.
+ * To iterate across all ports, use the #JOYPAD_PORT_FOREACH macro.
+ *
+ * To read the controllers, first call #joypad_scan once per frame to process
+ * the input data. #joypad_get_style will return which "style" of device is
+ * connected to a port (#joypad_style_t). #joypad_get_inputs will return the
+ * buttons and analog input state for a given controller port.
```
rasky: Reading this, I'm wondering if providing an "edge-triggered" interface for joypads / accessories changes might be a good idea. 

With "edge-triggered", I mean that the developer could register a callback for a "joypad event" which can be something like "the user just connected a GameCube joypad on port 3". This would allow the game to more easily react to these events rather than having manually to query all ports and all accessories and see if something changed. The callback should probably not be invoked under interrupt though -- I was thinking that maybe invoking it from joypad_scan() (documenting this) could be a good idea.

This is just a discussion to see if it's a good idea; it doesn't have to be done now, since this PR is already huge, we can keep this as an idea for a followup.

src/joypad.c
```diff
@@ -0,0 +1,1217 @@
...
+    joybus_exec_async(input, joypad_read_callback, NULL);
+}
+
+/**
+ * @brief Callback for VI interrupt to read and identify Joypads.
+ */
+static void joypad_vi_interrupt_callback(void)
+{
+    joypad_read_async();
+    if (joypad_identify_last_ticks + JOYPAD_IDENTIFY_INTERVAL_TICKS < timer_ticks())
```
rasky: For this specific use case, I don't think we need `timer_ticks()`. You can use the 32-bit `TICKS_READ()` and you can use `TICKS_COMPARE()` to perform an overflow-safe elapsed comparison.

The 32-bit comparison is safe unless you compare two values that differ more than 96/2 seconds apart. I don't think that can ever happen here; unless one keeps the interrupts disabled for a long time, or the VI interrupt specifically, and then re-enable it. But it seems quite an unlikely scenario.

include/joypad.h
```diff
@@ -0,0 +1,427 @@
...
+    JOYPAD_8WAY_LEFT       = 4,
+    /** @brief 8-way down-left direction. */
+    JOYPAD_8WAY_DOWN_LEFT  = 5,
+    /** @brief 8-way down direction. */
+    JOYPAD_8WAY_DOWN       = 6,
+    /** @brief 8-way down-right direction. */
+    JOYPAD_8WAY_DOWN_RIGHT = 7,
+} joypad_8way_t;
+
+joypad_inputs_t joypad_read_n64_inputs_sync(joypad_port_t port);
```
rasky: Maybe we could move this to joypad_internal.h. It seems a very specific use case, and the more general case (update joypads synchronously or more frequently) is already covered by `joypad_read_sync`. It would be hard to explain the difference between the two.

src/joypad.c
```diff
@@ -0,0 +1,1217 @@
...
+    
+    // Decrement the timer subsystem refcount (possibly closing it)
+    timer_close();
+}
+
+/**
+ * @brief Identify Joypads and wait for completion.
+ * 
+ * @param reset Whether to reset the devices.
+ */
```
rasky: When should this function called? Why? What does "resetting a device" means? I think these answers should be part of the documentation.

src/joypad.c
```diff
@@ -0,0 +1,1217 @@
...
+ * asynchronously under interrupt, so this function just synchronizes the
+ * internal state.
+ */
+void joypad_scan(void)
+{
+    ASSERT_JOYPAD_INITIALIZED();
+    // Bail early if the joypads have not been read since last call
+    static uint64_t prev_read_count = 0;
+    uint64_t read_count = joypad_read_count;
+    if (prev_read_count == read_count) { return; }
```
rasky: I'm a bit concerned about this. If I understand correclty, this means that the pressed/held status of buttons will not be affected, and will keep returning the previous one. I think this breaks my mental model for joypad_scan(). In my model, pressed/held state always refers "since the previous joypad_scan()", but with this bailout the invariant seems broken.

rasky (commented):
Thanks @meeq, this is FANTASTIC work. I'm in awe at the amount of work you poured into this, and the new library is going to be so great to work with.

I've started a review from joypad.h and joypad.c, mostly around the public API. I'll do the review in parts over several days.

include/joypad.h
meeq: It's just syntactic sugar. I wanted `joypad_inputs_t` to offer a flat namespace for convenience (so that devs could write `inputs.a` instead of forcing them to use `inputs.btn.a`) but unfortunately that means copy-pasting the content of `joypad_buttons_t` inline here.

I'd be interested to gather some wider user feedback on this, but my suspicion was that they would prefer to type fewer characters. It's not a huge burden on our side for a small quality-of-life enhancement.

include/controller.h
meeq: This was on my list of outstanding TODOs -- this function is only used by the vrutest example ROM, which is unfinished and undocumented.

The right path forward here would be building a proper subsystem for the VRU/VRS, but that's a research project which is out-of-scope for this PR.

I hope there aren't any projects depending on this function: it's very low-level and basically just a slight convenience wrapper around `joybus_exec`.

I think we should deprecate this function and the associated vrutest example. It's not fully-baked.

src/joypad.c
meeq: If I recall correctly, this was a defensive measure in case a developer called joypad_scan multiple times in a frame. I didn't see the point in re-scanning the inputs if they hadn't actually changed.

I'm not married to this check: it's an edge-case where the developer is most likely doing something wrong. I figured doing nothing was better than asserting, but perhaps the right answer is to let the developer do it wrong: it's basically harmless.

rasky: I don't think calling joypad_scan multiple times is "wrong". As you say it's just harmless: as long as the edge-triggered event semantic is preserved, you get a fully correct representation of joypad inputs, which yes, haven't changed since last scan, but that's also true across frames as well, as the user's hands don't change inputs 60 times per second anyway.

Also in the future we might want to expose a developer knob to allow polling joypads more frequently (could be exposed as "N times per frame" for instance, with N=1 being the default). In that case, it might even make sense to call joypad_scan() frequently, without necessarily being able to prove whether a joybus transaction happened since last time or not (imagine the developer playing with the N above to try different settings -- during those tests, the condition "has a joybus transaction happened since last joypad_scan" can change between true and false semi-randomically).

Anyway, preserving the semantic means dropping this check, so I'd just remove it. With this check, none of the above is true, and whether a transaction has actually been made or not affects there results of pressed/released functions for instance, which is surely not what the user expects.

So I think I'd simply remove the check, that's probably the best course of action.

include/controller.h
rasky: Even if we eventually decide to keep a "helper wrapper" around for people tinkering with joybus, surely it can't be called `execute_raw_command`. That's really a bad, unprefixed name. It will probably be called `joybus_exec_command` or something right that.

So we should probably deprecate this function ASAP anyway. Can you please add a deprecation warning to it? Probably the message should say to just use `joybus_exec` instead, and then we can revisit it in the future if we decide to introduce a helper.

include/joypad.h
rasky: OK. On the other hand, do we need `buttons` instead? I think you need it internally for implementing the pressed/released functions, but do we need it exposed? Could this be a private, double underscored, undocumented members instead?

I agree with your statement of a small burden on our side to improve user's life. What I'm just a bit concerned of is documenting two different fields that contain the same value. That seems confusing on the user side.

include/controller.h
networkfusion: We use it for pixelfx hdmi mod... https://github.com/Polprzewodnikowy/N64FlashcartMenu/blob/5ef09def6902dd9c49a636c77bfe3ac7d89e56f7/src/hdmi/hdmi.c#L53 but would be willing to update the name if required.

include/joypad.h
meeq: Good catch: this was an artifact from copying over the files from [my JoypadTest-N64 repo](https://github.com/meeq/JoypadTest-N64). Will fix.

meeq: Fixed in 6357b4aa59874d828cd04a62355aefb90294c3f7

include/joypad.h
meeq: Fixed in b45c1666d6d6df27c62ba7f076069d96ec2b3a1d

include/joypad.h
meeq: Changed `buttons` substruct to a private double-underscored undocumented member in dc2f2cae8cf9a852928b022cb25d1353cf09e72d

I originally exposed it publicly simply because I was using it internally and thought it might be useful, but I don't see any reason why a dev would need it considering `joypad_read_n64_inputs_sync` is no longer public and `joypad_get_buttons` exists.

include/joybus_n64_accessory.h
meeq: This is good feedback. LibDragon is a Nintendo 64 SDK, so it should be safe to assume its APIs are N64-specific unless noted otherwise.

Fixed in 6357b4aa59874d828cd04a62355aefb90294c3f7

I also renamed some functions, enums, and data structures that had unnecessary "N64" notations. I kept the N64 notations on the Joybus commands because Joybus commands are not N64-specific.

include/joypad.h
meeq: Fixed in c2d0d6d3dc41f0489b0e2efe92a0136a415385b1

src/joypad.c
meeq: Fixed in c2d0d6d3dc41f0489b0e2efe92a0136a415385b1

include/controller.h
meeq: Deprecated `execute_raw_command` and introduced `joybus_send_game_id` + `joybus_clear_game_id` functions in dc65d58a3330cd2e13184442277b88bca5091ce5

src/joypad.c
meeq: Fixed in 4b5917a3319684b0907411703bd90499a8bb202a (Please double-check my logic here)

include/joypad.h
meeq: I would advocate for keeping this in the public API. It nicely encapsulates the logic for looping through all controller ports. Selfishly, I kept it public because I intend on using it in my projects, and it's used in the `joypadtest` example.

I am open to renaming it, but I would like to have a good reason for making it internal-only.

include/joypad.h
meeq: Fixed in e6045b423fd4f79a67172171327d46f021767867

src/joypad.c
meeq: Fixed in e6045b423fd4f79a67172171327d46f021767867

include/joypad.h
rasky: Ok.

src/joypad.c
rasky: Seems ok!

include/joybus_accessory.h
```diff
@@ -0,0 +1,44 @@
...
+    /** @brief No N64 controller is connected. */
+    JOYBUS_ACCESSORY_IO_STATUS_NO_DEVICE = -1,
+    /** @brief No N64 accessory is connected. */
+    JOYBUS_ACCESSORY_IO_STATUS_NO_PAK    = -2,
+    /** @brief Joybus accessory communication was not successful. */
+    JOYBUS_ACCESSORY_IO_STATUS_BAD_CRC   = -3,
+} joybus_accessory_io_status_t;
+
+int joybus_accessory_read_sync(int port, uint16_t addr, uint8_t *data);
+int joybus_accessory_write_sync(int port, uint16_t addr, const uint8_t *data);
```
rasky: These should get docs in the header too.

I think we could drop the `sync` suffix here, as we default to sync/blocking for all our APIs. We explicitly mark `async` those that are async.

include/joypad.h
```diff
@@ -0,0 +1,668 @@
...
+ * * #joypad_get_stick_direction
+ * * #joypad_get_axis_pressed
+ * * #joypad_get_axis_released
+ * * #joypad_get_axis_held
+ * 
+ * This function is very fast. In fact, joypads are read in background
+ * asynchronously under interrupt, so this function just synchronizes the
+ * internal state.
+ */
+void joypad_scan(void);
```
rasky: I was thinking if we should maybe rename this to `joypad_poll`. I always found `scan` a bit of a weird word in `controller_scan`, especially after it became just a hot/cold memcpy. But I'm not a native speaker. What's your opinion on this?

include/joypad.h
```diff
@@ -0,0 +1,668 @@
...
+ * Stops reading Joypads during VI interrupt.
+ */
+void joypad_close(void);
+
+/**
+ * @brief Identify Joypads and wait for completion.
+ * 
+ * @param reset Whether to reset the devices.
+ */
+void joypad_identify_sync(bool reset);
```
rasky: What's the use case for this function? When should it be called?

include/libdragon.h
```diff
@@ -31,6 +31,8 @@
 #include "console.h"
 #include "debug.h"
 #include "joybus.h"
+#include "joybus_accessory.h"
+#include "joypad.h"
```
rasky, line 36: Shouldn't joypad_accessory.h be added as well?

include/joypad.h
```diff
@@ -0,0 +1,668 @@
...
+
+/**
+ * @brief Read Joypad inputs and wait for completion.
+ * 
+ * Implicitly scans the read inputs to save you a step.
+ */
+void joypad_read_sync(void);
```
rasky: Can you maybe move the two sync functions to the bottom of the file? I assume they are for very specialized use cases, so I think the top to bottom read should find them last.

The doc should explain when this function should be called, and that is not necessary for normal functionality of the library.

include/joybus.h
```diff
@@ -22,14 +31,246 @@
...
+ *             Buffer of recv_len bytes for reply from the Joybus device.
+ */
+void joybus_exec_raw_command(
+    int port,
+    uint8_t command_id,
+    uint8_t send_len,
+    uint8_t recv_len,
+    const uint8_t *send_data,
+    uint8_t *recv_data
+);
```
rasky: Given that we have `joybus_cmd_exec`, I don't think we need also this one. I would remove this one (and keep the implementation as implementation of the deprecated `execute_raw_command`), and then move `joybus_cmd_exec` here, keeping it as inline.

src/joybus_accessory.c
```diff
@@ -0,0 +1,231 @@
...
+            ret <<= 1;
+            if (i < JOYBUS_ACCESSORY_DATA_SIZE && data[i] & (0x01 << j))
+            {
+                ret |= 0x1;
+            }
+            ret ^= tmp;
+        }
+    }
+    return ret;
+}
```
rasky, line 57: A much faster implementation would be:

```C
uint8_t crc8(const uint8_t *data, size_t len) {
    unsigned int crc = 0;

    for (int i=0; i<len; i++) {
        unsigned int x = crc ^ data[i];
        crc = 0;
        if (x & 0x80) crc ^= 0x89;
        if (x & 0x40) crc ^= 0x86;
        if (x & 0x20) crc ^= 0x43;
        if (x & 0x10) crc ^= 0xE3;
        if (x & 0x08) crc ^= 0xB3;
        if (x & 0x04) crc ^= 0x9B;
        if (x & 0x02) crc ^= 0x8F;
        if (x & 0x01) crc ^= 0x85;
    }

    return crc;
}
```

rasky (commented):
Thanks again for all your work. This is round two, still not fully done. Takes a while...

include/joybus_accessory.h
meeq: Fixed in 78b3c2d794e1b18768bef37e5a0e4c380a02297e

include/joypad.h
meeq: Fixed in b97ad21e55673fa7310f6a8cd5307610b6b7280b

include/joypad.h
meeq: Removed from the public API in 21948595fc4c714d86aa94b32a77c127b3107d9c: I can't think of an external use-case for it.

include/joypad.h
meeq: Removed from the public API in 21948595fc4c714d86aa94b32a77c127b3107d9c: I can't think of an external use-case for this function.

include/joybus.h
meeq: Cleaned up in cecc95305bacc4e98e8878cf012e8cda620486df

src/joybus_accessory.c
meeq: This is awesome. Adopted in a9a85ebaa89470b4bf41d1a27774fadde2499ca5

include/libdragon.h
meeq: At the moment, `joypad_accessory.h` is not public.

The following are used internally by the Joypad subsystem and are not intended to be exposed:
* `joypad_accessory_detect_async`
* `joypad_rumble_pak_toggle_async`

The following are intended to be public, but require more consideration about usage:
* `joypad_get_transfer_pak_status`
* `joypad_transfer_pak_enable_async`
* `joypad_transfer_pak_load_async`
* `joypad_transfer_pak_store_async`

I don't want to use callbacks because they will be called-back under interrupt. I'm going to keep these internal for now until we can come up with a better strategy for handling asynchronous operations.

src/joypad.c
meeq: Removed from the public API.

include/joypad.h
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/430/commits/a8a5964175fcfb27363acbacf9fefe8d24976299

include/joypad.h
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/430/commits/a8a5964175fcfb27363acbacf9fefe8d24976299

include/joypad.h
meeq: Fixed in https://github.com/DragonMinded/libdragon/pull/430/commits/a8a5964175fcfb27363acbacf9fefe8d24976299

PR #435, opened 2023-09-11
Extended Model64 Support

gamemasterplc:
Now supports hierarchical and skinned models. Also implemented support for glCopyMatrixN64 to copy a matrix from a matrix stack to the current matrix stack and made gldemo use the function.

### DIFF
--- a/include/GL/gl.h
+++ b/include/GL/gl.h
@@ -26,24 +26,25 @@ typedef struct rdpq_texparms_s rdpq_texparms_t;
 #define GL_N64_reduced_aliasing         1
 #define GL_N64_interpenetrating         1
 #define GL_N64_dither_mode              1
+#define GL_N64_copy_matrix              1
 
 /* Data types */
 
-typedef uint8_t	    GLboolean;
-typedef int8_t	    GLbyte;
-typedef uint8_t	    GLubyte;
+typedef uint8_t     GLboolean;
+typedef int8_t      GLbyte;
+typedef uint8_t     GLubyte;
 typedef int16_t     GLshort;
-typedef uint16_t	GLushort;
+typedef uint16_t    GLushort;
 typedef int32_t     GLint;
-typedef uint32_t	GLuint;
+typedef uint32_t    GLuint;
 typedef uint32_t    GLsizei;
-typedef uint32_t	GLenum;
-typedef uint32_t	GLbitfield;
-typedef float	    GLfloat;
-typedef float	    GLclampf;
-typedef double	    GLdouble;
-typedef double	    GLclampd;
-typedef void	    GLvoid;
+typedef uint32_t    GLenum;
+typedef uint32_t    GLbitfield;
+typedef float       GLfloat;
+typedef float       GLclampf;
+typedef double      GLdouble;
+typedef double      GLclampd;
+typedef void        GLvoid;
 
 typedef intptr_t    GLintptrARB;
 typedef size_t      GLsizeiptrARB;
--- a/include/GL/gl.h
+++ b/include/GL/gl.h
@@ -321,6 +321,8 @@ void glPopMatrix(void);
 
 void glCurrentPaletteMatrixARB(GLint index);
 
+void glCopyMatrixN64(GLenum source);
+
 /* Texture coordinate generation */
 
 void glTexGeni(GLenum coord, GLenum pname, GLint param);
--- a/include/model64.h
+++ b/include/model64.h
@@ -22,6 +22,7 @@ typedef struct model64_node_s model64_node_t;
 model64_t *model64_load(const char *fn);
 model64_t *model64_load_buf(void *buf, int sz);
 void model64_free(model64_t *model);
+model64_t *model64_clone(model64_t *model);
 
 /**
  * @brief Return the number of meshes in this model.
--- a/include/model64.h
+++ b/include/model64.h
@@ -43,7 +43,7 @@ uint32_t model64_get_node_count(model64_t *model);
 /**
  * @brief Return the node at the specified index.
  */
-model64_node_t *model64_get_node(model64_t *model, uint32_t mesh_index);
+model64_node_t *model64_get_node(model64_t *model, uint32_t node_index);
 
 /**
  * @brief Return the first node with the specified name in the model.
--- a/include/model64.h
+++ b/include/model64.h
@@ -99,11 +99,11 @@ void model64_draw(model64_t *model);
 void model64_draw_mesh(mesh_t *mesh);
 
 /**
- * @brief Draw a single mesh.
+ * @brief Draw a single node.
  * 
- * This will draw all of the given mesh's primitives.
+ * This will draw a single mesh node.
  */
-void model64_draw_mesh(mesh_t *mesh);
+void model64_draw_node(model64_t *model, model64_node_t *node);
 
 /**
  * @brief Draw a single primitive.
--- a/include/model64.h
+++ b/include/model64.h
@@ -16,6 +16,9 @@ typedef struct mesh_s mesh_t;
 struct primitive_s;
 typedef struct primitive_s primitive_t;
 
+struct model64_node_s;
+typedef struct model64_node_s model64_node_t;
+
 model64_t *model64_load(const char *fn);
 model64_t *model64_load_buf(void *buf, int sz);
 void model64_free(model64_t *model);
@@ -30,6 +33,47 @@ uint32_t model64_get_mesh_count(model64_t *model);
  */
 mesh_t *model64_get_mesh(model64_t *model, uint32_t mesh_index);
 
+
+/**
+ * @brief Return the number of nodes in this model.
+ */
+uint32_t model64_get_node_count(model64_t *model);
+
+/**
+ * @brief Return the node at the specified index.
+ */
+model64_node_t *model64_get_node(model64_t *model, uint32_t mesh_index);
+
+/**
+ * @brief Return the first node with the specified name in the model.
+ */
+model64_node_t *model64_search_node(model64_t *model, const char *name);
+
+/**
+ * @brief Sets the position of a node in a model relative to its parent.
+ */
+void model64_set_node_pos(model64_t *model, model64_node_t *node, float x, float y, float z);
+
+/**
+ * @brief Sets the rotation of a node in a model relative to its parent in the form of an euler angle (ZYX rotation order) in radians.
+ */
+void model64_set_node_rot(model64_t *model, model64_node_t *node, float x, float y, float z);
+
+/**
+ * @brief Sets the rotation of a node in a model relative to its parent in the form of a quaternion.
+ */
+void model64_set_node_rot_quat(model64_t *model, model64_node_t *node, float x, float y, float z, float w);
+
+/**
+ * @brief Sets the scale of a node in a model relative to its parent.
+ */
+void model64_set_node_scale(model64_t *model, model64_node_t *node, float x, float y, float z);
+
+/**
+ * @brief Gets the transformation matrix between a model's root node and a node in a model.
+ */
+void model64_get_node_world_mtx(model64_t *model, model64_node_t *node, float dst[16]);
+
 /**
  * @brief Return the number of primitives in this mesh.
  */
@@ -43,9 +87,16 @@ primitive_t *model64_get_primitive(mesh_t *mesh, uint32_t primitive_index);
 /**
  * @brief Draw an entire model.
  * 
- * This will draw all primitives of all meshes that are contained the given model.
+ * This will draw all nodes that are contained in the given model while applying the relevant node matrices.
  */
 void model64_draw(model64_t *model);
+
+/**
+ * @brief Draw a single mesh.
+ * 
+ * This will draw all of the given mesh's primitives.
+ */
+void model64_draw_mesh(mesh_t *mesh);
 
 /**
  * @brief Draw a single mesh.
--- a/src/GL/rsp_gl.S
+++ b/src/GL/rsp_gl.S
@@ -22,7 +22,7 @@
         RSPQ_DefineCommand GLCmd_PreInitPipe,   4   # 0xD
         RSPQ_DefineCommand GLCmd_PreInitPipeTex,4   # 0xE
         RSPQ_DefineCommand GLCmd_SetPalettePtr, 4   # 0xF
-        RSPQ_DefineCommand GLCmd_MatrixCopy,    8   # 0x10
+        RSPQ_DefineCommand GLCmd_MatrixCopy,    4   # 0x10
     RSPQ_EndOverlayHeader
 
     RSPQ_BeginSavedState
@@ -351,8 +351,18 @@ GL_MatrixCopyMultiplyImpl:
     move ra, ra2
 
 GLCmd_MatrixCopy:
+    #############################################################
+    # GLCmd_SetFlag
+    #
+    # Sets or clears a flag
+    #
+    # ARGS:
+    #   a0: Bit 31..24: Command id
+    #       Bit 1..0: Source Matrix
+    #############################################################
     move ra2, ra
-    sll rhs, a1, 6
+    andi a0, a0, 0x3
+    sll rhs, a0, 6
     jal GL_GetMatrixIndex
     addiu rhs, %lo(GL_MATRICES)
     sll dst, mtx_index, 6
--- a/src/GL/rsp_gl.S
+++ b/src/GL/rsp_gl.S
@@ -352,19 +352,17 @@ GL_MatrixCopyMultiplyImpl:
 
 GLCmd_MatrixCopy:
     #############################################################
-    # GLCmd_SetFlag
+    # GLCmd_MatrixCopy
     #
-    # Sets or clears a flag
+    # Copies from 1 modelview stack to another
     #
     # ARGS:
     #   a0: Bit 31..24: Command id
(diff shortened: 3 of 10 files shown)


include/model64.h
```diff
@@ -30,6 +34,47 @@ uint32_t model64_get_mesh_count(model64_t *model);
...
+
+/**
+ * @brief Return the number of nodes in this model.
+ */
+uint32_t model64_get_node_count(model64_t *model);
+
+/**
+ * @brief Return the node at the specified index.
+ */
+model64_node_t *model64_get_node(model64_t *model, uint32_t mesh_index);
```
snacchus:
```suggestion
model64_node_t *model64_get_node(model64_t *model, uint32_t node_index);
```

src/GL/rsp_gl.S
```diff
@@ -348,6 +350,27 @@ GLCmd_MatrixLoad:
...
+    # GLCmd_SetFlag
+    #
+    # Sets or clears a flag
+    #
+    # ARGS:
+    #   a0: Bit 31..24: Command id
+    #       Bit 1..0: Source Matrix
+    #############################################################
+    move ra2, ra
+    andi a0, a0, 0x3
```
snacchus: This line is not needed. The command ID is in bits 31..24 (and even gets shifted upwards in the next line), so it's ignored when using it as a DMEM address
```suggestion
```

src/GL/rsp_gl.S
```diff
@@ -348,6 +350,27 @@ GLCmd_MatrixLoad:
...
+    sll rhs, a0, 6
```
snacchus: Instead of shifting here you could pre-shift on the CPU and save this opcode.

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+
+        child[i * 4 + 0] = r0;
+        child[i * 4 + 1] = r1;
+        child[i * 4 + 2] = r2;
+    }
+
+    child[12] += parent[12];
+    child[13] += parent[13];
+    child[14] += parent[14];
+}
```
snacchus, line 102: If I understand correctly, this an optimized matrix multiplication which assumes that both matrices are translation+rotation+scale only?

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+
+static void mtx_copy(float src[16], float dst[16])
+{
+    for(int i=0; i<16; i++)
+    {
+        dst[i] = src[i];
+    }
```
snacchus: Could also just use memcpy instead

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+        child[i * 4 + 1] = r1;
+        child[i * 4 + 2] = r2;
+    }
+
+    child[12] += parent[12];
+    child[13] += parent[13];
+    child[14] += parent[14];
+}
+
+static void mtx_copy(float src[16], float dst[16])
```
snacchus: Maybe this is just me, but I'm used to the convention of the first parameter being the destination and the second the source, like in memcpy for example

src/model64_internal.h
```diff
@@ -34,21 +34,73 @@ typedef struct primitive_s {
...
+    model64_joint_t *joints;    ///< Pointer to the first joint
+} model64_skin_t;
+
+/** @brief A node of the model */
+typedef struct model64_node_s {
+    char *name;                     ///< Name of a node
+    mesh_t *mesh;                   ///< Mesh a node refers to
+    model64_skin_t *skin;           ///< Skin a node refers to
+    node_transform_t transform;     ///< Initial transform of a node
+    uint32_t parent;                ///< Index of parent node
```
snacchus, line 76: I was wondering why the parent is represented with an index instead of a direct pointer. I assume the reason is that keeping pointers would make loading/unloading the structure more complicated? Or is there another reason?

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+void model64_set_node_pos(model64_t *model, model64_node_t *node, float x, float y, float z)
+{
+    uint32_t node_idx = get_node_idx(model, node);
+    assertf(node_idx < model->data->num_nodes, "Setting position of invalid node.");
+    float *dst = model->transforms[node_idx].transform.pos;
+    dst[0] = x;
+    dst[1] = y;
+    dst[2] = z;
+    calc_node_local_matrix(model, node_idx);
+    update_node_matrices(model);
```
snacchus, line 326: Would it not be sufficient to only update `node` and its descendants? You could also keep intermediate results and multiply the matrices iteratively.

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+{
+    uint32_t node_idx = get_node_idx(model, node);
+    assertf(node_idx < model->data->num_nodes, "Setting rotation of invalid node.");
+    float *dst = model->transforms[node_idx].transform.rot;
+    dst[0] = x;
+    dst[1] = y;
+    dst[2] = z;
+    dst[3] = w;
+    calc_node_local_matrix(model, node_idx);
+    update_node_matrices(model);
```
snacchus, line 354: See comment in `model64_set_node_pos`

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+void model64_set_node_scale(model64_t *model, model64_node_t *node, float x, float y, float z)
+{
+    uint32_t node_idx = get_node_idx(model, node);
+    assertf(node_idx < model->data->num_nodes, "Setting scale of invalid node.");
+    float *dst = model->transforms[node_idx].transform.scale;
+    dst[0] = x;
+    dst[1] = y;
+    dst[2] = z;
+    calc_node_local_matrix(model, node_idx);
+    update_node_matrices(model);
```
snacchus, line 366: See comment in `model64_set_node_pos`

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+        {
+            return &model->data->nodes[i];
+        }
+    }
+    return NULL;
+}
+
+static uint32_t get_node_idx(model64_t *model, model64_node_t *node)
+{
+    return node-model->data->nodes;
```
snacchus: Maybe just personal preference, but I'd prefer spaces around the minus operator for better readability

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+    dst[1] = y;
+    dst[2] = z;
+    calc_node_local_matrix(model, node_idx);
+    update_node_matrices(model);
+}
+
+void model64_get_node_world_mtx(model64_t *model, model64_node_t *node, float dst[16])
+{
+    uint32_t node_idx = get_node_idx(model, node);
+    assertf(node_idx < model->data->num_nodes, "Setting scale of invalid node.");
```
snacchus: Copy paste error in the assertion message

src/model64.c
```diff
@@ -155,9 +444,42 @@ void model64_draw_mesh(mesh_t *mesh)
...
+    if(node->mesh)
+    {
+        if(node->skin)
+        {
+            glMatrixMode(GL_MATRIX_PALETTE_ARB);
+            for(uint32_t i=0; i<node->skin->num_joints; i++)
+            {
+                glCurrentPaletteMatrixARB(i);
+                glCopyMatrixN64(GL_MODELVIEW); //Copy matrix at top of modelview stack to matrix palette
+                glMultMatrixf(model->transforms[node->skin->joints[i].node_idx].world_mtx);
```
snacchus, line 457: Does this also work if the skin's root node is not actually a root node in the entire hierarchy? I suppose it does, since it just applies the transforms of all ancestors as well, which is expected. 
In glTF, a skin also specifies its root node: https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_skin_skeleton
Is that not needed in our case? I'm not sure what it would actually be required for.

src/model64_internal.h
```diff
@@ -34,21 +34,73 @@ typedef struct primitive_s {
...
+    float rot[4];       ///< Rotation of a node (quaternion)
+    float scale[3];     ///< Scale of a node
+    float mtx[16];      ///< Matrix of a node
+} node_transform_t;
+
+/** @brief Transform state of a node of a model */
+
+typedef struct node_transform_state_s {
+    node_transform_t transform;     ///< Current transform state for a node
+    float world_mtx[16];            ///< World matrix for a node
```
snacchus, line 49: Is the naming "world_mtx" actually accurate? I guess it's in "world space" relative to the scene that is described by the model file. But one could also interpret that to mean world space in the application as a whole (even though that doesn't make much sense). Maybe "model_mtx" instead to avoid confusion?

snacchus (changes requested):
I've looked at all files except mkmodel.c because I've run out of time for now.

Looks good to me overall. The only thing I'm unsure about is whether skins should also specify their skeleton's root node like in glTF (see my comment).

Also I think updating the joint matrices could be optimized a lot more by working iteratively, instead of re-traversing the entire hierarchy again for each node.

include/GL/gl.h
```diff
@@ -321,6 +321,8 @@ void glPopMatrix(void);
 
 void glCurrentPaletteMatrixARB(GLint index);
 
+void glCopyMatrixN64(GLenum source);
```
rasky, line 325: Don't we need to declare an extension for this? Is this part of another extension?

src/GL/rsp_gl.S
```diff
@@ -348,6 +350,27 @@ GLCmd_MatrixLoad:
     j GL_MatrixMarkDirty
     move ra, ra2
 
+GLCmd_MatrixCopy:
+    #############################################################
+    # GLCmd_SetFlag
```
rasky: This name is wrong in the comment

src/model64.c
```diff
@@ -10,88 +11,376 @@
...
+}
+
+void model64_set_node_rot(model64_t *model, model64_node_t *node, float x, float y, float z)
+{
+    float cr = cosf(x * 0.5);
+    float sr = sinf(x * 0.5);
+    float cp = cosf(y * 0.5);
+    float sp = sinf(y * 0.5);
+    float cy = cosf(z * 0.5);
+    float sy = sinf(z * 0.5);
```
rasky, line 336: please use `fm_sincosf` here

tools/mkmodel/mkmodel.c
```diff
@@ -581,6 +832,120 @@ int convert(const char *infn, const char *outfn)
...
+            size_t num_values = num_components * ibm_accessor->count;
+            ibm_buffer = malloc(sizeof(float) * num_values);
+
+            // Convert all data to floats (because cgltf provides this very convenient function)
+            // TODO: More sophisticated conversion that doesn't always use floats as intermediate values
+            //       Might not be worth it since the majority of tools will probably only export floats anyway?
+            if (cgltf_accessor_unpack_floats(ibm_accessor, ibm_buffer, num_values) == 0) {
+                fprintf(stderr, "Error: failed reading attribute data\n");
+                free(ibm_buffer);
+                return 1;
```
rasky: Shouldn't this be a goto error as well?

src/model64_internal.h
gamemasterplc: It makes it easier to look up node matrices for an instance if one uses an index rather than a pointer.

src/model64.c
gamemasterplc: I think so, its copied from cgltf.

src/model64.c
gamemasterplc: I don't have fm_sincosf in my branch of libdragon.

tools/mkmodel/mkmodel.c
gamemasterplc: Yes, it should.

src/model64_internal.h
gamemasterplc: Would that also change the name of model64_get_node_world_mtx.

src/model64.c
gamemasterplc: The skeleton can be safely ignored in glTF. But I'm not sure if multiplying the modelview matrix by the node's world matrix is correct for skinned models.

src/model64.c
gamemasterplc: That was removed while I was trying to get node world matrices to work correctly.

include/GL/gl.h
gamemasterplc: Can I just add it to the extensions string in query.c? I don't think its part of another extension.

src/model64.c
snacchus: I think I get what it does now. It's basically just a shortened matrix multiplication that assumes the last row is (0, 0, 0, 1). Looks fine to me.

src/model64.c
snacchus: Ok, we can always add optimizations later on

src/model64.c
snacchus: Does it still work if you try to transform the entire model via GL_MODELVIEW?

src/model64_internal.h
snacchus: It would, but now I'm second guessing my proposed naming. Maybe "world_mtx" is fine for now.

include/GL/gl.h
snacchus: Yes, you need to add it to the extensions string, and also add a macro at the top of gl.h (all the other extensions are there as well). You can probably call the extension `GL_N64_copy_matrix` or something.

tools/mkmodel/mkmodel.c
```diff
@@ -552,13 +791,25 @@ int convert(const char *infn, const char *outfn)
...
+        goto error;
+    }
+    if (!data->scene) {
+        fprintf(stderr, "Error: input file has no default scene\n");
+        goto error;
+    }
+    if (data->scene->nodes_count <= 0) {
+        fprintf(stderr, "Error: input file has no root node in default scene\n");
+        goto error;
+    }
```
snacchus: Do we have to error out here? What if someone just wants to import some meshes and handle everything else (transform, animation etc.) themselves? Obviously skinned meshes will need nodes, but skins can only be defined on nodes anyway.

tools/mkmodel/mkmodel.c
```diff
@@ -581,6 +832,120 @@ int convert(const char *infn, const char *outfn)
             goto error;
         }
     }
+    // Convert skins
+    model->num_skins = data->skins_count;
+    model->skins = calloc(data->skins_count, sizeof(model64_skin_t));
```
snacchus: Probably shouldn't allocate if there are no skins

tools/mkmodel/mkmodel.c
```diff
@@ -581,6 +832,120 @@ int convert(const char *infn, const char *outfn)
...
+            free(ibm_buffer);
+        }
+    }
+    // Convert nodes
+    model->num_nodes = data->nodes_count;
+    // Add extra node if scene has multiple root nodes
+    if(data->scene->nodes_count > 1) {
+        model->num_nodes++;
+    }
+    model->nodes = calloc(model->num_nodes, sizeof(model64_node_t));
```
snacchus: Probably shouldn't allocate if there are no nodes (I know that would error out atm anyway, but see my other comment on that)

tools/mkmodel/mkmodel.c
```diff
@@ -150,85 +221,171 @@ void indices_write(FILE *out, uint32_t type, void *data, uint32_t count)
...
-            attribute_write(out, &primitive->position,  &vertices_placeholders[cur_primitive*ATTRIBUTE_COUNT + 0]);
-            attribute_write(out, &primitive->color,     &vertices_placeholders[cur_primitive*ATTRIBUTE_COUNT + 1]);
-            attribute_write(out, &primitive->texcoord,  &vertices_placeholders[cur_primitive*ATTRIBUTE_COUNT + 2]);
-            attribute_write(out, &primitive->normal,    &vertices_placeholders[cur_primitive*ATTRIBUTE_COUNT + 3]);
-            attribute_write(out, &primitive->mtx_index, &vertices_placeholders[cur_primitive*ATTRIBUTE_COUNT + 4]);
+            uint32_t color_data_ofs = data_ofs+attribute_get_data_size(&primitive->position);
+            uint32_t texcoord_data_ofs = color_data_ofs+attribute_get_data_size(&primitive->color);
+            uint32_t normal_data_ofs = texcoord_data_ofs+attribute_get_data_size(&primitive->texcoord);
+            uint32_t mtx_index_data_ofs = normal_data_ofs+attribute_get_data_size(&primitive->normal);
+            uint32_t index_ofs = mtx_index_data_ofs+attribute_get_data_size(&primitive->mtx_index);
```
snacchus, line 370: I get that the placeholders might be a bit cumbersome to use and I assume that's why you moved the code away from them. But I do think that they make the code easier to reason about. Now, whenever something about the file layout changes, you always have to make sure that the precomputed offsets are still correct, otherwise you'll get bugs that are hard to track. That problem does not exist with the placeholder technique.

We could try and come up with something that would make the placeholders less verbose maybe.

src/model64.c
gamemasterplc: It works if you change the modelview before calling model64_draw_node.

tools/mkmodel/mkmodel.c
gamemasterplc: I think that would involve something with a hash map that records a pointer, where the pointer points to in the file, and a list of references. The hashmap would index by pointer. We can use the stb data structures to help implement it.

tools/mkmodel/mkmodel.c
gamemasterplc: I changed it not to error out.

PR #436, opened 2023-09-15
Fix n64.mk dfs target nested directories

meeq:
The `n64.mk` rule for `%.dfs` makes an unreliable assumption that the ROM filesystem is flat. If prerequisites are nested in subdirectories of the filesystem root, this rule has unexpected behavior:

```make
%.dfs:
    @mkdir -p $(dir $@)
    @echo "    [DFS] $@"
    $(N64_MKDFS) $@ $(<D) >/dev/null
```

`$(<D)` gets the directory part of the first prerequisite. This only works correctly if the first prerequisite is in the root of the DFS filesystem. Ideally, this rule would use the common prefix of all prerequisites, but this is not trivial to do in a Makefile and has the potential to run up against command-line length limits.

This PR introduces a new `n64.mk` configuration variable: `N64_MKDFS_ROOT`, which defaults to "filesystem".
LibDragon projects will need to update their Makefile if they do not follow this convention.

Context from Discord: https://discord.com/channels/205520502922543113/974342113850445874/1152363061475278980

### DIFF
--- a/examples/dfsdemo/Makefile
+++ b/examples/dfsdemo/Makefile
@@ -9,6 +9,7 @@ OBJS = $(BUILD_DIR)/dfsdemo.o
 dfsdemo.z64: N64_ROM_TITLE = "DragonFS Demo"
 dfsdemo.z64: $(BUILD_DIR)/dfsdemo.dfs
 
+$(BUILD_DIR)/dfsdemo.dfs: filesystem/a_directory/a_file.txt
 $(BUILD_DIR)/dfsdemo.dfs: $(wildcard filesystem/*)
 $(BUILD_DIR)/dfsdemo.elf: $(OBJS)
 
--- a/n64.mk
+++ b/n64.mk
@@ -1,6 +1,9 @@
 BUILD_DIR ?= .
 SOURCE_DIR ?= .
 
+# Override this if your project uses a different directory for your DFS filesystem root
+N64_MKDFS_ROOT ?= filesystem
+
 N64_ROM_TITLE = "Made with libdragon" # Override this with the name of your game or project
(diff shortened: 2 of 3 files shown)


examples/dfsdemo/Makefile
```diff
@@ -9,6 +9,7 @@ OBJS = $(BUILD_DIR)/dfsdemo.o
 dfsdemo.z64: N64_ROM_TITLE = "DragonFS Demo"
 dfsdemo.z64: $(BUILD_DIR)/dfsdemo.dfs
 
+$(BUILD_DIR)/dfsdemo.dfs: filesystem/a_directory/a_file.txt
 $(BUILD_DIR)/dfsdemo.dfs: $(wildcard filesystem/*)
```
meeq, line 13: In my opinion, the usage of `wildcard` here is a Makefile anti-pattern: `$(BUILD_DIR)/dfsdemo.dfs` does not get re-built when a file in a subdirectory of `filesystem` is touched because wildcard is not recursive.

I didn't want to touch that as part of fixing this issue, but I did want to make sure at least one of our examples exercised the edge case that this PR addresses.

n64.mk
```diff
@@ -102,7 +102,12 @@ RSPASFLAGS+=-MMD
 %.dfs:
 	@mkdir -p $(dir $@)
 	@echo "    [DFS] $@"
-	$(N64_MKDFS) $@ $(<D) >/dev/null
+	set -e; \
+	MKDFSROOT="$(shell python3 -c 'import os.path; print( os.path.commonpath( """$^""".split() ) ) ')"; \
```
meeq: I don't love this, but every alternative I tried was worse. Can we safely assume Python exists?

meeq: Obviously we cannot assume Python exists, as demonstrated by GitHub Actions. 😢

PR #440, opened 2023-09-17
Add borders and aspect ratio support to resolution_t

SpookyIluha:
With this PR, there's now a way to tell libdragon to have no borders in the output (useful for emulators and modern TV's)

Also with this PR users can add hardware VI borders around the framebuffer to midigate overscan on most CRT TV's
Borders scale the image down instead of cropping it, so that the framebuffer is fully visible.

There's also now an aspect ratio setting which can be used to achieve letterbox widescreen.

With this PR NTSC/PAL/M-PAL systems and monitors should output identical pictures.


rasky:
I'm working on this with @SpookyIluha, we are testing many edge cases. Putting back to draft status for now

rasky:
A very similar API change was merged here:
https://github.com/DragonMinded/libdragon/pull/646

This PR also attempted to reverse engineer the actual VI border formula but that part will need to be rewritten as part of the new VI library anyway, so keeping this PR open doesn't help much at this point.

PR #441, opened 2023-09-17
mksprite: compute YUV error for IHQ

ghost:
So as discussed earlier I tried to use also the chroma U and V channels for IHQ format's intensity value selection. Unlike early experiments showed, it seems to just add a bit of contrast. I don't know why the results were different earlier.

Anyway, I think it's a small improvement.
image

There's room for optimization because it's clearly slower than the old error function:
```
Test:
time for i in 1 2 3 4 5 6 7 8 9 10; do ../mksprite --format IHQ ../stone.png ; done

Before:
real	0m0,109s
real	0m0,077s
real	0m0,088s

After:
real	0m0,151s
real	0m0,158s
real	0m0,145s
```

Also, some glitches are visible at the bottom row. Very obvious in "castle" image at least.


ghost:
The original images for reference. 
grass circuit stone castle

rasky:
I think the glitches might be a bug in my bilinear interpolation, I’ll have a look. Patch is ok for me

PR #450, opened 2023-09-25
Refactor videoplayer example Makefile

meeq:
The videoplayer Makefile will attempt to download the movie and re-encode it for playback on N64.

If curl/wget and ffmpeg are not available, it will log a message but will not fail.

If the movie cannot be encoded, it will be deleted from the filesystem when the ROM builds.

If the ROM cannot find the movie, it will fail at runtime with this message:
image


meeq:
~Argh, portability issues. `du` doesn't support `--bytes` on macOS.~

Edit: fixed in https://github.com/DragonMinded/libdragon/pull/450/commits/07c67f47427d951dc17225a88fa9e675492d9827

PR #452, opened 2023-09-26
Support Animations in Model64

gamemasterplc:
Added support for animations in model64. Up to 3 active animations can be played at once if they do affect different bones.


src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+
+static bool is_anim_slot_valid(model64_anim_slot_t slot)
+{
+    return (slot >= 0) && (slot < MAX_ACTIVE_ANIMS);
+}
+
+void model64_anim_play(model64_t *model, const char *anim, model64_anim_slot_t slot, bool paused, float start_time)
+{
+    int32_t anim_index = search_anim_index(model, anim);
+    assertf(anim_index != -1, "Invalid animation name");
```
snacchus: I'd suggest a more descriptive error message, e.g. "Animation '%s' was not found"

src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+    if(model->data->stream_buf_size != 0 && !model->active_anims[slot].stream_buf[0]) {
+        model->active_anims[slot].stream_buf[0] = malloc_uncached(model->data->stream_buf_size);
+        model->active_anims[slot].stream_buf[1] = malloc_uncached(model->data->stream_buf_size);
```
snacchus: These buffers are read on CPU, right? In that case I would allocate them with cached memory (and invalidate them accordingly), at least for now. If we offload the processing to the RSP at some point we can switch to uncached again when the time comes.

src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+void model64_anim_stop(model64_t *model, model64_anim_slot_t slot)
+{
+    assertf(is_anim_slot_valid(slot), "Invalid animation ID");
+    model->active_anims[slot].index = -1;
+}
+
+float model64_anim_get_length(model64_t *model, const char *anim)
+{
+    int32_t anim_index = search_anim_index(model, anim);
+    assertf(anim_index != -1, "Invalid animation name");
```
snacchus: See my other comment in `model64_anim_play`

src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+        next_data_idx = curr_data_idx+1;
+        buf_info->time = (anim_slot_ptr->time-((float)curr_data_idx/curr_anim->frame_rate))*curr_anim->frame_rate;
+    }
+    if(model->data->stream_buf_size > 0) {
+        uint32_t rom_addr = (uint32_t)model->data->anim_data_handle;
+        rom_addr += (uint32_t)curr_anim->data;
+        dma_read(anim_slot_ptr->stream_buf[0], rom_addr+(curr_data_idx*frame_size), frame_size);
+        buf_info->curr_buf = anim_slot_ptr->stream_buf[0];
+        dma_read(anim_slot_ptr->stream_buf[1], rom_addr+(next_data_idx*frame_size), frame_size);
+        buf_info->next_buf = anim_slot_ptr->stream_buf[1];
```
snacchus: Could we reuse previously loaded data here? For example if `stream_buf[1]` already contains the current frame (because it was the "next frame" previously), we don't need to DMA it again. And the other way around of course in case time is moving backwards.

src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+{
+    for(int i=0; i<MAX_ACTIVE_ANIMS; i++) {
+        if(model->active_anims[i].index == -1 || model->active_anims[i].paused || model->active_anims[i].speed == 0) {
+            if(model->active_anims[i].index != -1 && model->active_anims[i].new_pose) {
+                calc_anim_pose(model, i);
+            }
+            model->active_anims[i].new_pose = false;
+            continue;
+        }
+        model->active_anims[i].time += model->active_anims[i].speed*dt;
```
snacchus: Do we need to worry about accumulating error here? From what I understand, Users can set the current time freely via `model64_anim_set_time` anyway, so why does this function take a delta? If it took the absolute time instead we could let users handle calculating it, to whatever precision they require.

src/model64.c
```diff
@@ -464,3 +524,207 @@ void model64_draw(model64_t *model)
...
+        if(model->active_anims[i].loop) {
+            model64_anim_t *curr_anim = &model->data->anims[model->active_anims[i].index];
+            model->active_anims[i].time = fm_fmodf(model->active_anims[i].time, curr_anim->duration);
+            if(model->active_anims[i].time < 0) {
```
snacchus: In which case could `time` be negative here? If speed is negative?

src/model64_internal.h
```diff
@@ -88,19 +109,43 @@ typedef struct model64_data_s {
...
+    node_transform_state_t *transforms;             ///< List of transforms for each bone in a model instance
+    anim_state_t active_anims[MAX_ACTIVE_ANIMS];    ///< List of active animations
 } model64_t;
 
+/** @brief State of an active animation */
+typedef struct anim_buf_info_s {
+    void *curr_buf;     ///< Data buffer for current frame
+    void *next_buf;     ///< Data buffer for next frame
+    float time;         ///< Interpolation factor between current and next frame
+} anim_buf_info_t;
```
snacchus: This struct could be moved to model64.c since it is only used there as an implementation detail

include/model64.h
```diff
@@ -19,9 +30,13 @@ typedef struct primitive_s primitive_t;
 struct model64_node_s;
 typedef struct model64_node_s model64_node_t;
 
+model64_data_t *model64_load_data(const char *fn);
```
snacchus: There was already a discussion about this on Discord, but here is my take: I can see the point in separating the model data ("blueprint"?) from model "instances". If we were to go that way, I'd double down on the separation and remove `model64_clone`, `model64_load` and `model64_load_buf`. That way it's very clear how to load a model (`model64_load_data`) and how to create instances (`model64_create`). Otherwise there would seem to be many ways of achieving the same thing (creating a model instance), which would lead to confusion IMO. Also I'd work on the naming a bit more.

However, I must say that I also agree with @rasky in that this is just increasing the API surface with functions that kind of do the same things as the already existing ones. The only actual difference is that with the extra functions, you can load model data without actually creating an instance, which would save some memory on the order of a couple of kilobytes, depending on how many nodes the model contains. We can discuss whether that is worth it. OTOH, with the changes I suggested above, we'd go from 4 functions to 5 in total, so maybe it would actually be fine.

src/model64_internal.h
```diff
@@ -88,19 +109,43 @@ typedef struct model64_data_s {
...
+    void *stream_buf[2];    ///< Buffers for streaming animation
+} anim_state_t;
+
 /** @brief A model64 instance */
 typedef struct model64_s {
-    model64_data_t *data;                   ///< Pointer to the model data this instance refers to
-    node_transform_state_t *transforms;     ///< List of transforms for each bone in a model instance
+    model64_data_t *data;                           ///< Pointer to the model data this instance refers to
+    node_transform_state_t *transforms;             ///< List of transforms for each bone in a model instance
+    anim_state_t active_anims[MAX_ACTIVE_ANIMS];    ///< List of active animations
```
snacchus: I think the animation state should be heap allocated too, like the transforms. Otherwise we are always allocating space for it even if a model doesn't contain animations.

tools/mkmodel/mkmodel.c
```diff
@@ -72,8 +88,10 @@ void print_args( char * name )
     fprintf(stderr, "Usage: %s [flags] <input files...>\n", name);
     fprintf(stderr, "\n");
     fprintf(stderr, "Command-line flags:\n");
-    fprintf(stderr, "   -o/--output <dir>         Specify output directory (default: .)\n");
-    fprintf(stderr, "   -v/--verbose              Verbose output\n");
+    fprintf(stderr, "   -o/--output <dir>       Specify output directory (default: .)\n");
+    fprintf(stderr, "   --anim-no-stream        Output animation data for not streaming.\n");
```
snacchus: I'd suggest "Disable animation streaming" as the description

src/model64.c
gamemasterplc: The buffers are definitely read on the CPU.

src/model64.c
gamemasterplc: model64_anim_set_time is for seeking within the animation. The function takes in a delta for advancing the time of all active animations.

src/model64.c
gamemasterplc: Time can be negative if speed is negative because fmodf returns a negative value if time is negative. Negative speed can also cause time to go below 0.

src/model64_internal.h
gamemasterplc: Would each animation state be heap allocated or just the whole list of active animations. The stream buffers are still only allocated when playing the first animation on a slot even when unchanged.

include/model64.h
gamemasterplc: So, should I just remove the existing clone and plain load/load_buf functions. What should I rename the other functions to?

src/model64.c
gamemasterplc: Unfortunately, reusing the existing buffers is pretty hard without copying via DMA or the CPU.

snacchus: Why is it hard? You already have a level of indirection that you could use for this. You could swap the values of `buf_info->curr_buf` and `buf_info->next_buf` if required. Unless I'm missing something?

include/model64.h
gamemasterplc: Privatized these functions for now.

src/model64.c
gamemasterplc: Unfortunately, I have trouble getting interpolation between read keyframes to work correctly when reusing streaming buffers which results in visible shaking or jerking to keyframes especially at slow speeds. Btw, I'm also trying to skip both the current and next fetch when the frame number hasn't changed.

src/model64.c
```diff
@@ -545,6 +541,25 @@ static int32_t search_anim_index(model64_t *model, const char *name)
     return -1;
 }
 
+static void alloc_anim_slot(model64_t *model, model64_anim_slot_t slot)
+{
+    if(model->active_anims[slot]) {
+        return;
+    }
```
rasky: In general, this style of coding is slower on N64 because you pay the overhead of a function call (which includes switching to a different instruction cache line) just to check whether there's something to do. It would be better to move this check into all callers (and even mark it as unlikely) to make sure it impacts near zero during runtime

snacchus (approved):
Looks good to me

PR #456, opened 2023-10-09
mixer: Add helper function for common pattern when playing audio

thekovic:
There's a common sequence of calls that users are advised to call many times a frame in order to satiate the audio processing and avoid audio stutter. This PR adds an `audio_poll` API to libdragon that performs this action to streamline this pattern for users and clarify what the intended use of the audio subsystem is, in a similar vein to the `display_get` function from the display subsystem.

Addendum: The pattern relevant to this function is expected to disappear once audio is handled in a separate thread. Therefore, this function shouldn't be stabilized so that it can be deprecated and removed when threading support is added.

### DIFF
--- a/include/mixer.h
+++ b/include/mixer.h
@@ -358,6 +358,18 @@ void mixer_unthrottle(void);
 void mixer_poll(int16_t *out, int nsamples);
 
 /**
+ * @brief Request the mixer to try and write audio samples to be played,
+ * if possible.
+ * 
+ * This function is a user helper for asking the mixer and audio subsystems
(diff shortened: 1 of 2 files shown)


rasky (changes requested):
Since the function uses mixer_poll, it can't be defined in audio.c because it would be a layering violation (audio.c knows nothing of the mixer). It has to be defined in mixer.c/h, with a different name.

thekovic (commented):
Moved to mixer.c, function is now called `mixer_try_play` (could be renamed if you think)

rasky (approved):

PR #457, opened 2023-10-09
Fix the default reset procedure

SpookyIluha:
Added a die() function in the n64sys.h that is used whenever a reset button is pressed. Its triggered from a VI interrupt.

This will cause the N64 to gracefully shut down its processors just before a hard reset occurs, so that it will not crash or have unexpected results. 

The reset procedure still supports user behavior up to 200ms for special effects if need be.

### DIFF
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -226,7 +226,7 @@ void wait_ms( unsigned long wait_ms );
  * The system will recover after a reset or power cycle.
  * 
  */
-void die();
+void die(void);
 
 /**
  * @brief Force a data cache invalidate over a memory region
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -217,6 +217,18 @@ void wait_ticks( unsigned long wait );
 void wait_ms( unsigned long wait_ms );
 
 /**
+ * @brief Force a complete halt of all processors
+ *
+ * @note It should occur whenever a reset has been triggered 
+ * and its past its RESET_TIME_LENGTH grace time period.
+ * This function will shut down the RSP and the CPU, blank the VI.
+ * Eventually the RDP will flush and complete its work as well.
+ * The system will recover after a reset or power cycle.
+ * 
+ */
+void die();
+
+/**
  * @brief Force a data cache invalidate over a memory region
  *
  * Use this to force the N64 to update cache from RDRAM.
--- a/src/n64sys.c
+++ b/src/n64sys.c
@@ -5,6 +5,7 @@
  */
 
 #include <stdint.h>
+#include <stdlib.h>
 #include <assert.h>
 #include <malloc.h>
 #include "n64sys.h"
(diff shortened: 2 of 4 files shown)


include/n64sys.h
```diff
@@ -216,6 +216,18 @@ inline volatile unsigned long get_ticks_ms(void)
...
+ * @brief Force a complete halt of all processors
+ *
+ * @note It should occur whenever a reset has been triggered 
+ * and its past its RESET_TIME_LENGTH grace time period.
+ * This function will shut down the RSP and the CPU, blank the VI.
+ * Eventually the RDP will flush and complete its work as well.
+ * The system will recover after a reset or power cycle.
+ * 
+ */
+void die();
```
rasky: I think the correct C syntax is `void die(void)`. 

`void die();` is an old C syntax to signal that the argument list is "unknown", so the compiler will accept any potential list of argument. This is often confusing because in C++, instead, `void die()` means "no arguments accepted. But since this is a C declaration, we must follow the C convention.

You can try calling `die(1,2,3);` and it should successfully compile with this declaration.

src/n64sys.c
```diff
@@ -352,6 +357,29 @@ void wait_ms( unsigned long wait_ms )
...
+ * This function will shut down the RSP and the CPU, blank the VI.
+ * Eventually the RDP will flush and complete its work as well.
+ * The system will recover after a reset or power cycle.
+ * 
+ */
+__attribute__((noreturn)) void die(){
+    // Halt the RSP
+    *SP_STATUS = SP_WSTATUS_SET_HALT;
+    // Flush the RDP
+    *DP_STATUS = DP_WSTATUS_SET_FLUSH | DP_WSTATUS_SET_FREEZE;
```
rasky, line 377: I think here we want to pulse this, that is, after setting both FLUSH and FREEZE, reset them.

src/n64sys.c
```diff
@@ -352,6 +357,29 @@ void wait_ms( unsigned long wait_ms )
...
+    // Shut the video off
+    *VI_CTRL = *VI_CTRL & (~VI_CTRL_TYPE);
+    // Can't have any interrupts here
+    disable_interrupts();
```
rasky: Maybe start with this, so that if interrupts are enabled, there is no way they can interfere (eg: restart the RSP).

src/n64sys.c
```diff
@@ -352,6 +357,29 @@ void wait_ms( unsigned long wait_ms )
...
+    // Spin infinitely
+    while(true){};
```
rasky: Please call `abort()` here, so that the intent is clear. This way we also have a single abort point that might be useful for debugging (eg: gdb can autoset a breakpoint in abort() to catch when the program aborts for any reason).

src/vi.h
```diff
@@ -193,6 +193,11 @@ static const vi_config_t vi_config_presets[2][3] = {
 /** @brief VI_BURST Register: PAL default horizontal sync width in pixels. */
 #define VI_HSYNC_WIDTH_PAL                  58
 
+/** @brief VI period for showing one NTSC and MPAL picture in ms. */
+#define VI_PERIOD_NTSC_MPAL                 (1000/60)
+/** @brief VI period for showing one PAL picture in ms. */
+#define VI_PERIOD_PAL                       (1000/50)
```
rasky: I think these should be floating points, to avoid losing precision. Since they are going into vi.h which will be public one day, I'd rather export an accurate period for our users, rather than an approximated one.

PR #459, opened 2023-10-21
Implement Animation Compression

gamemasterplc:


src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
     vec_normalize(out, 4);
 }
 
-static void calc_anim_pose(model64_t *model, int anim_slot)
+static void decode_normalized_u16(uint16_t *in, float *out, size_t count, float min, float max)
+{
+    for(size_t i=0; i<count; i++) {
+        out[i] = ((in[i]/65535.0f)*(max-min))+min;
```
rasky: you could store in the animation just two factors:

```C
f1 = (max-min) / 65535.0f;
f2 = min;
```

and then do:

```C
out[i] = in[i] * f1 + f2;
```

This should save a couple of operations at runtime

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+}
+
+static void decode_quaternion(uint16_t *in, float *out)
+{
+    uint16_t values[3];
+    int max_component;
+    values[0] = ((in[0] & 0x1FFF) << 2)|(in[1] >> 14);
+    values[1] = ((in[1] & 0x3FFF) << 1)|(in[2] >> 15);
+    values[2] = in[2] & 0x7FFF;
+    max_component = (in[0] & 0x6000) >> 13;
```
rasky: I suggest to rearrange data to minimize operations here:

```C
values[0] = in[0] & 0x7FFF;
values[1] = in[1] & 0x7FFF;
values[2] = in[2] & 0x7FFF;
max_component = (in[0] & 0x8000) : 0 ? (in[1] & 0x8000) : 1 : 2;
sign = (int16_t)in[2];   // no need to normalize to 1 or -1, see below
```

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+    float decoded_values[3];
+    for(size_t i=0; i<3; i++) {
+        decoded_values[i] = (values[i]*0.000043159689f)-0.70710678f;
+    }
+    float sign = (in[0] & 0x8000) ? -1 : 1;
+    float mag2_decoded = (decoded_values[0]*decoded_values[0])+(decoded_values[1]*decoded_values[1])+(decoded_values[2]*decoded_values[2]);
+    float derived_component = sign*sqrtf(1.0f-mag2_decoded);
```
rasky: use `copysignf(sqrtf(1.0f - mag2_decoded), (float)sign)`

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+                case ANIM_COMPONENT_SCALE:
+                    decode_normalized_u16(frame->data, &decoded_values[j][0], 3, curr_anim->scale_min, curr_anim->scale_max);
+                    decoded_values[j][3] = 0.0f;
+                    break;
+                  
+                default:
+                    decoded_values[j][0] = decoded_values[j][1] = decoded_values[j][2] = decoded_values[j][3] = 0.0f;
+                    break;
+            }
+        }
```
rasky: I think it's possible to avoid doing most of the above work at each `calc_anim_pose` call, and just doing it when a new keyframe arrives. `decoded_values` can be stored somewhere in `anim_state` so that `calc_anim_pose` can just perform an interpolation.

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+        if(anim_state->time > time_next) {
+            weight = 1.0f;
+        } else {
+            if(anim_state->time < time) {
+                weight = 0.0f;
+            } else {
+                weight = (anim_state->time-time)/(time_next-time);
+            }
+            
+        }
```
rasky: I think this block should also be simplified by also storing `1/(time_next-time)`.

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
-        next_buf += data_size;
+        if(out == NULL) {
+            continue;
+        }
+        if(component == ANIM_COMPONENT_ROT) {
+            quat_lerp(out, &decoded_values[1][0], &decoded_values[2][0], weight);
+        } else {
+            catmull_calc_vec(&decoded_values[0][0], &decoded_values[1][0], &decoded_values[2][0], &decoded_values[3][0], out, weight, out_count);
+        }
+        calc_node_local_matrix(model, node);
```
rasky: Shouldn't this be done just once per node? It looks like it's done once per done per component.

src/model64.c
```diff
@@ -740,38 +688,185 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+static void calc_anim_pose(model64_t *model, model64_anim_slot_t anim_slot)
+{
+    anim_state_t *anim_state = model->active_anims[anim_slot];
+    model64_anim_t *curr_anim = &model->data->anims[anim_state->index];
+    float decoded_values[4][4];
+    for(uint32_t i=0; i<curr_anim->num_tracks; i++) {
+        uint32_t component = curr_anim->tracks[i] >> 14;
+        uint16_t node = curr_anim->tracks[i] & 0x3FFF;
+        for(uint32_t j=0; j<4; j++) {
+            int index = (i*4)+((j+anim_state->buf_idx[i]) % 4);
```
rasky: It seems to me that buf_idx is not really required. What it does is to save copying keyframes within their array when a new keyframe arrives. On the other hand:

* Copying a keyframe is literally 4 opcodes, given they are now just 16 bytes
* Moving three keyframes is only required when a new keyframe arrive, that with animation compression should be a rare event
* On the other hand, we pay for the buf_idx indirection every frame
* The code is harder to read and reason about

I would suggest dropping it.

tools/mkmodel/mkmodel.c
```diff
@@ -1100,9 +1120,93 @@ int read_anim_channel(cgltf_data *data, cgltf_animation_channel *in_channel, glt
...
         return 1;
     }
+    out_channel->output_min[0] = FLT_MAX;
+    out_channel->output_min[1] = FLT_MAX;
+    out_channel->output_min[2] = FLT_MAX;
+    out_channel->output_min[3] = FLT_MAX;
+    out_channel->output_max[0] = -FLT_MAX;
+    out_channel->output_max[1] = -FLT_MAX;
+    out_channel->output_max[2] = -FLT_MAX;
+    out_channel->output_max[3] = -FLT_MAX;
```
rasky, line 1134: I think these should be `FLT_MIN`

gamemasterplc: It should not be FLT_MIN since output_max could be less than 0 and FLT_MIN is approximately 0 rather than approximately  -(2^127). FLT_MAX is approximately 2^127.

src/model64.c
gamemasterplc: It could be done that way if one sorts the tracks.

src/model64.c
```diff
@@ -740,38 +686,184 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+    int max_component;
+    values[0] = in[0] & 0x7FFF;
+    values[1] = in[1] & 0x7FFF;
+    values[2] = in[2] & 0x7FFF;
+    max_component = (in[0]>>15) | (in[1]>>15<<1);
+    float decoded_values[3];
+    for(size_t i=0; i<3; i++) {
+        decoded_values[i] = (values[i]*0.000043159689f)-0.70710678f;
+    }
+    int16_t sign = (int16_t)in[2];;
```
snacchus: double semicolon here

src/model64.c
```diff
@@ -740,38 +686,184 @@ static void quat_lerp(float *out, float *in1, float *in2, float time)
...
+}
+
+static void copy_waiting_frame(model64_t *model, model64_anim_slot_t anim_slot)
+{
+    anim_state_t *anim_state = model->active_anims[anim_slot];
+    uint16_t track = anim_state->waiting_frame->track;
+    anim_state->frames[track*4] = anim_state->frames[(track*4)+1];
+    anim_state->frames[(track*4)+1] = anim_state->frames[(track*4)+2];
+    anim_state->frames[(track*4)+2] = anim_state->frames[(track*4)+3];
+    anim_state->frames[(track*4)+3] = *anim_state->waiting_frame;
```
snacchus: Could we maybe avoid doing these mem copies by treating the (sub) array of frames as a ring buffer? This would require keeping a head index for each track, but maybe it's worth doing. Or is there a specific reason why the frames need to be in order?

tools/mkmodel/mkmodel.c
```diff
@@ -95,7 +122,6 @@ void print_args( char * name )
     fprintf(stderr, "Command-line flags:\n");
     fprintf(stderr, "   -o/--output <dir>       Specify output directory (default: .)\n");
     fprintf(stderr, "   --anim-no-stream        Disable animation streaming\n");
-    fprintf(stderr, "   --anim-fps <value>      Frame rate of the animation (default: 30.0)\n");
```
snacchus, line 98: Why is this getting removed?

tools/mkmodel/mkmodel.c
```diff
@@ -1154,15 +1260,121 @@ void sample_anim_channel(gltf_anim_channel_t *channel, float time, float *out)
...
+    float d = 2*p2;
+    return 0.5f * (a+b+c+d);
+}
+
+void catmull_calc_vec(float *p1, float *p2, float *p3, float *p4, float *out, float t, int num_values)
+{
+    for(int i=0; i<num_values; i++) {
+        out[i] = catmull_calc(p1[i], p2[i], p3[i], p4[i], t);
+    }
+}
```
snacchus: These are duplicated from model64.c, right? Maybe put them in a shared header?

src/model64.c
gamemasterplc: I initially did it that way but it complicated the code some.

tools/mkmodel/mkmodel.c
gamemasterplc: Because animation framerate is just a parameter of the reduction algorithm now. It is only used to generate the initial samples.

tools/mkmodel/mkmodel.c
gamemasterplc: Did that

src/model64.c
gamemasterplc: Is doing the operation and clamping the weight value afterwards fine.

PR #471, opened 2023-12-11
Unstable: Fix Windows tools CI

networkfusion:
The latest tools in unstable require use of the MIPS GCC toolchain.

It seems that environment variables are not carried across tasks, so uses it directly for the makefile build.

Also updates GitHub actions in all of the CI with the exception of doxygen-action (as it fails build).


rasky:
Thanks!

PR #472, opened 2023-12-11
Unstable: Update Dockerfile image version

### LINKED ISSUE #466: Ubuntu in Dockerfile is EOL, update to 22.04

networkfusion:
Use the latest image rather than an EOL version.

The solution could be to adjust (at minimum) `FROM ubuntu:18.04` in `Dockerfile` to be `FROM ubuntu:22.04`, but we are not doing anything that "should" require it being fixed to a specific version.

Although this PR targets the Unstable branch, it is equally valid (and should have no issues) with main/stable/trunk.

Fixes #466

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # V0 - Use this comment to force a re-build without changing the contents
 
 # Stage 1 - Build the toolchain
-FROM ubuntu:18.04
+FROM ubuntu:latest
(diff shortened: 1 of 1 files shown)


rasky:
I'd rather have a fixed version, less surprises I'd say.

networkfusion:
@rasky , done.

PR #479, opened 2024-01-05
mkmodel: simple model64 texture support

ghost:
The commit message shows the general design:

>The path of the base color texture of each GLTF primitive is now stored in converted models. Texture's file extension is changed to '.sprite' and a 'rom://' prefix is added at runtime.
>
>Images are assumed to be in ROM filesystem in the same paths they are reported in the input GLTF file.  Models can share textures and they are deduplicated at model64 load time.
>
>It's users responsibility to convert textures to sprites.

I'm not sure the runtime memory management part is 100% bulletproof. Tested sharing manually by spawning two model64 instances from the same file.

### DIFF
--- a/src/model64.c
+++ b/src/model64.c
@@ -11,14 +11,162 @@
 #include "model64_internal.h"
 #include "asset.h"
 #include "debug.h"
+#include "sprite.h"
 #include "utils.h"
+#include "rdpq_tex.h"
 
 #include "model64_catmull.h"
+
+
+/** @brief Loading state of a texture_entry_t */
+typedef enum {
+    ENTRY_STATE_EMPTY = 0,
+    ENTRY_STATE_SPRITE_LOADED,
+    ENTRY_STATE_FULL
+} texture_entry_state_t;
+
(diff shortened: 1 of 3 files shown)


ghost:
Changes to previous version:
- prefix is extracted so relative paths also work automatically
- texture table is freed when not in use
- primitive_t struct has now both a `local_texture` and `shared_texture` indices so that it can be "unloaded" back to the serialized form

ghost:
I put in simplistic logic for non-power-of-two textures. A dimension is always clamped if it's not POT:
image

```c
  // If a dimension is not a power of two then clamp and otherwise repeat.
  float rs = (entry->sprite->width & (entry->sprite->width-1)) ? 1 : REPEAT_INFINITE;
  float rt = (entry->sprite->height & (entry->sprite->height-1)) ? 1 : REPEAT_INFINITE;
  glSpriteTextureN64(GL_TEXTURE_2D, entry->sprite, &(rdpq_texparms_t){.s.repeats = rs, .t.repeats = rt});
```

Perhaps a bit surprising behavior but at least all textures should show up now.

PR #480, opened 2024-01-06
model64: start model ref count from zero

ghost:
It's later incremented in model64_load_buf anyway.


rasky:
For the sake of maintainability, I'd rather keep the initialization to 1, and remove the two refcount increments. There's a call graph of similar loading functions that are a bit complex to follow, and it seems easier to put 1 at creation, rather than trusting all present and future exit paths to increment the refcount.

rasky:
Committed here:
https://github.com/DragonMinded/libdragon/commit/fa5a0878175a0bca9f5bfa3f999963795ca7112e

Issue #105, opened 2020-06-15
Improve format_mempak repair function


bryc:
libdragon currently overwrites the first 256 bytes of a mempak using Mupen64's template data. Many emulators use unique templates for the initialized data of an emulated .MPK file, and Project64/cen64/mupen64plus borrowed this from the original Mupen64 emulator.

This is bad for a few reasons.
1. Changes more data than is necessary to repair any corrupted data. The original contents of the label and ID area are effectively erased on the user's mempak.
2. Uses the exact same data each time, which can cause issues with a game's ability to detect when a different mempak has been inserted/replaced.

Also, the concept of 'formatting' a mempak (such as preparing it in the factory) is separate from what N64 software actually do through libultra, which is a simpler 'repair' operation, which includes a permanent indicator in the data that a repair had been done. 

I suggest emulating libultra's repair functions to some extent, incorporating a random element to the data.

Here is the general structure of the 32-byte ID block, according to libultra (Using the Mupen64 data as an  anti-example):

![image](https://user-images.githubusercontent.com/1408749/84608906-94951080-ae82-11ea-92da-ddb6c4598443.png)

The only bytes that matter are:
1. `checksum` and `inverted_checksum` must be correct.
2. `banks` must be set to `0x01` to specify the correct mempak size.
3. Bit 0 of `deviceid` must be set
4. `random` is intended to be random (libultra uses **CP0 Register 9, Timer count** for this)
5. `repaired` is written with `FF FF FF FF` to indicate a repair was performed, but has no function.

Everything else is **not checked, used or overwritten in any way**. However `version` is always 0 in over 300 DexDrive and legitimate dumps I have checked.  Here are some examples of libultra repaired data found in real mempaks:

![image](https://user-images.githubusercontent.com/1408749/84608978-e50c6e00-ae82-11ea-8802-6e0d0e9706f7.png)

And how it fits into page 0 (first 256 bytes of the mempak):

![image](https://user-images.githubusercontent.com/1408749/84608923-a4acf000-ae82-11ea-956a-ae3e43b467ec.png)

- Backup priority is 1, 2, 3 until found. When found, backup is overwritten to all other locations.
- `0x81` is written in the first byte of the label area during a repair, but it has no functional purpose.

[...]

PR #481, opened 2024-01-11
Correctly format a ControllerPak

networkfusion:
Still requires formal testing from someone.

For note, the Original PR was: https://github.com/rasky/libdragon/pull/7 which includes lots of extra detail.

Linked to (but does not fix) https://github.com/DragonMinded/libdragon/issues/105


networkfusion:
Doxygen issues seem unrelated.

networkfusion:
@meeq and @bryc are you able to check that this works / looks correct?

bryc:
The code itself seems fine to me. It correctly implements a full Controller Pak initialization.

But does this have a place in libdragon? 

If libdragon were to get a robust C-Pak and repair function, this would largely become obsolete. An argument could also be made that it would be better for a general N64 library to only modify what is _absolutely necessary_ instead of _completely overwriting_ the entire system area.

IMHO this kind of function is more suited for initializing empty C-Pak data in emulators or flashcart C-Pak managers, not homebrew games and such.

networkfusion:
> The code itself seems fine to me. It correctly implements a full Controller Pak initialization.
> 
> But does this have a place in libdragon?
> 
> If libdragon were to get a robust C-Pak and repair function, this would largely become obsolete. An argument could also be made that it would be better for a general N64 library to only modify what is _absolutely necessary_ instead of _completely overwriting_ the entire system area.
> 
> IMHO this kind of function is more suited for initializing empty C-Pak data in emulators or flashcart C-Pak managers, not homebrew games and such.

My take is to fix what is broken, and then deprecate it (since we should be looking at a new API anyway).

And thanks for the response.

Going further, I would suggest that this becomes an internal function that is only called upon in a replacement API when the repair function has failed.

But since the PR is valid with the suggested changes, It seems fine to merge with unstable.

rasky:
Well flashcart cpak manager could be written with libdragon too. I agree that homebrew games shouldn't call this function and that we need to rewrite the whole module. At the same time if this PR works better than what its in now, I think merging it is a good idea anyway

bryc:
This is certainly better than what currently sits at `format_mempak`.  Which merely writes an array of bytes taken from old era Mupen64, under the assumption that the emulator is doing it right (it was not).

Anyway, I'd be happy to assist if and when someone wants to work on a proper complete C-Pak lib 😃. Much of the logic in [MPKEdit](https://github.com/bryc/mpkedit) could be adapted (though it's in JS, it should be possible to port). The only thing MPKEdit lacks currently is a 'smart' `IndexTable` repair routine, which wasn't a priority because the app doesn't interact with live hardware.

Issue #461, opened 2023-10-31
Can't build toolchain on mac os arm M1 apple silicon aarch64


alexvnesta:
Hey - I tried to follow the install guide here, but unfortunately have an error - looks like an issue with the mac os libz? I tried removing `--with-system-zlib` from the gcc configure, but got a whole bunch of errors for `stddef.h`

Here is what I did so far:

https://github.com/DragonMinded/libdragon/wiki/Installing-libdragon

Computer model and OS: 2022 macbook air - Apple M2, 16gb ram, macOS Ventura 13.1

Homebrew version
```
(base) MacBook-Air-2:tools alex$ brew --version
Homebrew 4.1.18-33-g83b4b2f
```

Homebrew installed packages

```
(base) MacBook-Air-2:tools alex$ brew list
==> Formulae
armv7-unknown-linux-gnueabihf	glib				libxcb				python@3.10
autoconf			gmp				libxdmcp			python@3.11
automake			gnu-sed				libxext				python@3.12
bash				gnutls				libxrender			qemu
bdw-gc				graphite2			little-cms2			readline
binutils			grep				lz4				rhash
binwalk				guile				lzip				sevenzip
bison				harfbuzz			lzo				six
ca-certificates			icu4c				m4				snappy
cairo				isl				make				sqlite
capstone			jpeg-turbo			meson				squashfs
cargo-zigbuild			libevent			mips64-elf-binutils		ssdeep
cloudprober			libidn2				mpdecimal			tcl-tk
coreutils			libimagequant			mpfr				texinfo
crosstool-ng			libmpc				ncurses				tree
ddcctl				libnghttp2			nettle				unbound
dtc				libpng				ninja				vde
e2fsprogs			libraqm				numpy				webp
flex				libslirp			openblas			wget
fontconfig			libssh				openjpeg			xorgproto
freetype			libtasn1			openssl@3			xz
fribidi				libtiff				p11-kit				z3
gawk				libtool				p7zip				zig
gcc				libunistring			pcre2				zstd
gdbm				libusb				pillow
gettext				libx11				pixman
giflib				libxau				pkg-config

==> Casks
android-platform-tools	microsoft-auto-update	microsoft-office	paragon-vmdk-mounter	stats
(base) MacBook-Air-2:tools alex$ 
```

I attached the full build log, but here is the bit where I think the error occurred:
```
[...]

Issue #460, opened 2023-10-28
`build-toolchain.sh` should include `isl`


Jan200101:
`build-toolchain.sh` currently sets up the gcc build to include `gmp`, `mpc` and `mpfr` but not `isl` which is also a dependency of gcc.

Jan200101:
related issue #312

rasky:
I thought `isl` is optional?

networkfusion:
@Jan200101 Are you using WSL1 or WSL2 (given you have linked that issue)?!

Jan200101:
> I thought isl is optional?

Doesn't seem to be from looking at the gcc `configure.ac` script.
https://gcc.gnu.org/install/prerequisites.html also doesn't mention it as optional (though it doesn't mention anything being optional, just that they could be).

On a CI runner I ran into `Unable to find a usable isl.  See config.log for details.`

> Are you using WSL1 or WSL2 (given you have linked that issue)?!

Not using WSL at all, this occured on a Fedora based runner.
I linked the issue because it ran into the same error message

rasky:
OK I think the reason I've never realized this is that `isl` is a dependency also of `gmp`. So if you install gmp through a package manager, you get `isl`. And even if you uninstall it to test the gmp source build, there's still `isl` in your system :)

I guess we need to amend the script to also download and copy `isl` then.

rasky:
This has been implemented for Mac for now.

PR #483, opened 2024-01-12
Update build-toolchain.sh for mac os

alexvnesta:
This request will resolve 
https://github.com/DragonMinded/libdragon/issues/461

It also addresses the isl issue - but only for mac os.
https://github.com/DragonMinded/libdragon/issues/460

My proposed changes:

1) Add additional homebrew dependencies for mac os
 - they are specified in the libdragon compile from source guide, but it is more foolproof to add them in the configure script.

2) Require homebrew install of zlib on mac os.
- The system zlib throws a build error.

3) Remove the requirement for system zlib if on mac os.
- I made a platform check that uses system zlib if not on darwin, as I did not want to break other platforms.

I tested this by first removing all homebrew packages:

```
while [[ `brew list | wc -l` -ne 0 ]]; do
    for EACH in `brew list`; do
        brew uninstall --force --ignore-dependencies $EACH
    done
done
```

Then I cloned my modified repo and ran build-toolchain.sh on my m2 pro macbook running Mac OS Sonoma 14.2.1. Libdragon was able to successfully build without issue using these modifications. 

Previously I would get errors not being able to find certain tools (like zlib, gcc, binutils, etc.). These errors are addressed by adding to the brew install -q command.

Please note, that I did not test other platforms. I am not sure if this will break regular mac os intel builds or non-mac os builds.

### DIFF
--- a/tools/build-toolchain.sh
+++ b/tools/build-toolchain.sh
@@ -59,6 +59,10 @@ download () {
     fi
 }
 
+if [[ "$PLATFORM" != 'darwin' ]]; then
+    GCC_CONFIGURE_ARGS+=("--with-system-zlib")
+fi
+
 # Compilation on macOS via homebrew
 if [[ $OSTYPE == 'darwin'* ]]; then
     if ! command_exists brew; then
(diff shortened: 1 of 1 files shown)


alexvnesta:
@rasky I changed to if-else relying on the os-type as you suggested. Hopefully this looks better. Thanks for the suggestion!

rasky:
Pushed also to stable.

PR #487, opened 2024-01-22
Dockerfile improvements

networkfusion:
Minor changes for maintainability.

Note that certain dependencies are probably no longer required on unstable, but aims to be compatible with trunk.

Also noting the fact that we already have a toolchain package available that we could dogfood. That would be another PR though.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,14 +3,28 @@
 
 # Stage 1 - Build the toolchain
 FROM ubuntu:22.04
+
+# Setup paths for the libdragon toolchain
 ARG N64_INST=/n64_toolchain
 ENV N64_INST=${N64_INST}
 
 # install dependencies
 RUN apt-get update
-RUN apt-get install -yq wget bzip2 gcc g++ make file libmpfr-dev libmpc-dev zlib1g-dev texinfo git gcc-multilib
+RUN apt-get install -yq \
+    wget \
+    bzip2 \
+    gcc \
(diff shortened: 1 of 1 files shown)


networkfusion:
@anacierdem can you review?

Dockerfile
```diff
@@ -3,29 +3,55 @@
...
+    gcc \
+    g++ \
+    make \
+    git
+
+# perform cleanup
+RUN apt-get clean \
+    && apt autoremove -yq
+
+# Install toolchain (built in Stage 1)
```
rasky: I think this split in 3 RUN commands will cause image size to increase, because all intermediate steps are saved.

networkfusion: Maybe (a very small one), but it is just like running 3 commands from the console. It is surely worth it for maintainability?!

networkfusion: Addition... you could (successfully) argue that the copy command should be before the agt-get, since it is most likely to fail first. but the split of the commands makes it much easier to find reasons for failues.

networkfusion: Update: Moved the copy command back toward the start.

anacierdem: Especially on the second stage, it is important to combine the commands. o/w they create layers with the temporary files which is a waste. On the first stage, you can split them however you like though, that's the reason we split it into two stages

networkfusion: Ok, done.

PR #488, opened 2024-01-22
Add package dependencies

networkfusion:
(This a copy of my PR from rasky fork which still seems valid)

Adds package dependencies required by the libdragon toolchain.

This "should" fix issues reported here: https://discord.com/channels/205520502922543113/974342113850445874/1074452868469362709

NOTE: only valid for `trunk` as `Unstable` removes the need for `LibPng`!

But unable to test, so would be worth someone trying on a clean install.


rasky:
Not necessary: trunk doesn't use libpng anymore

PR #490, opened 2024-01-24
Improve mputest text

networkfusion:
Improves the text in the mputest and mptest ROMs to use the correct naming as per Nintendo 64 conventions.
Also fixes the menu for mputest as it was seemingly a copy and paste issue.

Note that in future, it would probably be better to merge the 2 ROM's


examples/mptest/mptest.c
```diff
@@ -55,11 +56,11 @@ int main(void)
                         {
                             if (err == -3)
                             {
-                                printf("Mempak is not formatted!");
+                                printf("Pak is not formatted!");
```
rasky, line 59: What's the reason for using `Pak` here rather than `CPak`?

rasky:
You should also rename the tests at this point, they should be called `cptest` / `cputest`.

examples/mptest/mptest.c
networkfusion: IMO it is already assumed by the menu option, although happy to change... let me know.

networkfusion:
> You should also rename the tests at this point, they should be called `cptest` / `cputest`.

I was hesitant to do that as currently, fully compatible with trunk, but can if wanted... let me know.

also... `cputest` would be missleading as... well... suggests it is testing the CPU.

rasky:
Sorry dropped the ball here. I'd still like the test to be renamed as well. Feel free to make a suggestion on a name that makes sense and is not confusing.

networkfusion:
@rasky , I have updated the names to `cpaktest` and `cpak-utest`. I thought the hyphen was necessary as it was confusing enough to have 2 examples with the same name (except one is a test and the other is a unit test), although, as mentioned, they are really too similar to have seperate examples, and should be merged at some point.

rasky:
Thanks! Can you make a similar change to trunk please? That'll help for future merges

PR #493, opened 2024-01-26
Bump RDPQ Buffer

HailToDodongo:
Increase buffer to 64kb to reduce RDP waits, higher values seem to have diminishing returns.
Note that there still seems to be a bug with values way bigger than 64kb in which the performance tanks instead.


snacchus (approved):
Looks good to me

PR #497, opened 2024-02-13
Add GCC snapshot version suppport

jopadan:
Support downloading LATEST-$GCC_V_MAJOR GCC snapshot info and retrieve date version


rasky:
What's the rationale of this? Is there a need to download snapshots? It seems like following released versions has served has pretty well

rasky:
Closed for now, waiting for feedback.

PR #498, opened 2024-02-15
mksprite -f AUTO detecting CI4 format

lambertjamesd:
Determine when an image has a bit depth of less than 4bpp when auto detecting format with a pallete

### DIFF
--- a/tools/mksprite/mksprite.c
+++ b/tools/mksprite/mksprite.c
@@ -326,7 +326,7 @@ bool load_png_image(const char *infn, tex_format_t fmt, image_t *imgout, palette
             else fmt = FMT_IA16;
             break;
         case LCT_PALETTE:
-            fmt = FMT_CI8; // Will check if CI4 (<= 16 colors) later
+            fmt = (state.info_png.color.bitdepth > 4) ? FMT_CI8 : FMT_CI4;
             break;
         case LCT_RGB: case LCT_RGBA:
             // Usage of 32-bit sprites/textures is extremely rare because of the


rasky:
Thanks! Can you describe if there’s a bug to fix or something to improve? Also maybe attaching a PNG that exhibits the modified behavior would help.

lambertjamesd:
I had some 64x64 4bpp textures I wanted to convert using the AUTO format
bush
but it was trying to do 8bpp which didn't fit in tmem using palettes. I noticed that the other formats simply checked the bit depth and doing the same for palettes seemed to work so now the attached image will convert to a CI4 image instead of a CI8 image.

rasky:
I can't reproduce the behavior with that file, it seems like it gets autoselected as CI4 for me:

```
$ ./mksprite -v -v --format AUTO bush.png
Converting: bush.png -> ./bush.sprite [fmt=AUTO tiles=0,0 mipmap=NONE dither=NONE]
loading image: bush.png
loaded bush.png (64x64, LCT_PALETTE)
palette: 15 colors (used: 15)
auto selected format: CI4
TMEM required: 4096 bytes
compressed: ./bush.sprite (2216 -> 1645, ratio 74.2%)
```

I'll look at the code again because I think I remember there was a counterintuitive reason for not doing the CI4 check there. Let me re-study the code a bit more

rasky:
I think the reason why we don't check for bitdepth there is that I've never seen a PNG where bitdepth is not 8 for a LCT_PALETTE file. It seems like most PNG writers don't bother with writing a real 4bpp PNG with a palette (actually packing two indices per byte). Either don't bother, or the standard even doesn't allow that, I don't know.

To auto-select CI4, what we actually do is to check the highest palette index actually used in the image:

```C
    // For a palettized image, copy the palette and also count the number of actually
    // used colors (aka, the highest index used in the image). This is useful later for
    // some heuristics.
    if (state.info_raw.colortype == LCT_PALETTE) {
        memcpy(palout->colors, state.info_png.color.palette, state.info_png.color.palettesize * 4);
        palout->num_colors = state.info_png.color.palettesize;
        palout->used_colors = 0;
        for (int i=0; i < width*height; i++) {
            if (image[i] >= palout->used_colors)
                palout->used_colors = image[i]+1;
        }
        if (flag_verbose)
            fprintf(stderr, "palette: %d colors (used: %d)\n", palout->num_colors, palout->used_colors);
    }
```

Notice that we don't do any palette swapping/reordering/compacting (that is we don't count if there are 16 colors actually used, and then proceeded to compact them and remap indices). mksprite has a strong guarantee that it won't touch your input palette if you provide one, as you might want to do palette tricks at runtime. The only degree of freedom we take is the one explained above: if you chose AUTO, and if the input image uses 16 colors or fewer, and those are within the first 16 colors of the palette, we assume you really wanted a CI4, so we select that format. We *still* write the whole PNG palette in the output sprite file, and you will still have that available at runtime, in case it contained eg different color variants that you want to use yourself.

So for instance if your bush has two colors, green and brown, you can prepare it with two 16 colors palettes (total 32 colors) in the PNG file, and then just export it in the green variant. mksprite will convert it to CI4 but bundle the 32 colors as well. By default, it will display as green of course (after all, it's just what you see when you open the PNG...). But at runtime, sprite_get_palette() will return all the 32 colors and you're free to use them as you wish.

lambertjamesd:
Oh, good to know. I was using a older version on unstable when I did the fix so either it wasn't present or I fixed the issue by rebuilding the palette in GIMP and not my code change. Thanks.

PR #513, opened 2024-04-02
Update workflow actions

networkfusion:
Due to node 16 deprecations.


networkfusion:
@rasky , can you cherrypick this for unstable and opengl, preview branches also?

rasky:
it will be done on next merge. I do that every once in a while.

PR #531, opened 2024-04-28
Fixed an error on gcc version (Ubuntu 13.2.0-23ubuntu4) 13.2.0

lambertjamesd:
I was getting an error while building the toolchain after setting up a computer on the latest version of ubuntu. I'm assuming the error is showing up now because gcc got more sophisticated error checking

I was getting this error at `src/audio/opus/celt_encoder.c:1014`
error: ‘mask’ may be used uninitialized


rasky:
Thanks. The warning is clearly spurious given the following for cycle, but anyway.

PR #541, opened 2024-05-10
toolchain: Update GCC to 14.1

sp1187:
Add -fpermissive to newlib CFLAGS since current upstream does not otherwise work with the stricter GCC 14 defaults.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Libdragon is an open-source SDK for Nintendo 64. It aims for a complete N64
 programming experience while providing programmers with modern approach to
 programming and debugging. These are the main features:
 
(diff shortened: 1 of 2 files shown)


gyrovorbis:
Hey, guys, just a heads up... Is this something you should also be concerned with? You building the C++ standard library? 

https://github.com/KallistiOS/KallistiOS/pull/554

sp1187:
From my testing I found that the TZ database is only linked into the actual ROM when you use the tzdb library directly, just using chrono::system_clock and similar was not enough. So it should be fine for now. As the default C++ version is 17 currently, the developer currently also needs to set the C++ version explicitly to C++20 in order to use tzdb.

gyrovorbis:
Ah, yeah, good points!

PR #545, opened 2024-05-12
Revert "backtrace: Fix alignas order syntax errors with GCC 14"

rasky:
This reverts commit 99997436913a1159ceae663fd0e68046af822b3e.


rasky:
Seems like this commit isn't responsible of the broken build after all

PR #552, opened 2024-05-19
Add RSP DMA control regs to rsp.h

Dragorn421:
I am not sure about the whitespace (indentation, newlines), feel free to tell me.

I think there should be longer comments and additional macros to work with these regs, but that would be a lot to improve with the other regs too. And maybe "look at the wiki" is good enough


Dragorn421:
Reminder for myself to also fix the comment on the spwrintrbrk macro while I'm at it

PR #553, opened 2024-05-20
mpeg: fix macroblock decoding run-on at slice end

thekovic:
Pulls a fix from upstream pl_mpeg that fixes stray macroblocks in video playback.


rasky:
Thanks!

PR #563, opened 2024-06-06
Refactor RTC subsystem [WIP]

meeq:
1. Optimizes away all RTC read operations except the initial sync and recalculates the date/time based on sync point using ticks.
1. Support for "Software RTC with no persistence": if your flashcart doesn't have an RTC, just set the time and we'll count forward from there (until you reset or power-off the console)
1. Support for switching between clock sources:
    * None (Software RTC)
    * Joybus RTC
    * N64DD (Not implemented yet)
    * BBPlayer/iQue (Not implemented yet)
1. Deprecates `rtc_time_t` struct in favor of standard C time types.
1. Adds support for `settimeofday` so standard C APIs can be used to get and set RTC

[N64brew Discord thread](https://discord.com/channels/205520502922543113/1332811445326319646)


include/rtc.h
```diff
@@ -39,12 +122,126 @@ typedef struct rtc_time_t
...
+ * @return whether RTC write persistence appears to be supported
+ */
+bool rtc_is_persistent( void );
+
+__attribute__((deprecated("use rtc_is_persistent instead")))
 bool rtc_is_writable( void );
+
+/**
+ * @brief Read the current date/time from the real-time clock.
+ *
```
gyrovorbis:
```suggestion
 * @deprecated Use rtc_get_time() instead.
 
```
Should you also mark it deprecated for the Doxygen?

include/rtc.h
```diff
@@ -39,12 +122,126 @@ typedef struct rtc_time_t
...
+ *              Destination pointer for the RTC time data structure
+ *
+ * @return whether the rtc_time destination pointer data was modified
+ */
+__attribute__((deprecated("use rtc_get_time instead")))
 bool rtc_get( rtc_time_t * rtc_time );
+
+/**
+ * @brief High-level convenience helper to set the RTC date/time.
+ *
```
gyrovorbis:
```suggestion
 * @deprecated Use rtc_set_time instead.
 
```

gyrovorbis (commented):
I'm the RTC guy for the KOS Sega Dreamcast SDK, and we also are using standard C `time_t` in our RTC API like you are now... I know my opinion doesn't matter even remotely here, but I think this a fantastic API and just wanted to make like 2 teensy comments about marking the API deprecated for the Doxygen so it gets added to the top-level "deprecated" page. lol.

Keep up the good work.

gyrovorbis:
By the way, if you guys ever do anything about "handling the `2038` problem," I'd love to hear it... We have our own `2086` problem over on the Dreamcast. lmao.

meeq:
Rasky informed me that LibDragon is already good on the 2038 problem: `time_t` is 64-bit in LibDragon.

I haven't actually tested to see what our practical limit is on the newlib side. On the RTC side, we're realistically constrained between 1900-2099 (Joybus RTC limit) or 1996-2095 (64DD RTC limit)

examples/rtctest/rtctest.c
```diff
@@ -1,13 +1,17 @@
...
 #include <libdragon.h>
 
 #define BLACK 0x000000FF
 #define WHITE 0xFFFFFFFF
 
 #define JOYSTICK_DEAD_ZONE 32
 
+#define MAX(a,b)  ({ typeof(a) _a = a; typeof(b) _b = b; _a > _b ? _a : _b; })
+#define MIN(a,b)  ({ typeof(a) _a = a; typeof(b) _b = b; _a < _b ? _a : _b; })
+#define CLAMP(x, min, max) (MIN(MAX((x), (min)), (max)))
```
meeq: Are these part of the C standard library yet?

gyrovorbis: NOPE. I think they just got added to C23 for *floats*, but nah, nothing that I know of as a generic macro... I even have these in my own codebase.

meeq:
Should the RTC subsystem be responsible for enforcing min/max dates based on RTC hardware limitations?

* The 1996-2095 range is dictated by the 64DD RTC, which only stores a two-digit year value (and canonically infers XX96-XX99 as 1996-1999).
* Joybus RTC is theoretically limited to the range 1900-2099, but practically limited to 2001-2099 due to emulation quirks in the various implementations.

Behavior | Pro | Con | My Opinion
--- | --- | --- | ---
`settimeofday` fails when setting a timeval that the RTC source clock cannot consistently read-back. RTC source will roll-back if read-back is not consistent with written timeval. | Conceptually simple: "If successful, the value I provided was written to the RTC source as-is. Otherwise, the RTC source will be unchanged." | Requires caller to check result for error. Requires caller to know intimate knowledge of RTC hardware quirks to avoid errors and provide UI affordances. | Non-viable: LibDragon should be shielding developers from the rough-edges of N64 hardware quirks in high-level APIs.
`settimeofday` accepts any timeval, attempts to write it to the RTC source, and keeps counting up from the provided timeval regardless of what was actually written to the RTC. | Easiest to implement for me | Violates the "principle of least-astonishment" | Non-viable: Irresponsibly lazy approach
`settimeofday` accepts any timeval, attempts to write it to the RTC source, and reads the time back from the RTC source, counting forward from the read-back value. | The RTC subsystem time will be consistent with the actual RTC time. | This will result in different behavior across emulators and flashcarts | Potentially viable
`settimeofday` silently clamps timeval (to `{1996,2000,2001}-2095`), attempts to write the clamped value to the RTC source, and keeps counting up from the clamped timeval. | This will generally result in consistent behavior across reasonably-compliant emulators and flashcarts, and keeps RTC subsystem time consistent with actual RTC time | Some users may be surprised that their input timeval is not respected | Potentially viable

Unfortunately, both "potentially viable" options have a huge gaping caveat: EverDrive 3.0/X7 does not support Joybus RTC write commands. If we read-back, the write will always have failed, so the player can't even just set the time for the current play session. If we don't read-back, the write will always silently fail to persist, but at least the time can be set ephemerally.

Animal Forest reads the Joybus RTC continously to update the time, so it would seem to favor the "read-back" approach, and I think this is actually the least-suprising, most-useful of the options.

([Discussion in N64brew Discord thread 💬 ](https://discord.com/channels/205520502922543113/974342113850445874/1249788239217430691))

examples/rtctest/rtctest.c
meeq: LibDragon has them in the "private header" `#include "utils.h"`: https://github.com/DragonMinded/libdragon/blob/8fcaba63faff7c73411606617044c295501fe936/src/utils.h#L12-L14

They seem useful enough that we should expose them to LibDragon devs, but they're generic enough that exposing them in `<libdragon.h>` could cause naming conflicts. (Basically every C code base has these peppered around in some form or another).

I know LibDragon strongly encourages including the entire library as a single header, but I definitely see the benefits for `utils.h` to be exposed as an "optional public header": `#include <libdragon/utils.h>` or something.

This is another one of those "it's a little late to change now" things, but sometimes I wish LibDragon used a common prefix so that we wouldn't have to worry (as much) about namespace collisions with other libraries and user code.

thekovic: A solution could be hiding them behind an opt-in preprocessor define. Like Microsoft does for defining `M_PI` and friends in `math.h`. I think you always get them with GCC but with Microsoft's implementation, you need `#define _USE_MATH_DEFINES` to have them defined.

include/rtc.h
```diff
@@ -6,124 +6,83 @@
...
+ * If no hardware real-time clock is available, the subsystem will provide
+ * a software-based real-time clock for the current play session. Note that
+ * this software RTC does not persist across resets or power cycles!
+ *
+ * Once the RTC subsystem is initialized, you can use ISO C Time functions
  * to get the current time, for example: `time(NULL)` will return the number of
  * seconds elapsed since the UNIX epoch (January 1, 1970 at 00:00:00).
- * To check if the real-time clock supports writes, call #rtc_is_writable.
- * To write a new time to the real-time clock, call #rtc_set.
+ * To check if the real-time clock supports writes, call #rtc_is_persistent.
```
rasky, line 33: Do you think `rtc_is_persistent` is actually better than `rtc_is_writable`? It would seem to be that "persistent" would actually mean "the date time will be persisted across reboots", which is an alias for "there is a RTC available"; the software RTC is "non persistent".

The previous name `rtc_is_writable` made more sense to me, but what's your position on this?

include/rtc.h
```diff
@@ -6,124 +6,83 @@
...
- * reverse-engineering and documenting the inner-workings of the Joybus RTC.
+ * format into a standard `struct tm` structure. You can use convert between
+ * `struct tm` and `time_t` using the standard C library functions `gmtime`,
+ * `mktime`, and `time`.
+ *
+ * In anticipation of support for the 64DD and iQue real-time clocks, the
+ * subsystem has APIs that allow homebrew to specify the preferred RTC source.
+ * Use #rtc_is_source_available to determine if a specific RTC source can be
+ * used, and #rtc_set_source to switch between available sources. Make sure
+ * to call #rtc_resync_time after switching sources!
```
rasky: Any reason why this is not automatic when you switch sources?

include/rtc.h
```diff
@@ -6,124 +6,83 @@
...
+ *
+ * Internally, Joybus RTC cannot represent dates before 1990-01-01, although
+ * some RTC implementations (like UltraPIF) only support dates after
+ * 2000-01-01.
+ *
+ * 64DD RTC only stores two digits for the year, so conventionally 96-99 are
+ * treated as 1996-1999 and 00-95 are treated as 2000-2095.
+ *
+ * For consistency, the RTC subsystem only supports dates between 1996-2095.
+ * Attempting to set the clock beyond this range will fail.
```
rasky: I'll add an advise here to only store the "real" date in the RTC. If the game supports time traveling it is better to handle that with some internal offset, rather than storing the wrong date into the RTC.

include/system.h
```diff
@@ -263,28 +264,39 @@ typedef struct
...
+     */
+    bool (*settime)( time_t );
+} time_hooks_t;
+
+/**
+ * @brief Register a filesystem with newlib
  *
  * @param[in] prefix
- *            Prefix of the filesystem
+ *            The prefix that was used to register the filesystem
```
rasky: "will be used"?

include/system.h
```diff
@@ -263,28 +264,39 @@ typedef struct
...
- *            If the registered filesystem doesn't support an operation, it
- *            should leave the callback null.
- * 
- * @retval -1 if the parameters are invalid
- * @retval -2 if the prefix is already in use
- * @retval -3 if there are no more slots for filesystems
- * @retval 0 if the filesystem was registered successfully
+ *            Pointer to structure containing filesystem callbacks
+ *
+ * @return 0 on successful registration or a negative value on failure.
```
rasky: Why dropping the description of the error codes here?

include/rtc.h
```diff
@@ -134,116 +93,121 @@ extern "C" {
...
- * a few milliseconds to complete.
- *
- * @param[out]  rtc_time
- *              Destination pointer for the RTC time data structure
- *
- * @return whether the rtc_time destination pointer data was modified
+ * @return the current RTC time as a UNIX timestamp
  */
-bool rtc_get( rtc_time_t * rtc_time );
+time_t rtc_get_time( void );
```
rasky: We might want to move these to some internal header, as they're only useful as internal implementation. For API, time() is sufficient.

rasky (approved):
Great work!

include/rtc.h
meeq: I'm sure I had a reason, but that rationale has been lost to time.

I agree upon looking at this with new eyes: by the principle of least-astonishment, I would expect switching the clock source to resync the time.

👍🏻

include/rtc.h
meeq: The distinction here is that even if:
1. The RTC implementation is not writable (e.g. emulators, ED64)
2. We are using a software RTC because no hardware RTC exists

...the RTC subsystem will still let you set the time.

My concern with `rtc_is_writable` is that it _looks like_ you're writing to the RTC when you change the time, but it doesn't actually persist.

I'd love others' opinions on the naming here.

include/rtc.h
meeq: Done https://github.com/DragonMinded/libdragon/pull/563/commits/9c0b45291f0dee239eb48ae485da21b901d76fe6

include/system.h
meeq: Not sure how that happened; reverted https://github.com/DragonMinded/libdragon/pull/563/commits/97ce533d4f0a9df128de2856be7514d3ce95409c

include/system.h
meeq: Fixed. https://github.com/DragonMinded/libdragon/pull/563/commits/97ce533d4f0a9df128de2856be7514d3ce95409c

PR #568, opened 2024-06-11
Setup handler callbacks for most BB interrupts

Thar0:
Some interrupt causes are still unhandled since it's still unclear to me how they should be acknowledged.

### DIFF
--- a/include/interrupt.h
+++ b/include/interrupt.h
@@ -178,6 +178,69 @@ void register_CART_handler( void (*callback)() );
  */
 void register_RESET_handler( void (*callback)() );
 
+/**
+ * @brief Register a BBPlayer FLASH callback
+ *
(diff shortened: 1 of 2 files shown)


rasky:
Should we maybe keep the “BB” prefix convention for these symbols that are mixed with N64 ones? For instance `register_BB_FLASH_handler`

Thar0:
Added prefixes

rasky:
You need a rebase here; this conflicts with another PR we merged where we moved all docs to header files

Thar0:
I'm not sure there is a conflict? My branch was based on dc3b17f9ccf6058d3a516685bd31eaf44d399b86 which already incorporates ff7420adf2d92b51d42867e2e625d0accbd07901

PR #569, opened 2024-06-19
Create devcontainer.json

niemasd:
Adds support for VS Code / GitHub Codespaces environment using the `Dockerfile` in the repo

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,5 @@
+{
+  "build": {
+    "dockerfile": "../Dockerfile"
+  }
+}


anacierdem:
There is a more elaborate setup [here](https://github.com/anacierdem/libdragon-docker), do you still see value having this in the main repository?

niemasd:
Good question! I added these here for easier development of `libdragon` itself, rather than development of tools that *use* `libdragon`. But perhaps that's not an issue?

anacierdem:
That one can also be used to develop libdragon itself, I mean the library. Developing the `Dockerfile` is a little more cumbersome but a devcontainer would not help much with that either. Is there a use case in your mind for this specific workflow? Maybe if you can describe how to use this, I can understand it better?

niemasd:
For folks who want to contribute to the `libdragon` library itself, they will first fork this repo and then develop (and ultimately open a PR). If they develop using GitHub Codespaces or VS Code, having `devcontainer` set up like this will mean that the dependencies will be automatically set up right when they click the "Create codespace" button or right when they clone within VS Code. If there is no `devcontainer` setup, my understanding is that they would need to manually install the dependencies in their development environment or would need to manually point to the other repo: to the best of my knowledge, a simple "open in Codespaces / clone in VS Code and just start developing" workflow isn't possible in the current setup (but I could be mistaken)

anacierdem:
> a simple "open in Codespaces / clone in VS Code and just start developing" workflow isn't possible in the current setup

Unfortunately this won't be possible with your proposed `devcontainer` setup either. It is a very barebones file you have in this PR and it will build the toolchain whenever you fire it up, which will take a long time considering you only want a quick development setup. OTOH, the [other](https://github.com/anacierdem/libdragon-docker) setup is more "ready to use" for such a use case. You can clone that repo and it will also clone this one as a submodule where you can open a PR targeting here if you like all within a dev container.  If you really want a similar setup here on this repository, I suggest:

- Go check it out, play with it and see how it [works](https://github.com/anacierdem/libdragon-docker).
- Create a similar setup here and make sure it is working in a useful manner.
- Document it in the readme of this repository.

Considering all of the above exists on the other repository, this might be a little unnecessary. Still, if you want to work on it feel free to update this PR when you feel it's ready.

niemasd:
Thanks for the additional context! I wasn't aware of the other one when I made this PR, but I'll play around with it! I'll close this PR, and if I revisit this and expand, I'll reopen it with updates

PR #571, opened 2024-06-20
Cleanup some nits (gcc -Wextra)

Dragorn421:


src/display.c
```diff
@@ -96,7 +96,6 @@ void display_init( resolution_t res, bitdepth_t bit, uint32_t num_buffers, gamma
     /* Can't have the video interrupt happening here */
     disable_interrupts();
 
-    /* Minimum is two buffers. */
```
thekovic, line 99: I think updating the comment is preferable to removing it. While code obviously takes precedence over comments in terms of information value, a comment as a snapshot in time serves as a possible reminder of any assumptions, motivations, or limitations that were held at the time the comment was written. Someone looking at this code 5 years from now might struggle to imagine why the code is doing what it's doing if there is no comment attached.

src/backtrace.c
```diff
@@ -205,9 +206,9 @@ static symtable_header_t symt_open(void) {
  * @param idx       Index of the entry to return
  * @return addrtable_entry_t  Entry of the address table
  */
-static addrtable_entry_t symt_addrtab_entry(symtable_header_t *symt, int idx)
+static addrtable_entry_t symt_addrtab_entry(symtable_header_t *symt, unsigned int idx)
 {
-    assert(idx >= 0 && idx < symt->addrtab_size);
+    assert(idx < symt->addrtab_size);
```
rasky, line 211: Is there a specific rationale for this int/unsigned changes? In general I prefer to use signed numbers because they have less suprising behaviors in case of bugs, I've had quite a few issues with unsigned having surprising behaviors.

src/backtrace.c
```diff
@@ -225,22 +226,25 @@ static addrtable_entry_t symt_addrtab_entry(symtable_header_t *symt, int idx)
...
     }
     addrtable_entry_t entry = symt_addrtab_entry(symt, min);
-    if (min > 0 && ADDRENTRY_ADDR(entry) > addr)
-        entry = symt_addrtab_entry(symt, --min);
-    if (idx) *idx = min;
+    if (min > 0 && ADDRENTRY_ADDR(entry) > addr) {
+        min -= 1;
+        entry = symt_addrtab_entry(symt, min);
+    }
+    if (idx != NULL) *idx = min;
```
rasky, line 247: Also what's the rationale of this? Is there some GCC -Wextra warning on this?

src/backtrace.c
```diff
@@ -509,7 +514,7 @@ static void backtrace_foreach(void (*cb)(void *arg, void *ptr), void *arg)
                         return;
                     }
                 }
-                // FALLTHROUGH!
+                __attribute__((fallthrough));
```
rasky, line 517: I thought this was matched by GCC, but I guess if you changed it, it wasn't?

I think it should match this regex:

```
[ \t.!]*(ELSE,? |INTENTIONAL(LY)? )?
FALL(S | |-)?THR(OUGH|U)[ \t.!]*(-[^\n\r]*)?```

which is one of the various used by `-Wimplicit-fallthrough`

File: src/graphics.c [hunk 5]
```diff
@@ -470,7 +470,7 @@ void graphics_draw_character( surface_t* disp, int x, int y, char ch )
 
         for( int yp = sy; yp < ey; yp++ )
         {
-            const register int run = yp * sprite_font.sprite->width;
+            register const int run = yp * sprite_font.sprite->width;
```
rasky, line 473: Since you're touching this, just drop `register`

rasky (commented):
Thanks! If you want, I can take care of the requested changes, just let me know

PR #573, opened 2024-06-30
Updated Liver.ttf to bugfixed version, and added Liver_mono.ttf

8bitprodigy:


rasky:
I've manually pushed the latest version of Liver.ttf from this PR, but I'm not going to add the monospace variant. I keep only one variant per font in the gallery.

I suggest you setup a small webpage with the font (or on some aggregator like itch.io), and I can then link it from the gallery.

PR #579, opened 2024-07-10
Add a flag --ares-tpak to enable N64 Transfer Pak compatibility for the Ares Emulator

risingPhil:
Add a flag --ares-tpak to enable N64 Transfer Pak compatibility for the Ares Emulator. This flag is useful for debugging transfer pak functionality in Ares Emulator.

This uses the Advanced Homebrew Header controller 1 config field. When Ares Emulator reads this field in <ares-repo>/mia/nintendo-64.cpp, it will ask you to select the desired Gameboy rom to use.

Note: the gameboy SRAM sav file needs to be named <filename_of_the_rom_without_extension>.ram to be loaded as well.

PR #580, opened 2024-07-10
Implement --controller<x> options in ed64romconfig to change the respective Advanced Homebrew Header fields

risingPhil:
Implement --controller<x> options in ed64romconfig to change the respective Advanced Homebrew Header fields

Modifying these fields may influence emulator behaviour, such as Ares'

The context in which I implemented this is to make debugging N64 Transfer Pak code easier with Ares Emulator: 
After using the "--controller1 n64,pak=transfer" option on your rom, Ares will ask for a gameboy rom for the emulated transfer pak everytime you load the rom. 

(Note: In Ares, the gameboy save file that will be loaded is <gb_rom_filename_without_extension>.ram)

This pull request replaces https://github.com/DragonMinded/libdragon/pull/579


tools/ed64romconfig.c
```diff
@@ -51,9 +67,23 @@ int print_usage(const char * prog_name)
...
+	fprintf(stderr, "\t-c, --rtc                       Declare real-time clock support.\n");
+	fprintf(stderr, "\t-r, --regionfree                Declare region-free ROM.\n");
+	fprintf(stderr, "\t-C1, --controller1 <type>       Define controller 1 hardware type. <type> should be one of:\n");
+	fprintf(stderr, "\t    default                     Emulators should follow standard config for this port.\n");
+	fprintf(stderr, "\t    n64,pak=rumble              N64 controller with Rumble Pak\n");
+	fprintf(stderr, "\t    n64,pak=controller          N64 controller with Controller Pak\n");
+	fprintf(stderr, "\t    n64,pak=transfer            N64 controller with Transfer Pak\n");
+	fprintf(stderr, "\t    nothing                     Nothing attached to this port\n");
+	fprintf(stderr, "\t    mouse                       N64 mouse\n");
+	fprintf(stderr, "\t    vru                         VRU\n");
```
meeq, line 81: Tangential: Does Ares make a distinction between VRU/VRS?

tools/ed64romconfig.c
```diff
@@ -51,9 +67,23 @@ int print_usage(const char * prog_name)
...
-	fprintf(stderr, "\t-r, --regionfree       Declare region-free ROM.\n");
+	fprintf(stderr, "\t-w, --savetype <type>           Declare cartridge save type.\n");
+	fprintf(stderr, "\t-c, --rtc                       Declare real-time clock support.\n");
+	fprintf(stderr, "\t-r, --regionfree                Declare region-free ROM.\n");
+	fprintf(stderr, "\t-C1, --controller1 <type>       Define controller 1 hardware type. <type> should be one of:\n");
+	fprintf(stderr, "\t    default                     Emulators should follow standard config for this port.\n");
+	fprintf(stderr, "\t    n64,pak=rumble              N64 controller with Rumble Pak\n");
+	fprintf(stderr, "\t    n64,pak=controller          N64 controller with Controller Pak\n");
+	fprintf(stderr, "\t    n64,pak=transfer            N64 controller with Transfer Pak\n");
+	fprintf(stderr, "\t    nothing                     Nothing attached to this port\n");
```
meeq: I can think of a couple other aliases that I would try before "nothing" if I wasn't sure of the syntax:
* none
* empty
* disconnected

Consistency with save_type would suggest that "none" is the preferable option here.

tools/ed64romconfig.c
```diff
@@ -51,9 +67,23 @@ int print_usage(const char * prog_name)
...
+	fprintf(stderr, "\t-r, --regionfree                Declare region-free ROM.\n");
+	fprintf(stderr, "\t-1, --controller1 <type>        Define controller 1 hardware type. <type> should be one of:\n");
+	fprintf(stderr, "\t    default                     Emulators should follow standard config for this port.\n");
+	fprintf(stderr, "\t    n64,pak=rumble              N64 controller with Rumble Pak\n");
+	fprintf(stderr, "\t    n64,pak=controller          N64 controller with Controller Pak\n");
+	fprintf(stderr, "\t    n64,pak=transfer            N64 controller with Transfer Pak\n");
+	fprintf(stderr, "\t    nothing                     Nothing attached to this port\n");
+	fprintf(stderr, "\t    mouse                       N64 mouse\n");
+	fprintf(stderr, "\t    vru                         VRU\n");
+	fprintf(stderr, "\t    ngc                         Gamecube controller\n");
```
meeq: `ngc` is less clear than `gamecube`

tools/ed64romconfig.c
risingPhil: No. As a matter of fact it appears to be currently unimplemented.

<ares-repo>/mia/medium/nintendo-64.cpp (line 740) says "not supported yet" for all options starting at the N64 mouse. In practice, that means only the "pak" and default options are actually supported by Ares for the moment.

Btw, I was aware of this going in. But not being supported right now doesn't mean they never will. I just added all the valid options according to https://n64brew.dev/wiki/ROM_Header#Advanced_Homebrew_ROM_Header

tools/ed64romconfig.c
```diff
@@ -51,9 +67,23 @@ int print_usage(const char * prog_name)
...
 	fprintf(stderr, "Command-line flags:\n");
-	fprintf(stderr, "\t-w, --savetype <type>  Declare cartridge save type.\n");
-	fprintf(stderr, "\t-c, --rtc              Declare real-time clock support.\n");
-	fprintf(stderr, "\t-r, --regionfree       Declare region-free ROM.\n");
+	fprintf(stderr, "\t-w, --savetype <type>           Declare cartridge save type.\n");
+	fprintf(stderr, "\t-c, --rtc                       Declare real-time clock support.\n");
+	fprintf(stderr, "\t-r, --regionfree                Declare region-free ROM.\n");
+	fprintf(stderr, "\t-1, --controller1 <type>        Define controller 1 hardware type. <type> should be one of:\n");
+	fprintf(stderr, "\t    default                     Emulators should follow standard config for this port.\n");
+	fprintf(stderr, "\t    n64,pak=rumble              N64 controller with Rumble Pak\n");
```
meeq, line 76: Should probably have a `n64` (no `,pak=`) option for completeness

tools/ed64romconfig.c
```diff
@@ -51,9 +67,23 @@ int print_usage(const char * prog_name)
...
 	fprintf(stderr, "\n");
 	fprintf(stderr, "Command-line flags:\n");
-	fprintf(stderr, "\t-w, --savetype <type>  Declare cartridge save type.\n");
-	fprintf(stderr, "\t-c, --rtc              Declare real-time clock support.\n");
-	fprintf(stderr, "\t-r, --regionfree       Declare region-free ROM.\n");
+	fprintf(stderr, "\t-w, --savetype <type>           Declare cartridge save type.\n");
+	fprintf(stderr, "\t-c, --rtc                       Declare real-time clock support.\n");
+	fprintf(stderr, "\t-r, --regionfree                Declare region-free ROM.\n");
+	fprintf(stderr, "\t-1, --controller1 <type>        Define controller 1 hardware type. <type> should be one of:\n");
+	fprintf(stderr, "\t    default                     Emulators should follow standard config for this port.\n");
```
meeq: Why would you ever explicitly set default instead of just omitting the parameter?

risingPhil: I suppose that's true. I'll remove it. ;-)

n64.mk
```diff
@@ -8,6 +8,7 @@ N64_ROM_RTC = # Set to true to enable the Joybus Real-Time Clock
 N64_ROM_REGIONFREE = # Set to true to allow booting on any console region
 N64_ROM_REGION = # Set to a region code (emulators will boot on a specific console region)
 N64_ROM_ELFCOMPRESS = 1 # Set compression level of ELF file in ROM
+N64_ROM_CONTROLLER_TYPE1 = # Sets controller type in the Advanced Homebrew Header. This influences emulator behaviour such as Ares'
```
meeq: Suggestion: rename to `N64_ROM_CONTROLLER1`

Also, consider adding 2, 3, 4

tools/ed64romconfig.c
risingPhil: Hmm... Based on https://n64brew.dev/wiki/ROM_Header#Advanced_Homebrew_ROM_Header, It seems that that might be the "default" option.

...which brings us back to your previous comment. So up to you: should I rename the "default" argument to "n64"? Or just omit it entirely?

tools/ed64romconfig.c
risingPhil: Okay. Makes sense. I just followed the name from the n64brew page.

But I'll change it.

tools/ed64romconfig.c
risingPhil: Okay, I'll change it.

tools/ed64romconfig.c
meeq: I assume an emulator could make the distinction by the ROM's region. That would prevent Japanese games from being able to use VRU, and English games from using VRS, but that's a pretty extreme edge case that I don't see being a practical limitation.

n64.mk
risingPhil: Okay :-)

tools/ed64romconfig.c
```diff
@@ -170,6 +309,14 @@ int main(int argc, char *argv[])
 	fwrite(cart_id, 1, CART_ID_SIZE, write_file);
 	fseek(write_file, VERSION_OFFSET, SEEK_SET);
 	fwrite(&config, 1, VERSION_SIZE, write_file);
+	fseek(write_file, CONTROLLERTYPE1_OFFSET, SEEK_SET);
+	fwrite(&controller_type1, 1, 1, write_file);
```
meeq: ["When in Rome, do as the Romans do"](https://tommcfarlin.com/coding-standards-and-guidelines/)

A `CONTROLLERTYPE_SIZE` constant is not strictly necessary, but I do think it is "better".

tools/ed64romconfig.c
meeq: I'd say rename "default" to "n64".

tools/ed64romconfig.c
risingPhil: Okay, I'll change it.

tools/ed64romconfig.c
thekovic: I'll just add that official styling is "GameCube" (capital C) so the printed text should follow that.

tools/ed64romconfig.c
risingPhil: closing because of comment below

tools/ed64romconfig.c
risingPhil: done

tools/ed64romconfig.c
risingPhil: done both things now ;-)

n64.mk
risingPhil: done

tools/ed64romconfig.c
risingPhil: done

n64.mk
```diff
@@ -62,6 +66,7 @@ N64_TOOLFLAGS += $(if $(N64_ROM_REGION),--region $(N64_ROM_REGION))
 N64_ED64ROMCONFIGFLAGS =  $(if $(N64_ROM_SAVETYPE),--savetype $(N64_ROM_SAVETYPE))
 N64_ED64ROMCONFIGFLAGS += $(if $(N64_ROM_RTC),--rtc) 
 N64_ED64ROMCONFIGFLAGS += $(if $(N64_ROM_REGIONFREE),--regionfree)
+N64_ED64ROMCONFIGFLAGS += $(if $(N64_ROM_CONTROLLER_TYPE1),--controller1 $(N64_ROM_CONTROLLER_TYPE1))
```
meeq, line 69: Missing 2, 3, 4

risingPhil: shoot. Missed those.

I've added them now. Good catch!

PR #587, opened 2024-08-16
Allow input from all 4 controllers on the error screen.

FazanaJ:
Revolutionise the error screen experience with multiple controllers and allow all the family to join in on the agonising fun with finding where they went wrong.


rasky:
I think the internal API to poll controllers from inspector is showing its limit. Nonetheless, this can go in as-is meanwhile.

PR #588, opened 2024-08-17
Speed up DFS path lookups

gamemasterplc:
Will use hash lookups by default for paths unless if 2 paths have the same hash where it will try to look up by the path.


include/dfsinternal.h
```diff
@@ -52,6 +55,30 @@ typedef struct dfs_open_file_s
...
+typedef struct dfs_file_entry_s {
+    /** @brief Hash of file path */
+    uint32_t path_hash;
+    /** @brief Top 12 bits: length; lowest 20 bits: offset */
+    uint32_t path_ofs;
+    /** @brief Data offset for file */
+    uint32_t data_ofs;
+    /** @brief Data length for file */
+    uint32_t data_len;
+} dfs_file_entry_t;
```
rasky: I suggest to rename this to something that associates it with the lookup table.

Maybe call this `dfs_lookup_file_t` and the other `dfs_lookup_t`

include/dfsinternal.h
```diff
@@ -52,6 +55,30 @@ typedef struct dfs_open_file_s
...
+
+/** @brief Data for DFS file lookup */
```
rasky: In general, please try to elaborate a bit more in comments. For instance, here explains what is this "lookup" data structure, what it contains and why it's used.

include/dfsinternal.h
```diff
@@ -52,6 +55,30 @@ typedef struct dfs_open_file_s
...
+typedef struct dfs_file_lookup_s {
+    /** @brief Number of files */
+    uint32_t num_files;
+    /** @brief Size of path buffer */
```
rasky: Same for this, a reader has no clue what a "path buffer" is.

src/dragonfs.c
```diff
@@ -555,13 +572,15 @@ static int __dfs_init(uint32_t base_fs_loc)
...
+    if(id_node.flags == ROOT_FLAGS && !strcmp(id_node.path, ROOT_PATH))
     {
         /* Passes, set up the FS */
         base_ptr = base_fs_loc;
         clear_directory();
-
+        if(id_node.file_pointer == 0) {
+            return DFS_EBADFS;
+        }
+        init_dfs_lookup(&id_node);
```
rasky: Can you maybe move this check into init_dfs_lookup and make it return a bool, at least it's clear that the check is related to the lookup.

src/dragonfs.c
```diff
@@ -684,36 +703,93 @@ int dfs_dir_findnext(char *buf)
...
+    uint32_t hash = 0;
+    while(*str) {
+        char c = *str++;
+        hash = (hash * prime) + c;
+    }
+    return hash;
+}
 
-int dfs_open(const char * const path)
+static int32_t lookup_file_hash_entry(uint32_t hash)
```
rasky: Add a comment to this function to explain what it does, specifically the fact that in case of conflicts it will return the first index

src/dragonfs.c
```diff
@@ -684,36 +703,93 @@ int dfs_dir_findnext(char *buf)
...
-        /* File not found, or other error */
-        return ret;
+static dfs_file_entry_t *lookup_file_path_entry(const char *const path, int32_t start_index, uint32_t hash)
+{
+    size_t buf_len = ROUND_UP(file_lookup->path_buf_size, 16);
+    char alignas(16) path_buf[buf_len];
+    uint32_t src_path_len = strlen(path)+1;
+    if(src_path_len % 2 != 0) {
+        src_path_len++;
+    }
```
rasky, line 733: This shouldn't be required, dma_read handles odd sizes too.

src/dragonfs.c
```diff
@@ -684,36 +703,93 @@ int dfs_dir_findnext(char *buf)
...
-    /* Try to find a free slot */
-    dfs_open_file_t *file = malloc(sizeof(dfs_open_file_t));
+static dfs_file_entry_t *lookup_file_entry(const char *const path)
+{
+    uint32_t hash = prime_hash(path, DFS_LOOKUP_PRIME);
+    int32_t index = lookup_file_hash_entry(hash);
+    if(index == -1) {
+        return NULL;
+    }
+    return lookup_file_path_entry(path, index, hash);
```
rasky: I found a bit confusing have 3 functions with similar names. They seem to lookup different things, while instead it is a single algorithm split into one main functions and two sub-functions.

I would suggest maybe to inline everything into lookup_file_entry. I don't think the subfunctions help much with readability. It's not that big anyway, and it might make sense to read it linearly.

tools/mkdfs/mkdfs.c
```diff
@@ -257,6 +291,96 @@ uint32_t add_directory(const char * const path)
...
+    //Sort in hash order
+    dfs_file_t *entry1 = (dfs_file_t *)a;
+    dfs_file_t *entry2 = (dfs_file_t *)b;
+    if(entry1->path_hash > entry2->path_hash) {
+        return 1;
+    } else if(entry1->path_hash < entry2->path_hash) {
+        return -1;
+    } else {
+        return 0;
+    }
```
rasky, line 305: It can be simplified to `entry2->path_hash - entry1->path_hash`

src/dragonfs.c
```diff
@@ -684,36 +703,93 @@ int dfs_dir_findnext(char *buf)
...
+}
 
-    if(ret != DFS_ESUCCESS)
-    {
-        /* File not found, or other error */
-        return ret;
+static dfs_file_entry_t *lookup_file_path_entry(const char *const path, int32_t start_index, uint32_t hash)
+{
+    size_t buf_len = ROUND_UP(file_lookup->path_buf_size, 16);
+    char alignas(16) path_buf[buf_len];
```
rasky: It seems that there is quite a bit of code to calculate this `path_buf_size` in mkdfs (it is the maximum path length as far as I can see), and that's just to make sure here to declare a stack buffer big enough.

Can't you just declare this buffer later, in the for cycle, just before the dma call? There, you can declare it exactly of `path_len` size, which is the length you need to read it. At each iteration of the loop, the length will change, but the compiler will allocate and release the stack allocation dynamically for you.

So I don't think that calculating this maximum size and keeping it stored in the lookup table is necessary.

tools/mkdfs/mkdfs.c
```diff
@@ -257,6 +291,96 @@ uint32_t add_directory(const char * const path)
...
+    rom_lookup->path_ofs = SWAPLONG(path_ofs);
+    
+    uint32_t path_curr_ofs = 0;
+    
+    for(uint32_t i=0; i<num_files; i++) {
+        uint32_t len = strlen(dfs_files[i].path)+1;
+        if(len % 2 != 0) {
+            len++;
+        }
+        rom_lookup->files[i].path_ofs = SWAPLONG((len << 20)|path_curr_ofs);
```
rasky, line 341: I suggested above that maybe you don't need to roundup the length here. But anyway, I wouldn't store the rounded up length in the lookup table. It is an implementation detail that you need the length to be rounded up. I would keep the data structure clean of this detail.

tools/mkdfs/mkdfs.c
```diff
@@ -288,7 +412,8 @@ int main(int argc, char *argv[])
 
         return -1;
     }
-
+    sort_dfs_entries();
+    write_dfs_entries();
```
rasky: I would use the name "dfs lookup" here too. So maybe just do one function write_dfs_lookup() (which will also take care of sorting)

rasky (commented):
Seems good. The speedup is fantastic, great work!

Comments are mostly about a bit more docs and naming issues.

include/dfsinternal.h
gamemasterplc: This member is no longer present here

tools/mkdfs/mkdfs.c
gamemasterplc: Was not able to simplify it into a subtraction of either order without breaking ordering.

src/dragonfs.c
gamemasterplc: Function no longer exists.

tools/mkdfs/mkdfs.c
rasky: Right, on second thought this would work for signed integers, not unsigned. So I guess the explicit ifs are ok.

src/dragonfs.c
gamemasterplc: Code updated since stored path_len is true path_len.

tools/mkdfs/mkdfs.c
gamemasterplc: Spacing between strings is rounded up now but not the actual length of each string to prevent DMA phase issues.

include/dfsinternal.h
gamemasterplc: Have given basic purpose of dfs_lookup_t.

PR #595, opened 2024-08-25
debug: added support for SummerCart64 AUX register

Polprzewodnikowy:


src/n64sys.c
```diff
@@ -175,6 +180,41 @@ void wait_ms( unsigned long wait_ms )
...
+    while( *SP_STATUS & (SP_STATUS_IO_BUSY | SP_STATUS_DMA_BUSY) );
+
+    // Flush the RDP and wait for DMA to finish fetching commands
+    *DP_STATUS = DP_WSTATUS_SET_FLUSH;
+    RSP_WAIT_LOOP(200) {
+        if ( !(*DP_STATUS & DP_STATUS_DMA_BUSY) ) {
+            break;
+        }
+    }
+    *DP_STATUS = DP_WSTATUS_RESET_FLUSH;
```
Polprzewodnikowy, line 207: Currently this code does not safely shutdown the RDP - there are occasional cases of RDP hard lock that prevents console from using it until full power cycle, thus limiting the soft reboot functionality severely.

PR #606, opened 2024-09-12
Support pifiles in wav64 and opus

nopjne:
Allow wav64_open to work with "pi:/" based files.

### DIFF
--- a/include/wav64.h
+++ b/include/wav64.h
@@ -32,9 +32,9 @@ typedef struct {
 	 */
 	waveform_t wave;
 
-	/** @brief Absolute ROM address of WAV64 */
-	uint32_t rom_addr;
-
+	/** @brief File descriptor to read WAV64 */
+	int current_fd;			 ///< File descriptor for the wav64 file
+	int base_offset;		 ///< Start of Wav64 data.
 	int format;			     ///< Internal format of the file
 	void *ext;               ///< Pointer to extended data (internal use)
 } wav64_t;
--- a/src/audio/wav64.c
+++ b/src/audio/wav64.c
@@ -16,6 +16,7 @@
 #include "debug.h"
 #include "utils.h"
 #include "rspq.h"
+#include "asset_internal.h"
 #include <stdbool.h>
 #include <string.h>
 #include <assert.h>
@@ -23,6 +24,8 @@
 #include <errno.h>
 #include <limits.h>
 #include <stdalign.h>
+#include <fcntl.h>
+#include <unistd.h>
 
 /** @brief Set to 1 to use the reference C decode for VADPCM */
 #define VADPCM_REFERENCE_DECODER     0
@@ -148,7 +151,16 @@ static inline void rsp_vadpcm_decompress(void *input, int16_t *output, bool ster
 
 #endif /* VADPCM_REFERENCE_DECODER */
 
-void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wlen, int bps) {
+void raw_waveform_read(samplebuffer_t *sbuf, int current_fd, int wpos, int wlen, int bps) {
+	uint8_t* ram_addr = (uint8_t*)samplebuffer_append(sbuf, wlen);
+	int bytes = wlen << bps;
+
+	uint32_t t0 = TICKS_READ();
+	read(current_fd, ram_addr, bytes);
+	__wav64_profile_dma += TICKS_READ() - t0;
+}
+
+void raw_waveform_read_address(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wlen, int bps) {
 	uint32_t rom_addr = base_rom_addr + (wpos << bps);
 	uint8_t* ram_addr = (uint8_t*)samplebuffer_append(sbuf, wlen);
 	int bytes = wlen << bps;
@@ -165,7 +177,10 @@ void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wl
 static void waveform_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wlen, bool seeking) {
 	wav64_t *wav = (wav64_t*)ctx;
 	int bps = (wav->wave.bits == 8 ? 0 : 1) + (wav->wave.channels == 2 ? 1 : 0);
-	raw_waveform_read(sbuf, wav->rom_addr, wpos, wlen, bps);
+	if (seeking) {
+		lseek(wav->current_fd, wav->base_offset + (wpos << bps), SEEK_SET);
+	}
+	raw_waveform_read(sbuf, wav->current_fd, wpos, wlen, bps);
 }
 
 static void waveform_vadpcm_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wlen, bool seeking) {
@@ -175,12 +190,12 @@ static void waveform_vadpcm_read(void *ctx, samplebuffer_t *sbuf, int wpos, int 
 	if (seeking) {
 		if (wpos == 0) {
 			memset(&vhead->state, 0, sizeof(vhead->state));
-			vhead->current_rom_addr = wav->rom_addr;
+			lseek(wav->current_fd, wav->base_offset, SEEK_SET);
 		} else {
 			assertf(wpos == wav->wave.len - wav->wave.loop_len,
 				"wav64: seeking to %x not supported (%x %x)\n", wpos, wav->wave.len, wav->wave.loop_len);
 			memcpy(&vhead->state, &vhead->loop_state, sizeof(vhead->state));
-			vhead->current_rom_addr = wav->rom_addr + (wav->wave.len - wav->wave.loop_len) / 16 * 9;
+			lseek(wav->current_fd, (wav->wave.len - wav->wave.loop_len) / 16 * 9, SEEK_CUR);
 		}
 	}
 
@@ -208,8 +223,7 @@ static void waveform_vadpcm_read(void *ctx, samplebuffer_t *sbuf, int wpos, int 
 		void *src = (void*)dest + ((nframes*16) << SAMPLES_BPS_SHIFT(sbuf)) - src_bytes;
 
 		// Fetch compressed data
-		dma_read(src, vhead->current_rom_addr, src_bytes);
-		vhead->current_rom_addr += src_bytes;
+		read(wav->current_fd, src, src_bytes);
 
 		#if VADPCM_REFERENCE_DECODER
 		if (wav->wave.channels == 1) {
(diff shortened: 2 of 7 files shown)


include/wav64.h
```diff
@@ -32,8 +33,9 @@ typedef struct {
 	 */
 	waveform_t wave;
 
-	/** @brief Absolute ROM address of WAV64 */
-	uint32_t rom_addr;
+	/** @brief File descriptor to read WAV64 */
+	int currentFile;
```
rasky: We don't use camelcase in generale, so this should be `current_file` or better `current_fd` or even just `fd` since there's not a "previous" or "next" one to justify calling this "current".

include/wav64.h
```diff
@@ -32,8 +33,9 @@ typedef struct {
...
+	int baseOffset;
```
rasky: `base_offset`

src/audio/wav64.c
```diff
@@ -148,8 +150,23 @@ static inline void rsp_vadpcm_decompress(void *input, int16_t *output, bool ster
...
+	uint8_t* ram_addr = (uint8_t*)samplebuffer_append(sbuf, wlen);
+	int bytes = wlen << bps;
+
+	uint32_t t0 = TICKS_READ();
+	// Run the DMA transfer. We rely on libdragon's PI DMA function which works
+	// also for misaligned addresses and odd lengths.
+	// The mixer/samplebuffer guarantees that ROM/RAM addresses are always
+	// on the same 2-byte phase, as the only requirement of dma_read.
+	read(currentFile, ram_addr, bytes);
+	lseek(currentFile, -bytes, SEEK_CUR);
```
rasky: I can't understand how it works if you're seeking back, and ignoring `wpos` at the same time. I suggest you test the implementation with `rom:/` since `pi:/` is pretty new too.

src/audio/wav64.c
```diff
@@ -165,7 +182,7 @@ void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wl
 static void waveform_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wlen, bool seeking) {
 	wav64_t *wav = (wav64_t*)ctx;
 	int bps = (wav->wave.bits == 8 ? 0 : 1) + (wav->wave.channels == 2 ? 1 : 0);
-	raw_waveform_read(sbuf, wav->rom_addr, wpos, wlen, bps);
+	raw_waveform_read(sbuf, wav->currentFile, wpos, wlen, bps);
```
rasky: You should check `seeking` before calling this function. When `seeking` is `false`, the mixer guarantee that the requested read if subsequent to the previous one, so you could just call `read` without any seek. Instead, if `seeking` is true, you should really first `lseek` to `wpos << bps` and then call `read()`. Basically when seeking is false, `wpos` is just informative but it's not really required.

src/audio/wav64.c
```diff
@@ -175,12 +192,14 @@ static void waveform_vadpcm_read(void *ctx, samplebuffer_t *sbuf, int wpos, int
 	if (seeking) {
 		if (wpos == 0) {
 			memset(&vhead->state, 0, sizeof(vhead->state));
-			vhead->current_rom_addr = wav->rom_addr;
+			vhead->currentFile = wav->currentFile; // TODO: The file* is not is not an instance, so seeking may not be safe but it works.
+			lseek(vhead->currentFile, wav->baseOffset, SEEK_SET);
```
rasky: I don't understand the comment. Can you explain?

src/audio/wav64.c
```diff
@@ -255,19 +273,12 @@ static void waveform_vadpcm_read(void *ctx, samplebuffer_t *sbuf, int wpos, int
...
-		fn += 5;
-	}
-
-	int fh = dfs_open(fn);
-	assertf(fh >= 0, "error opening file %s: %s (%d)\n", fn, dfs_strerror(fh), fh);
-
+	// Open the input file.
+	int fptr = 0;
+	fptr = open(fn, O_RDONLY);
+	assertf(fptr >= 0, "error opening file %s: FILE == NULL\n", fn);
```
rasky: There's a function called `must_open()` in `asset.c`. You can declare it in `asset_internal.h` and call it from here. It is a wrapper to `open()` that provides nice error messages in case of common programming mistakes.

src/audio/wav64.c
```diff
@@ -283,8 +294,10 @@ void wav64_open(wav64_t *wav, const char *fn) {
 	wav->wave.frequency = head.freq;
 	wav->wave.len = head.len;
 	wav->wave.loop_len = head.loop_len; 
-	wav->rom_addr = dfs_rom_addr(fn) + head.start_offset;
+	wav->currentFile = open(fn, O_RDONLY);
```
rasky: I don't think there's a need to open a second file here, is there? You could just store the file descriptor here and avoiding closing it later.

src/audio/wav64.c
```diff
@@ -309,20 +321,21 @@ void wav64_open(wav64_t *wav, const char *fn) {
...
 
-	dfs_close(fh);
+	close(fptr);
 }
 
 void wav64_play(wav64_t *wav, int ch)
 {
 	// Update the context pointer, so that we try to catch cases where the
 	// wav64_t instance was moved.
+	lseek(wav->currentFile, wav->baseOffset, SEEK_SET);
```
rasky: This shouldn't be required if you implement the correct semantic when `seeking=True`.

src/audio/wav64.c
```diff
@@ -362,5 +375,8 @@ void wav64_close(wav64_t *wav)
 			break;
 		}
 		wav->ext = NULL;
+		if (wav->currentFile >= 0) {
+			close(wav->currentFile);
```
rasky: Please also invalidate the variable by setting it to -1. It helps with diagnostics if the user tries to use the wav object again by mistake.

src/audio/wav64_opus.c
```diff
@@ -54,26 +55,19 @@ typedef struct {
 /// @brief Wav64 Opus state
 typedef struct {
     wav64_opus_header_ext xhead;    ///< Opus header extension
-    uint32_t current_rom_addr;      ///< Current ROM reading address
+    int currentfile;      ///< Current ROM reading address
```
rasky: `int fd` here too.

src/audio/wav64_opus.c
```diff
@@ -97,17 +91,19 @@ static void waveform_opus_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wl
...
+            read(st->currentfile, buf, nb);
 
             // Update ROM pointer after current frame
-            st->current_rom_addr += nb + 2;
-            if (st->current_rom_addr & 1)
-                st->current_rom_addr += 1;
+            int currentStreamOffset = lseek(st->currentfile, 0, SEEK_CUR);
+            if (currentStreamOffset & 1) {
+                lseek(st->currentfile, 1, SEEK_CUR);
+            }
```
rasky, line 101: The file position is always 2-byte aligned here at the point that the frame begins:
https://github.com/DragonMinded/libdragon/blob/7ce2ab0a487de6c9234c619a0d2ae6a6943afdb3/tools/audioconv64/conv_wav64.c#L401

In other words, `nb` always starts at an even offset. So basically the idea here is that you must skip 1 byte only if `nb` is odd. Or even better, to save the `lseek` call, you could simply read 1 more byte in the read() when `nb` is odd (obviously make sure to increase `but` by one byte in that case).

src/audio/wav64internal.h
```diff
@@ -43,8 +43,15 @@ typedef struct samplebuffer_s samplebuffer_t;
 /**
  * @brief Utility function to help implementing #WaveformRead for uncompressed (raw) samples.
  * 
- * This function uses PI DMA to load samples from ROM into the sample buffer.
+ * This function uses a FILE* to load samples from ROM into the sample buffer.
```
rasky: It's not a `FILE*`, it is a `file descriptor`

rasky (changes requested):
Thanks for the PR, I've sent a few comments in

src/audio/wav64.c
nopjne: What I meant is: The original assignment of rom_addr could be used to allow wav->rom_addr and vhead->current_rom_addr to diverge later on. This is currently not happening based on testing, but if it were to happen we woudl need to open the file again here.

src/audio/wav64_opus.c
nopjne: I can get rid of the first 1st lseek and check (nb & 1) instead. Nice optimization!
For the second one Using lseek to advance the fileptr sound like a better solution here right? Otherwise read would compile to a load instruction that may cause a cache miss and have to go over the PI bus for nothing. Also have to allocate some stack space for a read.

src/audio/wav64.c
nopjne: There are reads that happen later in this function, while the receiver of the wav struct assumes the file pointer not to have advanced from this position. I can do it the way you suggest but I'll have to lseek back because the file is not instanced.

src/audio/wav64_opus.c
rasky: It's just one additional byte. PI bus is 16 bit anyway, so if you request an odd number of bytes, one more byte is transferred anyway. lseek is fast but can't be faster than just reading a byte. If you're not convinced, you could measure it with `get_ticks()`, I might be wrong.

src/audio/wav64.c
rasky: `head.start_offset` is not guaranteed to be the current position at that point; the format specifies the actual starting offset in the header specifically so that the body of data can begin at any point after the header. So you have to seek to `head.start_offset` before returning anyway.

include/wav64.h
```diff
@@ -8,6 +8,7 @@
 #define __LIBDRAGON_WAV64_H
 
 #include "mixer.h"
+#include <stdio.h>
```
rasky: This is not required anymore

src/audio/opus/celt_decoder.c
```diff
@@ -194,15 +194,15 @@ int celt_decoder_init(CELTDecoder *st, opus_int32 sampling_rate, int channels)
       return ret;
    st->downsample = resampling_factor(sampling_rate);
    if (st->downsample==0)
-      return OPUS_BAD_ARG;
+      assert(0);
```
rasky: Why these changes are necessary?

nopjne: It was easier to debug when I knew where the assert was. Otherwise the faulting location was non obvious because it was the calling function that would assert. If you prefer I can leave this out.

rasky: I'd like to minimize the changes in libopus itself to facilitate merges. It's true that asserts in our codebase make it easier to debug but that's true for hundreds of similar return codes, and we can't just change the whole libopus.

src/audio/wav64.c
```diff
@@ -165,7 +181,15 @@ void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wl
...
-	raw_waveform_read(sbuf, wav->rom_addr, wpos, wlen, bps);
+	int start_position;
+	if (seeking) {
+		start_position = lseek(wav->current_fd, 0, SEEK_CUR);
+		lseek(wav->current_fd, wav->base_offset + (wpos << bps), SEEK_SET);
+	}
+	raw_waveform_read(sbuf, wav->current_fd, wpos, wlen, bps);
+	if (seeking) {
+		lseek(wav->current_fd, start_position, SEEK_SET);
+	}
```
rasky: There is no need to seek back to the previous position when `seeking` is `true`. This also means you don't need the initial `lseek` to read the current position.

src/audio/wav64_opus.c
```diff
@@ -54,29 +55,29 @@ typedef struct {
 /// @brief Wav64 Opus state
 typedef struct {
     wav64_opus_header_ext xhead;    ///< Opus header extension
-    uint32_t current_rom_addr;      ///< Current ROM reading address
+    int current_fd;                 ///< Current file descriptor for the wav64
```
rasky: Given that the code below always have a pointer to the wav64 and you have the file descriptor there, I think you don't need this variable here. There is only one object (the wav64), only one file descriptor, and only one current position being read.

src/audio/wav64_opus.c
```diff
@@ -54,29 +55,29 @@ typedef struct {
...
+    int rom_offset;
```
rasky: I think this is unused?

src/audio/wav64_opus.c
```diff
@@ -97,17 +98,18 @@ static void waveform_opus_read(void *ctx, samplebuffer_t *sbuf, int wpos, int wl
...
+                aligned_frame_size += 1;
+            }
 
-            // Update ROM pointer after current frame
-            st->current_rom_addr += nb + 2;
-            if (st->current_rom_addr & 1)
-                st->current_rom_addr += 1;
+            // Read frame
+            data_cache_hit_writeback_invalidate(buf, aligned_frame_size);
+            read(st->current_fd, buf, aligned_frame_size);
```
rasky: You need to increase the size of `buf` by 1, otherwise this can cause a buffer overflow.

include/wav64.h
```diff
@@ -32,8 +32,9 @@ typedef struct {
 	 */
 	waveform_t wave;
 
-	/** @brief Absolute ROM address of WAV64 */
-	uint32_t rom_addr;
+	/** @brief File descriptor to read WAV64 */
+	int current_fd;
+	int base_offset;
```
rasky: This is missing a documentation comment.

src/audio/wav64.c
```diff
@@ -148,7 +151,20 @@ static inline void rsp_vadpcm_decompress(void *input, int16_t *output, bool ster
...
-void raw_waveform_read(samplebuffer_t *sbuf, int base_rom_addr, int wpos, int wlen, int bps) {
+void raw_waveform_read(samplebuffer_t *sbuf, int current_fd, int wpos, int wlen, int bps) {
+	uint8_t* ram_addr = (uint8_t*)samplebuffer_append(sbuf, wlen);
+	int bytes = wlen << bps;
+
+	uint32_t t0 = TICKS_READ();
+	// Run the DMA transfer. We rely on libdragon's PI DMA function which works
+	// also for misaligned addresses and odd lengths.
+	// The mixer/samplebuffer guarantees that ROM/RAM addresses are always
+	// on the same 2-byte phase, as the only requirement of dma_read.
```
rasky: This comment does not apply anymore. `read` will always work irrespective of the buffer alignment, thought it might force a memory copy if the alignment is not respected. I would just remove the comment.

rasky (changes requested):
I want to give a final look after all changes are made, but I guess it's starting to look fine.

PR #608, opened 2024-09-18
Support DiskIO interface override

nopjne:
This change adds the ability for an app using libdragon to override the diskio interface.


rasky:
Prompted by this, I opted for a more radical refactoring of FAT handling code and extracted a proper FAT library in the preview branch:
https://github.com/DragonMinded/libdragon/blob/preview/include/fat.h

So I'm closing this PR now, as the new library should solve the same issue.

PR #609, opened 2024-09-21
Add function for getting the size of a DFS file without calling `dfs_open`

danbolt:
## About
This change implements `dfs_rom_size`, which returns the size of a file without calling `dfs_open`.

## Rationale
This can be used in conjunction with `dfs_rom_addr` and `dma_read` to get ROM data without needing to open a file handle. This can be useful in scenarios where a programmer might want to avoid invoking `malloc` (eg: after a level has been loaded).

### DIFF
--- a/src/dragonfs.c
+++ b/src/dragonfs.c
@@ -701,6 +701,24 @@ int dfs_dir_findnext(char *buf)
     return __dfs_findnext(buf, &next_entry);
 }
 
+static bool can_use_hash(const char *path)
+{
+    //Check for chdir
+    if(directory_top != 0) {
+        return false;
+    }
+    //Check for ./ or .. in path
+    while(*path) {
+        if(path[0] == '.') {
+            if(path[1] == '.' || path[1] == '/') {
+                return false;
+            }
+        }
+        path++;
+    }
+    return true;
+}
+
 static uint32_t prime_hash(const char *str, uint32_t prime)
 {
     uint32_t hash = 0;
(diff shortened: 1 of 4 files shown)


danbolt:
I can amend `test_dfs.c` to accommodate this function.

rasky:
I'm usually a bit "meh" about adding APIs to dragonfs.h as I'd like people to use syscalls instead, but in this case you're looking for the lowest possible overhead and in fact you're basically bypassing dfs as well, you just want ROM data. So I guess it's fine. 

Please add the test, yes.

src/dragonfs.c
```diff
@@ -979,6 +979,22 @@ uint32_t dfs_rom_addr(const char *path)
...
+    //Skip initial slash
+    if(path[0] == '/') {
+        path++;
+    }
+    dfs_lookup_file_t *entry = lookup_file(path);
+    
+    if(!entry)
+    {
+        //File not found
+        return 0;
```
rasky: Can you maybe change this to a negative error value like -1 to signal the error condition? I don't remember offhand whether we support 0-byte files in the filesystem, but it seems safer in general to use negative values for error values. It's not like you need more than 31 bits to represent the size anyway.

src/dragonfs.c
```diff
@@ -979,6 +979,22 @@ uint32_t dfs_rom_addr(const char *path)
...
+    }
+    return base_ptr+entry->data_len;
```
rasky: I'm pretty sure you shouldn't add `base_ptr` here?

rasky:
Also have a look at https://github.com/DragonMinded/libdragon/pull/607 which I just merged

src/dragonfs.c
danbolt: I definitely shouldn't! Unless I wanted to have a Libultra-style SEGEMENT_END bit!

src/dragonfs.c
danbolt: Updated. I've also added some explicit casting between signed/unsigned when returning the value.

Let me know if there are any style concerns about bracket usage.

rasky (approved):
Thanks I think it's OK. The only nit is to change the error code to DFS_ENOFILE, but I'll do that myself.

danbolt:
Thank you Rasky! It's heartening to see this added! <3

PR #611, opened 2024-10-02
Add basic TLS Support

gamemasterplc:


src/exception.c
```diff
@@ -38,6 +43,13 @@ extern volatile reg_block_t __baseRegAddr;
 /** @brief Syscall exception handlers */
 static syscall_handler_entry_t __syscall_handlers[MAX_SYSCALL_HANDLERS];
 
+/** Flag for kernel being initialized */
+extern bool __kernel;
```
rasky: Include kernel_internal.h instead of redefining this.

src/kernel/kernel_internal.h
```diff
@@ -41,20 +41,18 @@
...
-		#endif
-    } tls;  ///< Thread-local storage
+	/** TLS Pointer */
+	void *tp_value;
+	/** Mirror of __interrupt_depth */
+	int interrupt_depth;
+	/** Mirror of __interrupt_sr  */
+	int interrupt_sr;
+	/** Newlib reentrancy */
+	struct _reent *reent_ptr;
```
rasky: Any reason to remove the tis structure? I think It helped grouping them and providing also some implicit documentation.

src/exception.c
```diff
@@ -16,6 +16,11 @@
 #include <stdbool.h>
 #include <math.h>
 
+/** Invalid TLS Minimum Address */
+#define TLS_INVALID_MIN (uint32_t)(KERNEL_TP_INVALID-28672)
+/** TLS Size */
+#define TLS_SIZE ((uint32_t)(__tls_end)-(uint32_t)(__tls_base))
```
rasky: Can you maybe move these to kernel_internal.h? You could move there also TP_OFFSET so that you reuse the magic offset.

src/kernel/kernel.c
```diff
@@ -15,6 +15,10 @@
 #define STACK_COOKIE   0xDEADBEEFBAADC0DE
 #define STACK_GUARD    64
 
+#define TDATA_SIZE ((uint32_t)(__tdata_end)-(uint32_t)(__tdata_start))
+#define TLS_SIZE ((uint32_t)(__tls_end)-(uint32_t)(__tls_base))
+#define TP_OFFSET 28672
```
rasky: You could spell this 0x7000 so that it looks a bit less arbitrary

src/kernel/kernel.c
```diff
@@ -359,8 +389,10 @@ kthread_t* kernel_init(void)
 	th_main.name = "main";
 	th_main.stack_size = 0x10000; // see STACK_SIZE in system.c
 	th_main.flags = TH_FLAG_DETACHED; // main thread cannot be joined
+	th_main.tp_value = th_main_tp;
```
rasky: th_main_tp seems to be a constant. Any reason to define it as a variable, rather than just using `__tls_base+TP_OFFSET` here?

src/kernel/kernel.c
gamemasterplc: Done in kernel_internal.h

src/kernel/kernel.c
gamemasterplc: That also works, so done.

Issue #158, opened 2021-07-27
Remove the header from repository: create an open source IPL3


anacierdem:
We can add a new one once we have a non commercial IPL3.

jago85:
What's the reason for this? To prevent copyright issues? 

What should we do until we have a new IPL3? Find the current one somewhere else (or in the revision history 😉)? I mean, we cannot create a bootable ROM without the loader.

Is there any work in progress for a new IPL3?

For the UltraPIF I also need a modified version of the loader. I've alway been unsure if I'm allowed to publish this code as it might be copyrighted. (My modifications are only a small part to load the application from the PIF instead of the cart.)

Keep in mind that the PIF compares the checksum calculated on the IPL3 to the checksum stored in the CIC. 

I have reverse engineered that checksum algorithm from the PIF ROM. (https://github.com/jago85/PifChecksum) But we'd need a reversed algorithm. The new IPL3 needs to be modified that it gives the same checksum or the PIF will lock out the system.

With the UltraPIF I don't have this problem, because I'm not checking this. But I need the loader as it initializes the RDRAM stuff (where I copy the UltraPIF menu to). I started reverse engineering the RDAM initialization but lost motivation halfway through. :laughing:

rasky:
The idea is that closing this issue requires an open source IPL3.

There is a stub here: https://github.com/hcs64/boot_stub

This already passes the checksum check in IPL2 (thanks a GPU-accelerated cracker that was written, linked from that repo).

The complex part is writing an open source IPL3 that does something similar/compatible to the official ones, while being clean-room implemented. The main task in there is RDRAM initialization, which is a complex system which has been already partially reversed and documented. 

From an implantation standpoint, it is hard to debug what happens while the RAM is not working. My plan would be to implement a printf primitive in assembly which goes through 64drive USB (similar to what debug.c does).

anacierdem:
Yes it needs significant effort to implement a custom clean room IPL3. Until then, it will live for us for some while. As it is one of the major setbacks for a fully free sdk, this will need collaboration from the whole homebrew community.
I'm pretty sure someone has asm for printing stuff on screen at hand, that will be useful indeed.

jago85:
Maybe I can help someday. With the UltraPIF a can inject any binary to SP_DMEM to be executed without any checks. I can also provide a simple way to send data/characters to a terminal via the UltraPIF's UART using IO writes to the PIFRAM.

bsmiles32:
[...]

PR #617, opened 2024-10-11
Merge open-source IPL3 to trunk

rasky:
This commit series finally merges the new open-source IPL3 to trunk, using it by default for all ROMs. 

The IPL3 underwent large scale testing in all preview-based ROMs, on many different consoles, including iQue. It had several rounds of improvements and bug fixes. It now seems very stable. Currently there are no known bugs.

It boots normal homebrew ROMs in less than 100 ms (more than 5 times faster than Nintendo's IPL3), and support compressed code/data, using the same battle-tested algorithms of the asset library. It also has no minimum ROM size requirement, so many small examples and tests start from ~150 KiB.

Please refer to `boot/README.md` for technical information on the new IPL3.


anacierdem:
Closes #158

PR #630, opened 2024-10-18
Change prefix handling

anacierdem:
The build system is stripping the file prefix to clean it and make it readable. OTOH that was preventing changing it, which is useful when you have multi-sub-projects built by n64.mk, such as tiny3d. The debug symbols then are all relative paths like `/src/*` and it is not possible to properly map them to different sub projects for debugging. This change allows overriding the prefix so that it is possible to do this:

```Makefile
.PHONY: tiny3d
tiny3d:
	$(MAKE) -C $(T3D_INST) FILE_PREFIX=tiny3d
```

Then, all files become prefixed by `tiny3d/` allowing mapping to the workspace source files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ all: libdragon
 V = 1  # force verbose (at least until we have converted all sub-Makefiles)
 SOURCE_DIR = src
 BUILD_DIR = build
-FILE_PREFIX=libdragon
 include n64.mk
 INSTALLDIR = $(N64_INST)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,6 +3,7 @@ all: libdragon
 V = 1  # force verbose (at least until we have converted all sub-Makefiles)
 SOURCE_DIR = src
 BUILD_DIR = build
+FILE_PREFIX=libdragon
```
rasky: I think `FILE_PREFIX` is too generic. This is a very specific macro for a very specific thing, so I'd use a longer name. Something like `N64_BACKTRACE_FILE_PREFIX`.

I'd use the `N64` prefix because it's something related to N64 not your PC host, and we use the `N64` prefix on some/most of those macros.

n64.mk
```diff
@@ -1,5 +1,6 @@
 BUILD_DIR ?= .
 SOURCE_DIR ?= .
+FILE_PREFIX ?=
```
rasky: I'd rather to move this down together with the other N64 macros, and with a comment

anacierdem:
Thank you! I did the changes. Notice that I've also changed `?=` to `=` b/c changing the variable globally doesn't mean much. It is only meant to be overridden via [make parameters](https://www.gnu.org/software/make/manual/html_node/Overriding.html) or [target specific variables](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html).

PR #631, opened 2024-10-21
Allow pifiles to be writable in the DOM2 area

nopjne:
This change adds code to be able to open DOM2 space as a regular file handle.

### DIFF
--- a/src/pifile.c
+++ b/src/pifile.c
@@ -24,11 +24,6 @@ typedef struct {
 
 static void *__pifile_open(char *name, int flags)
 {
-    if (flags != O_RDONLY) {
-        errno = EACCES;
-        return NULL;
-    }
-
     // Parse the name in the foramt "ADDR:SIZE" as hexdecimal numbers
     char *end;
     uint32_t base = strtoul(name, &end, 16);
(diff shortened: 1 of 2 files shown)


src/pifile.c
```diff
@@ -123,6 +125,41 @@ static int __pifile_read(void *file, uint8_t *buf, int len)
...
+        // Go through a temporary buffer
+        uint8_t *tmp = alloca(512+1);
+        if ((f->base + f->ptr) & 1) tmp++;
+
+        while (len > 0) {
+            int n = len > 512 ? 512 : len;
+            data_cache_hit_writeback_invalidate(tmp, n);
+            dma_write_raw_async(tmp, f->base + f->ptr, n);
+            dma_wait();
+            memcpy(buf, tmp, n);
```
rasky, line 153: This seems wrong, seems like you haven't updated this code.

src/pifile.c
```diff
@@ -42,6 +37,13 @@ static void *__pifile_open(char *name, int flags)
         return NULL;
     }
 
+    if ((base < 0x08000000) || (base >= 0x09000000)) {
+        if (flags != O_RDONLY) {
+            errno = EACCES;
+            return NULL;
+        }
+    }
+
```
rasky, line 46: I don't think we need this at this level. In general any device can expose any writeable area at any PI address, so I wouldn't hardcode the range. Eg: many flashcarts allow to write the internal SDRAM by writing at 0x10000000, maybe after enabling writes with some register.

So I would just remove the check and let the user know what they are doing.

rasky (changes requested):
Thanks! Unfortunately I don't trust `dma_write_raw_async` especially for misaligned writes because I have not verified it on hardware and also Ares doesn't implement it.

If you want to push this forward, we need a test that performs DMA in a R/W area (eg: SRAM) for writing, and the performs a series of writes + reads + tests at different addresses, different sizes, and different length. Check `test_dma_read_misalign`, we would need something similar for DMA writes.

PR #634, opened 2024-10-24
Move threads.h to new newlib override folder searched before other includes

sp1187:
This PR primarily makes it easier to create software packages for libdragon that doesn't conflict with newlib packages, though it is also useful for documentation reasons (easy to see which files are meant to override newlib includes) and possibly useful for testing or reverting during libdragon development as well.


Makefile
```diff
@@ -212,6 +211,8 @@ install: install-mk libdragon
 	install -Cv -m 0644 src/fatfs/ff.h $(INSTALLDIR)/mips64-elf/include/fatfs/ff.h
 	install -Cv -m 0644 src/fatfs/ffconf.h $(INSTALLDIR)/mips64-elf/include/fatfs/ffconf.h
 	install -Cv -m 0644 include/ioctl.h $(INSTALLDIR)/mips64-elf/include/ioctl.h
+	mkdir -p $(INSTALLDIR)/mips64-elf/include/newlib_overrides
```
anacierdem, line 214: Should we name it to something like early_includes or is this only meant for newlib only?

sp1187: The current file in the folder overrides newlib, so I thought the chosen name here was the most clearly documenting one (it describes its purpose rather than just what it does), and I also suspect it will be common use-case for future includes. If we think there is a need to override files from some other source (libstdc++? some GCC-provided includes?) in the future, I can rename the folder.

PR #639, opened 2024-11-03
Fix kcond assertion

jameskeane:
The assertion condition was backwards.


rasky:
Manually merged: https://github.com/DragonMinded/libdragon/commit/d8c732130428e3b52c32faad313266b408955554

PR #640, opened 2024-11-06
videoplayer example bug - Update videoplayer.c

joshkautz:
Add `rom:/` prefix to properly locate the ROM wav64 asset.

Currently, running the ROM compiled from the example's Makefile will yield a CPU Assertion:
```
ASSERTION FAILED: fd >= 0
file "src/asset.c", line 83, function: must_open
File not found: movie.wav64
Did you forget the filesystem prefix? (e.g. "rom:/")
```

videoplayer 2024-11-06 14-06-43

This fixes the issue for me. Hopefully this is correct! 🤞

### DIFF
--- a/examples/videoplayer/videoplayer.c
+++ b/examples/videoplayer/videoplayer.c
@@ -77,7 +77,7 @@ int main(void)
 
 	// Open the audio track and start playing it in channel 0.
 	wav64_t audio_track;
-	wav64_open(&audio_track, "movie.wav64");
+	wav64_open(&audio_track, "rom:/movie.wav64");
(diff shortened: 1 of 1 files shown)


rasky:
This is OK but the code should work already because `wav64_open()` should have an undocumented backward compatibility codepath to allow unprefixed filenames to work. This codepath was erroneously removed in #606 and I've just readded it in https://github.com/DragonMinded/libdragon/commit/4185cd486024a86aa9d2bae019b5922fa21dcb18.

This said, your PR is still OK to bring the example code to the latest standards.

PR #657, opened 2025-01-03
Nixify

knedlsepp:
This adds a development environment using the nix package manager.
It's somewhat of an alternative to the current Dockerfile, but has more
powerful caching and reproducibility guarantees.
Also I included an integration test that:
- Builds libdragon
- Builds the examples
- Starts a virtual machine
- Runs the `audioplayer.z64` example ROM in `ares`
- Uses OCR to wait for the "Module Audio Player" to appear on screen
- Takes a screenshot for later inspection.

It can be run using `nix build .#checks.x86_64-linux.vmTest`.
There's lots more neat things that can be done with nix.
I'm putting this out as a draft mostly for visibility.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1735531152,
+        "narHash": "sha256-As8I+ebItDKtboWgDXYZSIjGlKeqiLBvjxsQHUmAf1Q=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "3ffbbdbac0566a0977da3d2657b89cbcfe9a173b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-24.11",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,212 @@
+{
+  description = "Open source library for N64 development.";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
+
+  outputs = { self, nixpkgs }:
+    let
+
+      lastModifiedDate = self.lastModifiedDate or self.lastModified or "19700101";
+      version = builtins.substring 0 8 lastModifiedDate;
+      supportedSystems = [ "x86_64-linux" ];
+      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+      nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; overlays = [ self.overlays.default ]; });
+
+    in
+
+    {
+      overlays.default = final: prev: {
+        # FIXME: Check if we can directly use nixpkgs' mips64-embedded toolchain
+        toolchain = with final; stdenv.mkDerivation {
+          pname = "toolchain";
+          inherit version;
+
+          src = lib.fileset.toSource {
+            root = ./.;
+            fileset = lib.fileset.intersection
+              (lib.fileset.gitTracked ./.)
+              (lib.fileset.unions [
+                ./tools/build-toolchain.sh
+              ]);
+          };
+
+          buildPhase =
+            let
+              binutils = fetchurl {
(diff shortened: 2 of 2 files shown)


bates64:
I've had success using a custom `crossSystem` and then using `pkgs.mkShell` to get the toolchain. [Example](https://github.com/bates64/papermario-dx/blob/main/flake.nix)
```nix
let
    crossSystem = {
        config = "mips64-elf";
        system = "mips64-elf";
        gcc.arch = "vr4300";
        gcc.tune = "vr4300";
        gcc.abi = "32";
    };
    pkgsCross = import nixpkgs { inherit system, crossSystem; };
in
    devShells.default = pkgsCross.mkShell {
        # ...
    }
```
Using NixOS's mips64-embedded might work, but it's not tuned to the vr4300, and I don't believe it's built on Hydra anyway. I've been using a Cachix cache so users don't have to build from source: https://papermario-dx.cachix.org

knedlsepp:
> I've had success using a custom `crossSystem`

Neat! Thanks for sharing! 😊

rasky:
Hello! This PR has been in draft for a while, so I thought I could comment on it.

I personally don't use nix, and I don't think any other of Libdragon maintainers do as well. I know very little of it and I am not able to maintain this script. There are other situations where build scripts for package managers are externally maintained (eg. AUR).

If this script must be committed to the repository for nix to work correctly, I think we should probably adjust a few things:

* I think it's mandatory to have the CI build the project via nix to verify the script doesn't bitrot, in case it ceases being actively maintained.
* The script should incorporate the minimum amount of information about libdragon, to minimize depending on details that might change over time. By just eyeing the script, I see:
  * There is a list of tar.gz files with URLs which are normally downloaded by build-toolchain.sh. Not sure why the list of those packages, versions and URLs is duplicated in the nix script, but if possible they should be removed, otherwise it's a second place we must remember to update anytime we update the GNU packages (and we normally do multiple times per year)
  * There is a list of files/directories including src, boot, n64.ld, n64.mk. I'm not exactly sure why this list exists in the script and what is the rationale for listing only those files/directories. This also smells like a potential maintenance problem as libdragon might evolve to need more/fewer files in that list. I don't have a suggestion about this, but possibly this knowledge should be evicted from the script if possible at all.

Let us know your plan on this PR, thanks!

knedlsepp:
Hi @rasky!
Thank you for taking the time to comment on this draft! First of all I agree with all your mentioned changes in order to consider a merge of this PR.
I'd like to provide some further information on the points you mentioned. Firstly this PR is not strictly required in order to support building libdragon using nix. Out-of-tree packaging can be done similarly as on AUR as you mention. (Searching for this reveals that in fact somebody already created such an out-of-tree package for libdragon on nix: https://github.com/nadiaholmquist/libdragon-flake recently). The major limitation with this strategy is that it's not really possible to use the exact same "blessed" toolchain that is used here in `tools/build-toolchain.sh` due to some peculiarities of `nix`.

> I think it's mandatory to have the CI build the project via nix

Fully agree.

> There is a list of tar.gz files with URLs which are normally downloaded by build-toolchain.sh

As you noticed I duplicated all links to tarballs in the `flake.nix`. This is because in `nix` all externally downloaded data needs to be specified with a checksum to guarantee reproducibility and provide automatic download caching. In order to not have this information duplicated across `flake.nix` and `build-toolchain.sh` this would need to be stored in some machine readable form supported by `nix` such as a JSON file that would also be read by `build-toolchain.sh`.

> There is a list of files/directories including src, boot, n64.ld, n64.mk.

These lists in `flake.nix` enable fine-grained build caching by nix that's very similar to `Makefile` target dependencies. A change in a `README.md` will thus not trigger a rebuild of the library. (It is even better than a  `Makefile` target dependency because the nix build sandbox ensures that no other files will be made available to the build. This ensures no dependency can be forgotten.) So evicting these lists could be replaced by `./.` with the accompanying performance penalty.

> Let us know your plan on this PR, thanks!

I opened this PR because I was playing around with the library and thought it might be interesting to share that I could get some integration tests going that run libdragon in an emulator and then assert on the video output as I had mixed results with running the examples in different emulators. I was also curious if there was some overlap in libdragon and nix users.

Seeing that `nix` is still quite niche and there is little familiarity with it, I now think it's probably a bad idea to introduce potentially unmaintained code to this repository when it likely will go unused given that I haven't found any time to play around with libdragon myself since this PR.

Thank you for maintaining this cool library and sorry for having taken your time with this. :relaxed:

lucasfcnunes:
@knedlsepp, https://github.com/nadiaholmquist/libdragon-flake is giving a 404 now. ☹

PR #663, opened 2025-01-07
Updating root .gitignore file and example .gitignore files.

joshkautz:
This pattern has already been established in the `customfont` and `gldemo` examples, among others. Correct me if I'm naively overlooking something. The **_vifx_** example/demo was causing some goofy git file tracking on the `examples/vifx/filesystem/philips.rgba32.sprite` file that was created.


rasky:
thanks!

PR #667, opened 2025-01-15
Fix handling of multiple sources of texparms for configured textures

SpookyIluha:
Mainly to fix the IHQ format not behaving as expected in tiny3d

### DIFF
--- a/src/rdpq/rdpq_sprite.c
+++ b/src/rdpq/rdpq_sprite.c
@@ -34,6 +34,36 @@ static void sprite_upload_palette(sprite_t *sprite, int palidx, bool set_mode)
     }
 }
 
+// Helper function to merge the texparms from various sources including handling the detail texparms
+// Returns the merged texparms for the main texture plus the detail texparms are modified with main texture's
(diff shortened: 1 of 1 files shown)


rasky:
The PR implements a semantic which I'm not convinced of. It tries to merge the texparms specified in the .sprite file with texparms specified on the fly at runtime. Semantically speaking, I don't think this makes sense in the general case.

If I specified to mksprite that the texture is repeats=3,mirror=true and then at runtime i specify repeat=1,mirror=false, what should be the outcome? I believe one possible outcome might be an assertion ("conflicting parameters") or have the runtime one overrides the sprite ones. Instead, what this PR does, is to end up with repeats=4,mirror=true which I really can't convince myself is what the user wanted to achieve:

```C
out.s.mirror = argparms.s.mirror | sprparms.s.mirror;
out.s.repeats = argparms.s.repeats + sprparms.s.repeats;
```

I don't recall exactly why this happens to fix IHQ on Tiny3D (is that something to do with scaling?), but tiny3d unfortunately is bound to fast64 parameters which in turn are not designed for rdpq and don't match rdpq parameters very well.  So it might be that the correct solution is to be found in the longer run of making fast64 rdpq aware (emit rdpq materials instead) and switch tiny3d to use rdpq materials.

PR #669, opened 2025-01-19
Fix linker error when using Os optimization

maxgerhardt:
I noticed this bug when compiling libdragon with my own build scripts, which used `-Os` instead of `-O2`.

This generated a final linking error of

```
C:/Users/Max/.platformio/packages/toolchain-gccmips64/bin/../lib/gcc/mips64-elf/14.2.0/../../../../mips64-elf/bin/ld.exe: .pio\build\n64\FrameworkLibdragon\display.o: in function `display_init':
(.text.display_init+0x360): undefined reference to `vi_write_config'
```
I then noticed that `vi_write_safe()` and `vi_write_config()` were the only two functions that were `inline` instead of `static inline`.

After adding the `static`, the linker error went away.

Edit: Corrected a similiar error in `include/n64sys.h`.

### DIFF
--- a/include/n64sys.h
+++ b/include/n64sys.h
@@ -225,7 +225,7 @@ extern "C" {
 #endif
 
 /** @brief Return true if we are running on a iQue player */
-inline bool sys_bbplayer(void) {
+static inline bool sys_bbplayer(void) {
     extern int __boot_consoletype;
     return __boot_consoletype != 0;
 }
--- a/src/vi.h
+++ b/src/vi.h
@@ -209,7 +209,7 @@ static const vi_config_t vi_config_presets[2][3] = {
  * @param[in] value
  *            Value to be written to the register
  */
-inline void vi_write_safe(volatile uint32_t *reg, uint32_t value){
+static inline void vi_write_safe(volatile uint32_t *reg, uint32_t value){
     assert(reg); /* This should never happen */
     *reg = value;
     MEMORY_BARRIER();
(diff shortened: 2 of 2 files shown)


rasky:
Hi, thanks for this. Adding static unfortunately means that the function disappears from doxygen. The workaround used in other places is to declare the functions as "extern inline" in the C file. Grep for it to see how it's done.

maxgerhardt:
So the other `static inline` functions that *do* have doxygen comments on them like

https://github.com/DragonMinded/libdragon/blob/e037067e672a18a7f88222c1e0fa012570e71fe5/src/vi.h#L241-L251

are okay to keep as `static`? I'm not quite sure how the distinciton should be made between `static inline` being okay or not. Other files like [surface.h](https://github.com/DragonMinded/libdragon/blob/e037067e672a18a7f88222c1e0fa012570e71fe5/include/surface.h#L255-L335) also have tons of only `inline` functions, I wonder if that breaks too with `-Os` or `-Og`/`-O0` optimization, need to check.

rasky:
> So the other `static inline` functions that _do_ have doxygen comments on them like

Even if they do, they still don't show up in doxygen:
https://libdragon.dev/ref/vi_8h.html
 
> are okay to keep as `static`? I'm not quite sure how the distinciton should be made between `static inline` being okay or not. 

There is no distinction basically. You should never use `static inline` in a public header of libdragon, because that would not be documented. You can still use static inline in lidbragon .c files because those are far less important. We still generate doxygen for them for reasons, but we don't care if some functions are missing.

> Other files like [surface.h](https://github.com/DragonMinded/libdragon/blob/e037067e672a18a7f88222c1e0fa012570e71fe5/include/surface.h#L255-L335) also have tons of only `inline` functions, I wonder if that breaks too with `-Os` or `-Og`/`-O0` optimization, need to check.

They won't break because they are defined as `extern inline` as I advised you to do:
https://github.com/DragonMinded/libdragon/blob/e037067e672a18a7f88222c1e0fa012570e71fe5/src/surface.c#L75-L78

rasky:
Superseded by #689

Issue #668, opened 2025-01-19
Add mips64-elf-gdb into toolchain package


maxgerhardt:
The prebuild package at https://github.com/DragonMinded/libdragon/releases/tag/toolchain-continuous-prerelease is great! But for a complete developer setup, you also need a GDB. Now you can grab a `gdb-multiarch` from another source for sure, but it would be great if the package also just contained `mips64-elf-gdb` executable already.

meeq:
This is being worked on in https://github.com/DragonMinded/libdragon/pull/679

Once this lands, GDB will be installed in the Docker image and the packages, but native toolchain users will need to run `build-gdb.sh` after `build-toolchain.sh`.

meeq:
#679 has been merged to preview and backported to trunk. This issue can be closed as completed.

Thank you for your suggestion @maxgerhardt

PR #679, opened 2025-02-11
Add script to build + install GDB in the toolchain

meeq:
Inspired by https://github.com/DragonMinded/libdragon/issues/668

`build-gdb.sh` is now an optional follow-up script to `build_toolchain.sh`. Accepts the same environment variables.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,55 +1,66 @@
 # syntax=docker/dockerfile:1
 # V0 - Use this comment to force a re-build without changing the contents
 
+ARG BASE_IMAGE=ubuntu:22.04
+
 # Stage 1 - Build the toolchain
-FROM ubuntu:22.04
+FROM ${BASE_IMAGE} AS builder
 
 # Setup paths for the libdragon toolchain
 ARG N64_INST=/n64_toolchain
+ARG BUILD_PATH=/tmp/n64_toolchain
 ENV N64_INST=${N64_INST}
+ENV BUILD_PATH=${BUILD_PATH}
 
-# install dependencies
-RUN apt-get update
-RUN apt-get install -yq \
-    wget \
-    bzip2 \
-    gcc \
-    g++ \
-    make \
-    file \
-    libmpfr-dev \
-    libmpc-dev \
-    zlib1g-dev \
-    texinfo \
-    git
+# Install system dependencies
+RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=locked \
+    --mount=target=/var/cache/apt,type=cache,sharing=locked \
+    rm -f /etc/apt/apt.conf.d/docker-clean \
+    && apt update \
+    && apt upgrade -y \
+    && apt install -y --no-install-recommends \
+        bzip2 \
+        ca-certificates \
+        file \
+        g++ \
+        gcc \
+        git \
+        libmpc-dev \
+        libmpfr-dev \
+        make \
+        texinfo \
+        wget \
+        zlib1g-dev \
+    && apt autoremove -yq
 
 # Build toolchain
-COPY ./tools/build-toolchain.sh /tmp/tools/build-toolchain.sh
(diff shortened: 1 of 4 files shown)


anacierdem:
I'm pro GDB. In the past I was also looking into the feasibility of using it inside of docker, so would love to see it is included by default at least in the docker image.

meeq:
Build times on a Macbook Air M2, doing a clean toolchain build with dependencies already downloaded (but not unpacked):
Reference (no GDB): 2284.34s user 565.86s system 360% cpu 13:10.01 total
GDB_V=16.2: 3666.01s user 844.71s system 321% cpu 23:23.77 total

That's a pretty considerable build time increase.

anacierdem:
That’s why I prefer docker in general :) but yeah, not ideal for the regular build.

Dockerfile
```diff
@@ -1,55 +1,66 @@
...
 
 # Build toolchain
-COPY ./tools/build-toolchain.sh /tmp/tools/build-toolchain.sh
-WORKDIR /tmp/tools
-RUN ./build-toolchain.sh
-
-# Remove locale strings which are not so important in our use case
-RUN rm -rf ${N64_INST}/share/locale/*
-
+RUN --mount=type=bind,source=./tools,target=/tools \
```
anacierdem: **question:** Why bind mount instead of copying the bash scripts?

Dockerfile
```diff
@@ -1,55 +1,66 @@
...
+    --mount=type=cache,target=${BUILD_PATH} \
```
anacierdem: **question:** I think this cache mount is here to simplify continuing the build when running locally?

tools/build-gdb.sh
```diff
@@ -15,7 +15,7 @@ if [ -z "${N64_INST-}" ]; then
 fi
 
 # Dependency source libs (Versions)
-GDB_V=${GDB_V:-"16.2"}
+GDB_V=16.2
```
anacierdem, line 18: Do we want to explicitly lock the version here?

Dockerfile
meeq: Avoids an unnecessary copy, which would add an unnecessary layer to the image.

anacierdem: IIRC docker used to already process the whole build context (i.e the contents of the path Dockerfile resides in by default) and it is just a small file. Furthermore, it will be stripped in the second stage anyways. If you really think this will benefit, fine but I think it just makes it a little bit more verbose.

tools/build-gdb.sh
meeq: This is what we do for GCC and friends. I changed to a hard-coded string because of the version number grep in the GitHub Actions workflow:
https://github.com/DragonMinded/libdragon/pull/679/files#diff-be00bc5292287b3090bfd3ce8babaa79307d5539139221a6243bbdf32c4852ccR96

rasky: yeah I agree we should hardcode it, and then periodically bump it as we do with all other parts of the toolchain

anacierdem: That is, interesting :)

Dockerfile
meeq: Cache mount has been removed https://github.com/DragonMinded/libdragon/pull/679/commits/145ef5f41a9a4351d842a6116845acfecade6d48

This will bloat the Docker image cache with all of these temporary files, but they won't be in the final image.

meeq: One unfortunate effect of not caching the build path is that the tarballs need to be redownloaded every time this build step runs.

Dockerfile
meeq: Thanks for your feedback. Changed back to a COPY in https://github.com/DragonMinded/libdragon/pull/679/commits/0311fac1cb055dde51d485db38944ddb1b57d62f

rasky:
Cherry picked to trunk as well

PR #681, opened 2025-02-12
Respect BBPlayer Hack Flags in Joypad subsystem

meeq:


src/joypad.c
```diff
@@ -725,12 +746,41 @@ void joypad_poll(void)
...
 
 bool joypad_is_connected(joypad_port_t port)
 {
     ASSERT_JOYPAD_INITIALIZED();
     ASSERT_JOYPAD_PORT_VALID(port);
+
+    // Respect the BBPlayer "Hack Flags" register to swap controller ports.
+    // The device state is swapped during poll, but identifiers need to be swapped on read.
+    port = bb_hack_flags_swap_port(port);
+
```
rasky: Why do `joypad_is_connected` and `joypad_get_identifier` use the `hot` structure rather than the `cold`? I noticed since they are the only two functions that require an explicit fix for the BB swap, but I was wondering what is the rationale.

src/joypad.c
```diff
@@ -725,12 +746,41 @@ void joypad_poll(void)
...
+                joypad_devices_cold[JOYPAD_PORT_1] = joypad_devices_cold[JOYPAD_PORT_3];
+                joypad_devices_cold[JOYPAD_PORT_3] = swap;
+                break;
+            case 3:
+                // Swap port 1 and 4
+                joypad_devices_cold[JOYPAD_PORT_1] = joypad_devices_cold[JOYPAD_PORT_4];
+                joypad_devices_cold[JOYPAD_PORT_4] = swap;
+                break;
+        }
+    }
```
rasky: I think it might be more efficient to change the parsing loop above. At the top of the parsing loop, we have this:

```C
        if (sys_bbplayer()) {
            // iQue has a very fixed layout for commands, and it also tends
            // to corrupt other parts of PIF-RAM. So better jump to fixed positions
            // while parsing.
            i = (port * 8) + 1;
        }
```

It would seem like changing that to:

```C
i = (bb_hack_flags_swap_port(port) * 8) + 1;
```

would be sufficient? It's already a BB-only codepath btw.

rasky (approved):

src/joypad.c
meeq: Thanks for this suggestion. Implemented in https://github.com/DragonMinded/libdragon/pull/681/commits/79e3127f725d54fcc411af4f7fa41f861af03922

src/joypad.c
meeq: As it was implemented, device identifiers were not part of the "cold device state" struct. For consistency, this has been fixed in https://github.com/DragonMinded/libdragon/pull/681/commits/79e3127f725d54fcc411af4f7fa41f861af03922

Thank you for your thoughtful reviews.

rasky:
Merged in #682

Issue #581, opened 2024-07-15
Disable rumble pak on reset/crash


s4Ys369:
Rumble pak stays active if active when the console is reset or the exception screen pops up.

A forced reset/deactivation should be added to these, and possibly other, edge cases.

gyrovorbis:
Heh. I know a certain other homebrew SDK for a certain Caster of Dreams console which has this issue too. :rofl:

networkfusion:
If it helps, this has been added to the N64Flashcart menu on menu load in the latest release.

s4Ys369:
Using an EDX7 and iirc support is WIP

networkfusion:
Then, 
``` 
JOYPAD_PORT_FOREACH (port) {
    joypad_set_rumble_active(port, false);
}
```
might help, but I am not sure if that is able to be performed by the crash handler, as the system is probably already locked up...

rasky:
Partially fixed with #687. It does behave correctly now when pressing the RESET button.

As for the inspector, there is currently not an easy way of doing it that doesn't bring in too much additional code, for what it's worth. So we're currently punting there.

rasky:
Reopening as it's only fixed on preview for now

rasky:
Fixed with the merge of the new joypad library.

PR #687, opened 2025-02-20
Stop rumble motors upon exception or reset

meeq:
Resolves https://github.com/DragonMinded/libdragon/issues/581

Introduces a new internal API called `joypad_emergency_rumble_stop` that is used by the inspector.

The NMI/Reset handler uses an optimized code-path that only sends stop commands if needed.

Unfortunately for the inspector, we cannot rely on Joypad subsystem being initialized or in a usable state, so the "emergency stop" needs to:

1. Identify all controllers
2. Determine if they support rumble
3. Disable the rumble motors

Double-unfortunately, all of that needs to happen synchronously.

I tested this with both N64 Rumble Pak and GCN rumble-compatible controllers.

Attached is a modified joypadtest ROM that allows you to trigger an exception by pressing "B":  [joypadtest.z64.zip](https://github.com/user-attachments/files/18893272/joypadtest.z64.zip)

### DIFF
--- a/examples/joypadtest/joypadtest.c
+++ b/examples/joypadtest/joypadtest.c
@@ -95,7 +95,11 @@ int main(void)
     {
         console_clear();
 
-        printf("LibDragon Joypad Subsystem Test\n\n");
+        printf("\n");
+        printf("LibDragon Joypad Subsystem Test\n");
+        printf("Hold A to test rumble motors.\n");
+        printf("Press B to trigger an exception.\n");
+        printf("\n");
 
         joypad_poll();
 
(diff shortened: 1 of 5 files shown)


meeq:
@rasky Any other feedback on this?

As of now, only the reset handler is hooked up; an exception will not stop the rumble motors. Should I remove the emergency rumble stop completely? I don't think there's a simpler way to stop them from an unknown state.

rasky:
Yeah I'll drop the emergency rumble stop as there's not currently a good use case for it I think.

PR #700, opened 2025-03-22
toolchain version bump to latest versions

jopadan:
**Changes**
- gcc-15.2.0
- binutils-2.45
- mpfr-4.2.2
- newlib-4.5.0.20241231


rasky:
Thanks, but I'd rather wait for GCC 15 to be released so that we don't fragment the user base across slightly different toolchains.

sp1187:
Could you explain why you chose to switch the binutils variant? I don't think we use gold, and it is now deprecated as well so no need to include it in my view.

PR #711, opened 2025-05-04
Make color macros usable as constants

gamemasterplc:
Allows them to be used to initialize a global variable or in-function static variable in C. Also avoids generation of global constructors for the variables in C++.


meeq:
This will make defining color constants much cleaner and eliminate the need for hacks like this:
https://github.com/n64brew/N64brew-GameJam2024/blob/1519c402c3db8d65e966e79f438fae095b276f71/code/landgrab/color.h#L6-L7

PR #714, opened 2025-05-11
Networkfusion/test docker multiplatform

networkfusion:
Tests ability to generate multi platform docker images.

- Update docker images to use 24.04 (LTS) - 20.04 was too old and 22.04 will be too old soon.
- Support for 20.04 has been dropped in most GH actions, so start using 24.04.
- Updated GH actions for docker build and push.
- Added toolchain package generation for arm64.
- Added multiplatform docker image to support both AMD64 and ARM64 targets.
- Renamed workflows to better reflect usage.
- Improved checks based on file changes to decide whether actions should be ran.
- grep warnings from build log and add them to the action summary (requires better character handling).

NOTE: there are a couple of lines commented out as `# FIXME: This is ready in this fork but not yet in the main repo.` which require the repo to contain the packages before they can be consumed (when released from the `main` branch).


networkfusion:
These fail because of `403: forbidden` access (which is expected in this PR I guess, as I dont have the required permissions). 
image

On my fork: https://github.com/networkfusion/libdragon/actions/runs/14957748306

rasky:
@anacierdem can you please review this? Is this ok for you?

anacierdem (commented):
In general, things look correct to me. The only thing I haven't properly reviewed is the gh permissions, b/c I don't know them well enough. I'll do some reading and try to understand latest state of things. No blocker for a merge though as everything seems safe.

PR #715, opened 2025-05-15
joypad: Add custom command to inform controllers of the player #

ZenoArrows:
Useful to light up player LEDs on various custom controllers like the NSO N64 controller or an Xbox 360 controller.

### DIFF
--- a/include/joypad.h
+++ b/include/joypad.h
@@ -631,6 +631,17 @@ int joypad_get_axis_released(joypad_port_t port, joypad_axis_t axis);
  */
 int joypad_get_axis_held(joypad_port_t port, joypad_axis_t axis);
 
+/**
+ * @brief Sends the player ID to each controller.
+ *
+ * This function sends the player ID to each controller coresponding to the
+ * joybus slot the controller is plugged in to.
+ *
+ * This is a custom command which is not supported by official controllers.
+ *
+ */
+ void joypad_send_player_id(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/joybus_commands.h
+++ b/src/joybus_commands.h
@@ -129,6 +129,13 @@
  * extended controller data, including analog A and B button values.
  */
 #define JOYBUS_COMMAND_ID_GCN_CONTROLLER_READ_LONG    0x43
+
+/**
+ * @brief "Player ID" Joybus command identifier.
+ * 
+ * Custom command that is not supported by official controllers.
+ */
+#define JOYBUS_COMMAND_ID_PLAYER_IDENTIFY 0xF0
 /** @} */ /* JOYBUS_COMMAND_ID */
 
 #ifdef __cplusplus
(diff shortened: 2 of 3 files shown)


rasky:
cc @meeq

networkfusion:
An improvement here would be that the accessory is able to be detected, rather than just blindly sending the command and hoping it is successful (so we dont have the same situation as the N64Digital).

meeq:
It seems like we're sliding down the slippery slope of non-standard Joybus commands.

Which Joybus adapters support the `0xF0` command? Is this command publicly documented anywhere for interoperability (similar to the [PixelFX Game ID wiki page](https://gitlab.com/pixelfx-public/n64-game-id))?

I'm not inclined to expose this as a public API: it should not be up to each individual homebrew developer to remember to call this function, and then call it again every time the controller port assignments change.

ZenoArrows:
This is just for my own joybus adapter project, I am not aware of any adapter offering similar functionality. The idea was to simply have the N64FlashcartMenu send this command on bootup to each controller. Since the menu won't be running anymore while other homebrew is running I didn't bother accounting for the scenario where the controller port assignments change.

If this idea is accepted into the N64FlashcartMenu repo I assume we'd document it there. My main reason for suggesting this as a public API is because `joybus_exec_async ` is currently a private API.

> It seems like we're sliding down the slippery slope of non-standard Joybus commands.

The community will need some sort of central authority to assign (or re-assign) values to these non-standard Joybus commands to avoid conflicts. It might actually be a good idea to have libdragon serve as that authority.

rasky:
I think the best would be start documenting the custom commands in the n64brew wiki *and* libdragon. If we try to setup this central source of command IDs, I think it's fine for libdragon to collect also implementations for them. Maybe I would move them into some specific header like `joybus_custom.h` and/or specifically mention the product names /websites / links that implement them

meeq:
This feature would make more sense to integrate into the joypad subsystem so that every LibDragon game gets this ability implicitly. It would also make a lot more sense if there was significant adoption of this custom command in the ecosystem.

What does BlueRetro do? That’s probably the most popular Joybus adapter in the market right now. From what I can tell, they don’t have a mechanism for setting the controller LED based on port number.

Might be worth reaching out into the hardware community to see if there’s any feedback on this approach and try to get more widespread adoption.

include/joypad.h
```diff
@@ -631,6 +631,17 @@ int joypad_get_axis_released(joypad_port_t port, joypad_axis_t axis);
...
+/**
+ * @brief Sends the player ID to each controller.
+ *
+ * This function sends the player ID to each controller coresponding to the
+ * joybus slot the controller is plugged in to.
+ *
+ * This is a custom command which is not supported by official controllers.
+ *
+ */
+ void joypad_send_player_id(void);
```
meeq, line 643: If we're going to create a custom Joybus command for this, there should be lower-level joybus APIs to exercise the command, something like:

```c
/* Sets the LED for a port to the port number */
void joybus_custom_led_port( joypad_port_t port );

/* Sets the LED for a port to an arbitrary value */
void joybus_custom_led_set( joypad_port_t port, uint8_t value );
```

Additionally, as I stated in a previous comment, this high-level operation of setting the LEDs on each controller should be handled within the joypad subsystem: this feature will not be widely (or correctly) adopted if it requires developers to opt-in and handle reconnects in every LibDragon project.

`void joypad_send_player_id(void)`  is useful for testing and getting buy-in on this feature, but this approach is a little too naïve (and I don't mean that in a discouraging way). This is a great starting point. Thank you for submitting this; I can definitely see the utility in this feature for more than just your project.

ZenoArrows: I did think about that, but I didn't want to make it part of the subsystem that's baked into all homebrew, at least not immediately. It's impossible to rule out that there's some controller third-party out there that will completely break when it encounters an unknown joypad command. If it's just part of the N64FlashcartMenu and nothing else, then it's easier to back out of it if incompatibilities are found.

meeq:
I am not in favor of merging this as it is currently implemented: it’s stuck in a weird spot where there are no open-source or commercially-available devices that actually support the custom command, and the API puts the burden on individual homebrew titles to use the custom command.

I move to close this PR for now.  I think this is an interesting suggestion, but requires more buy-in from the ecosystem. [The author made this as a PR against LibDragon because joybus_exec_async was not a public API](https://discord.com/channels/205520502922543113/974342113850445874/1372614611496603748), which has been resolved.

rasky:
Ok I guess this can be a non-libdragon API for a while until the ecosystem matures.

PR #718, opened 2025-05-19
Remove unused rtctest constant

meeq:
Resolves a warning when building examples


meeq:
The RTC subsystem no longer has persistence checks (because it now supports all known RTC implementations), so this constant was just a leftover from an obsolete codepath.

PR #723, opened 2025-05-25
Add rdp interlace function

SpookyIluha:


include/rdpq.h
```diff
@@ -468,6 +468,35 @@ inline void rdpq_set_yuv_parms(uint16_t k0, uint16_t k1, uint16_t k2, uint16_t k
...
+ * This interlacing mode saves half the RDP's fillrate since it skips half the lines
+ * to draw. This is very helpful for performance in high-res mode since the system
+ * is primairly fillrate constrained.
+ * 
+ * 
+ * @param[in]   enable      True if the RDP interlacing is enabled, disabled otherwise
+ * @param[in]   keepodd     True if the RDP should only draw odd lines, otherwise only even lines are drawn
+ * 
+ */
+#define rdpq_set_interlace(enable, keepodd) ({ \
```
thekovic: I would personally prefer an always inlined (`__attribute__((always_inline))`) function over a macro given the "setter" behaviour of this API.

include/rsp_rdpq.inc
```diff
@@ -674,6 +674,28 @@ rdpq_update_fillcopy:
     #############################################################
     .func RDPQ_WriteSetScissor
 RDPQ_WriteSetScissor:
+
+    # Check whether we have sent a scissor rect command or interlacing command by checking 26th bit
+    li t6, 0x1 << 26
+    and t6, t6, a1
+    beqz t6, 3f
+    nop
```
HailToDodongo: You can move the `lw a2, %lo(RDPQ_SCISSOR_RECT) + 0x4` up into the delay-slot NOP here.
That way you can remove the other one later on after `3:` too.

include/rsp_rdpq.inc
```diff
@@ -674,6 +674,28 @@ rdpq_update_fillcopy:
...
+
+    # Only update the interlacing info in low word if we've sent an rdp interlacing command
+    lw a2, %lo(RDPQ_SCISSOR_RECT) + 0x4
+    li t6, ~(3 << 24)
+    and a2, a2, t6
+    or a1, a2, a1
+    sw a1, %lo(RDPQ_SCISSOR_RECT) + 0x4
+    j 1f
+    nop
```
HailToDodongo: the `sw a1, %lo(RDPQ_SCISSOR_RECT) + 0x4` can be put in the delay-slot NOP

include/rsp_rdpq.inc
```diff
@@ -674,6 +674,28 @@ rdpq_update_fillcopy:
...
+3:  
+    # Load the interlacing info from the state info if we've sent a scissor rect command
+    lw a2, %lo(RDPQ_SCISSOR_RECT) + 0x4
+    li t6, (3 << 24)
+    and a2, a2, t6
```
HailToDodongo, line 693: You should be getting a stall here due to the load not spacing 2 cycles.
(check with `emux_trace_start` & `emux_trace_stop`)

You should be able to move the `sw a0, %lo(RDPQ_SCISSOR_RECT) + 0x0` or something else upwards to avoid it

(if the above change is done, i may work though)

include/rsp_rdpq.inc
```diff
@@ -674,7 +674,26 @@ rdpq_update_fillcopy:
     #############################################################
     .func RDPQ_WriteSetScissor
 RDPQ_WriteSetScissor:
+
```
rasky, line 677: What is the reason you put the change here, instead of in RDPQCmd_SetScissorEx? RDPQ_WriteSetScissor is called any time a ucode needs to reset the scissoring (eg: rdpq_mode has to call it when it switches cycle type). So its main use is "reapply the current scissor settings".

I believe the change should go to RDPQCmd_SetScissorEx instead, which is what is called to process the rspq command used to set scissoring (and now interlacing). That's probably where the special semantic should go.

include/rdpq.h
```diff
@@ -468,6 +468,55 @@ inline void rdpq_set_yuv_parms(uint16_t k0, uint16_t k1, uint16_t k2, uint16_t k
...
+ * 
+ * This interlacing mode saves half the RDP's fillrate since it skips half the lines
+ * to draw. This is very helpful for performance in high-res mode since the system
+ * is primairly fillrate constrained.
+ * 
+ * 
+ * @param[in]   draw_field      True if the RDP should only draw odd lines, otherwise only even lines are drawn
+ * 
+ */
+__attribute__((always_inline))
```
rasky: Can you change this to a standard inline?
Also please add the extern inline definitions at the bottom of rdpq.c.

include/rdpq.h
```diff
@@ -468,6 +468,55 @@ inline void rdpq_set_yuv_parms(uint16_t k0, uint16_t k1, uint16_t k2, uint16_t k
...
+ * to draw. This is very helpful for performance in high-res mode since the system
+ * is primairly fillrate constrained.
+ * 
+ * 
+ * @param[in]   draw_field      True if the RDP should only draw odd lines, otherwise only even lines are drawn
+ * 
+ */
+__attribute__((always_inline)) 
+inline void rdpq_enable_interlaced(int draw_field) { 
+    int32_t o = (draw_field == 1)? 1 : 0;
```
rasky: I would simply go with `draw_field & 1` which is faster, also I believe it has a more "expected behavior" if you pass a number like 122.

include/rdpq.h
thekovic: Standard `inline` keyword may be ignored by the compiler under certain circumstances. If the function is intended to be guaranteed to always be inlined, the `always_inline` attribute must be used.

rasky: I don't think this function is intended to be always inlined, at least no more than any other similar short wrapper in rdpq, and all of them are marked `inline`. We trust the compiler on this, we don't have any input that we should override its decisions on this function or any other of rdpq.

Frosty-J:
`FILTERS_DEDITHER` does not seem to work when `rdpq_enable_interlaced()` is used. Instead of dedithering anything, it just darkens the image slightly. Can this be avoided?

rasky:
I don't think we can use dedithering with rdpq-level interlacing; dedithering is a filter that uses a kernel of a few pixels around each one; in case of an interlaced setup, that'd affect lines of different fields and thus different frames.

We can try to warn about this in the future.

PR #721, opened 2025-05-19
CI fixes

networkfusion:
* Adds a new CI job for pre-push of docker images to test libDragon lib compatibility when building against a PR.
* Fixes latest image not being pushed when targeting trunk.
* Fixes rom builds failing.
* Fixes native toolchain not being released.
* Adds ARM docker image smoke test.
* Removes path diff action as no longer used.
* Works towards matrix's dependent on ARM packages (still commented out due to package not being released yet).

This should be applied to both preview and trunk.

Generated by AI:
```
# Libdragon Toolchain Multi-Platform Docker Workflow
#
# This workflow automates building, testing, and publishing a multi-platform Docker image for the
# libdragon N64 GCC toolchain. It ensures the image is up-to-date, works on both x86_64 and ARM64,
# and only pushes the 'latest' tag if all tests pass and the build is on the default branch.
#
# Workflow Overview:
# ----------------------------------------------------------------------------------------------
# 1. On PRs: Builds and tests the Docker image for both x86_64 and ARM64, but does NOT push.
# 1a. Smoke-tests the image by building libdragon (including examples and tests).
# 2. On non-PRs: Builds and pushes platform images by digest for both architectures.
# 3. Merges platform images into a multi-platform manifest and pushes it to the registry.
# 4. Smoke-tests the multi-platform image by building libdragon (including examples and tests)
#    on both architectures.
# 5. If all tests pass and the build is on the default branch, tags the image as 'latest'.
#
# Key Features:
# - Multi-platform: Supports both x86_64 and ARM64 builds and tests.
# - Safe publishing: Only pushes 'latest' if all builds and tests succeed on the default branch.
# - Artifacts: Uploads digests and built ROMs for traceability and debugging.
# - Lowercased repo name: Ensures Docker image names are always lowercase for registry compatibility.
```

```
# Libdragon Library & ROMs Build/Test Workflow
#
# This workflow ensures that the libdragon library, its examples, and test ROMs build correctly
# using the latest toolchain and Docker images. It runs on both x86_64 and ARM64 platforms,
# and uploads the resulting ROMs as artifacts for inspection.
#
# Workflow Overview:
# -----------------------------------------------------------------------------------------------
# 1. smoketest-multiplatform-docker-images:
#    - Runs on both x86_64 and ARM64.
#    - Pulls the latest multi-platform Docker image from the registry.
#    - Builds libdragon, examples, and tests inside the container.
#    - Uploads all built ROMs as artifacts.
#
# 2. smoketest-linux-packages:
[...]

PR #727, opened 2025-05-30
Improve CI for build-tool-windows.yml

networkfusion:
* Improve naming of workflow to mention msys2
* Ensure job targets a known working windows runner (2022), rather than latest to ensure that the upcoming windows runner (2025) does not break things. 

split out from https://github.com/DragonMinded/libdragon/pull/721

### DIFF
--- a/.github/workflows/build-tool-windows.yml
+++ b/.github/workflows/build-tool-windows.yml
@@ -1,12 +1,6 @@
 name: Build-Tools-msys2-Windows
 
-on:
-  push:
-    paths:
-      - 'tools/*'
-  pull_request:
-    paths:
-      - 'tools/*'
+on: [push, pull_request]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-tool-windows.yml
```diff
@@ -1,10 +1,16 @@
...
+name: Build-Tools-msys2-Windows
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - 'tools/*'
+  pull_request:
+    paths:
+      - 'tools/*'
```
rasky: this is wrong because tools include also files from src

rasky (changes requested):

.github/workflows/build-tool-windows.yml
networkfusion: reverted.

PR #731, opened 2025-06-17
Fix audioconv64 build on macOS using GCC

meeq:
`make tools` fails on macOS when building with GCC 15:
```bash
export CC=/opt/homebrew/bin/gcc-15 CXX=/opt/homebrew/bin/g++-15
```

```
gcc-15 (Homebrew GCC 15.1.0) 15.1.0
Copyright (C) 2025 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
In file included from audioconv64/libsamplerate.c:4:
audioconv64/libsamplerate/src_sinc.c:167:27: error: expected ';' before 'const'
  167 | LIBSAMPLERATE_DLL_PRIVATE const char*
      |                           ^~~~~
audioconv64/libsamplerate/src_sinc.c:186:27: error: expected ';' before 'const'
  186 | LIBSAMPLERATE_DLL_PRIVATE const char*
      |                           ^~~~~
In file included from audioconv64/libsamplerate/samplerate.c:19,
                 from audioconv64/libsamplerate.c:3:
audioconv64/libsamplerate/common.h:52:37: error: unknown type name '__private_extern__'
   52 |   #define LIBSAMPLERATE_DLL_PRIVATE __private_extern__
      |                                     ^~~~~~~~~~~~~~~~~~
audioconv64/libsamplerate/src_sinc.c:260:1: note: in expansion of macro 'LIBSAMPLERATE_DLL_PRIVATE'
  260 | LIBSAMPLERATE_DLL_PRIVATE SRC_STATE *
      | ^~~~~~~~~~~~~~~~~~~~~~~~~
```

Root cause appears to be that libdragon's vendored checkout of libsamplerate does not set the `HAVE_VISIBILITY` preprocessor definition when compiling, and [GCC on macOS does not support `__private_extern__`](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106672)

### DIFF
--- a/tools/audioconv64/libsamplerate.c
+++ b/tools/audioconv64/libsamplerate.c
@@ -1,3 +1,8 @@
+#ifndef __MINGW32__
+// Enable usage of visibility attributes for GCC and Clang, except when using MinGW
+#define HAVE_VISIBILITY
+#endif
+
 #include "libsamplerate.h"
 
 #include "libsamplerate/samplerate.c"


rasky:
Can you check if simply adding a `#define HAVE_VISIBILITY` at the top of libsamplerate.c is enough to fix the problem? I'm pretty sure the config check is not worth it as attribute hidden was added in GCC 4 in 2005.

Moreover, I'd rather to also avoid simply `CFLAGS += -DHAVE_VISIBILITY` in our Makefile as it's something related to the configuration of the vendored library; the problem doesn't exist upstream because they run autoconf, so they assume a config.h exists that would have `#define HAVE_VISIBILITY` on all compilers in the last 20 years. We don't commit a config.h so our "config.h" can just be manually handled in libsamplerate.c.

meeq:
Unfortunately, mingw does not support the visibility attribute, so we can't just hard-code it to be enabled ([which is what I tried first](https://github.com/DragonMinded/libdragon/actions/runs/15715349384/job/44283576710#step:6:65)).

I also didn't want to modify our vendored checkout of libsamplerate so that it would be easier to pull changes from upstream.

I could simplify this check so that we set `HAVE_VISIBILITY` for non-mingw builds.

meeq:
@rasky Updated -- I kept the define in the Makefile because we need the `-Wno-error=attributes` for MinGW

rasky:
Ah mingw...

What about this? It works on all the configurations I can check:

```patch
diff --git i/tools/audioconv64/libsamplerate.c w/tools/audioconv64/libsamplerate.c
index 9e0ad9044..7824cbe76 100644
--- i/tools/audioconv64/libsamplerate.c
+++ w/tools/audioconv64/libsamplerate.c
@@ -1,3 +1,7 @@
+#ifndef __MINGW32__
+#define HAVE_VISIBILITY
+#endif
+
 #include "libsamplerate.h"

 #include "libsamplerate/samplerate.c"
 ```

btw libsamplerate.c is not vendored

PR #736, opened 2025-07-07
Update joybus_rtc.c by renaming debugf to tracef

networkfusion:
Rename debugf to tracef and add defines.

When debugging on a Summercart64, there was some noise produced from the joybus_rtc source. This removes it when un-needed, but keeps the option for when ED64 (or other flashcarts, emulators and such) require further debugging.


networkfusion:
@meeq @rasky

PR #749, opened 2025-08-17
[rdpq_tex] never use block loads for YUV

pekkavaa:
The YUV_FMT16 pixel format needs special handling at load time which only LOAD_TILE can do. So texload_set_rect now checks for that case.


rasky:
Added more complete support to rdpq_tex (including LOAD_BLOCK) in 519106b0d2e9725ea0012a73d5054848d7c1369b

PR #754, opened 2025-08-23
Enable positional arguments in newlib build

gamemasterplc:
Requires toolchain rebuild to see benefits of.


rasky:
We measured an increase of run time for printf/sprintf calls, even those not using positional arguments, in addition to an increase of text size. Moreover, we're planning a switch from newlib to picolibc so we should probably reevaluate the performance implications there.

PR #757, opened 2025-08-27
Retry accessory read/write that returns NO_PAK status

meeq:
Update the meaning of `JOYBUS_ACCESSORY_IO_STATUS_NO_PAK` to indicate any of the following:

* Nothing is connected to the accessory port.
* The accessory command address checksum was invalid.
* A new accessory has been connected since the last identify command. 

Attempt to retry accessory read/write commands that encounter `JOYBUS_ACCESSORY_IO_STATUS_NO_PAK` once before treating it as an accessory disconnect event in case there was a transmission error.

### DIFF
--- a/include/joybus_accessory.h
+++ b/include/joybus_accessory.h
@@ -26,9 +26,16 @@ typedef enum
     JOYBUS_ACCESSORY_IO_STATUS_OK        = 0,
     /** @brief No N64 controller is connected. */
     JOYBUS_ACCESSORY_IO_STATUS_NO_DEVICE = -1,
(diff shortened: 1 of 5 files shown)


meeq:
Based on research and discussion in N64brew Discord:
The checksums are calculated by NUS-CNT, which has no idea whether the accessory actually accessed the correct address or sent/received the correct data over the accessory port. The address checksum and data CRC are for validating the Joybus serial transmission, not the accessory port parallel data transfers.

An accessory command responding with NO_PAK status probably worth retrying _once_, but if that retry fails we can assume the more-likely scenarios of the Pak having been disconnected.

Issue #753, opened 2025-08-21
vi_init could turn on display sync


rasky:
Currently, `vi_init()` leaves the TV not synced because it doesn't actually turn on the VI (that is `VI_CTRL.TYPE` is 0). This issue proposes to change the behavior so that `vi_init()` and thus also `vi_reset()` actually turns on the TV sync (activating blank mode though, so that no framebuffer is actually fetched and memory bandwidth is not impacted).

 * This would cause VI interrupts to start triggering. In turn, it would allow the `joypad` library to call `vi_init()` to make sure that happens, while right now it depends on vblank interrupts to exist without being able to guarantee that's the case.
   *  While we could use timer interrupts to poll joypads, this would cause variable input latency, at least for fixed framerate games.
 * Currently there is no `vi_close()`. We could add a `vi_close()` for symmetry, that also shuts down the VI. 
   *  Notice though that display_close() will not want to call `vi_close()`, because `display_close()` must leave the sync on. Calling `display_close()` is still the only way to change framebuffer resolution using the display API, and we don't want that to cause TV sync loss. So this will never change.
 * We should document that `vi_init()` leaves blank mode set. `vi_show()` disables it automatically, but `vi_set_origin()` doesn't. So this is a breaking API change (though VI library is still preview-only).

rasky:
Implemented in: https://github.com/DragonMinded/libdragon/commit/e71ca1a6b930ab332df21d7d9d41a49e8b583a61

PR #758, opened 2025-08-28
Prefer timer callback for Joypad subsystem polling

meeq:
As discussed in N64brew Discord: https://discord.com/channels/205520502922543113/768169699564453910/1408136313101619260


rasky:
We discussed this on Discord but we ended realizing that the timer will drift compared to vblank since they are different clock sources, and eventually the input latency will change. Probably not worth.

There is underlying issue that I'd like to solve: the joypad library is silently not working when display is not initialized. We have some ideas here:
https://github.com/DragonMinded/libdragon/issues/753

After we make vi_init() turn on the VI, we can have `joypad_init()` call `vi_init()` and the issue should be solved.

PR #763, opened 2025-09-02
mkfont: Support custom axis values in variable fonts

sp1187:
This feature allows the developer to specify custom parameters for variable fonts in mkfont using the new `--var-axis` flag, e.g. `--var-axis wght=800` for very bold text or `--var-axis wght=100` for very thin text.

Example below with the same font ([Science Gothic](https://github.com/googlefonts/science-gothic)) with different custom slants, weights and widths:
Skärmbild_20250902_224659
Skärmbild_20250902_225003

### DIFF
--- a/tools/mkfont/mkfont.cpp
+++ b/tools/mkfont/mkfont.cpp
@@ -17,6 +17,7 @@
 #include <sys/stat.h>
 #include <assert.h>
 #include <vector>
+#include <unordered_map>
 #include <unordered_set>
 
 #include "../../include/surface.h"
@@ -54,6 +55,7 @@ float flag_ttf_char_spacing = 0;
 float flag_ttf_aspect_ratio = 1.0;
 tex_format_t flag_bmfont_format = FMT_NONE;
 std::unordered_set<uint32_t> flag_charset;
+std::unordered_map<uint32_t, float> flag_var_axis_values;
 
 typedef struct {
     const char *name;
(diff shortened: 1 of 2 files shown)


tools/mkfont/mkfont.cpp
```diff
@@ -382,6 +384,9 @@ void print_args( char * name )
     fprintf(stderr, "                             Default assumes 4:3 display ratio\n");
     fprintf(stderr, "   --format <format>         Specify the output texture format for color fonts.\n");
     fprintf(stderr, "                             Valid options are: RGBA16, RGBA32, CI4, CI8 (default: autoselect)\n");
+    fprintf(stderr, "   --axis <tag=value>        Override axis value of variable font\n");
+    fprintf(stderr, "                             (e.g., --tag wght=800))\n");
```
rasky:
```suggestion
    fprintf(stderr, "                             (e.g., --axis wght=800))\n");
```

tools/mkfont/mkfont.cpp
```diff
@@ -54,6 +55,7 @@ float flag_ttf_char_spacing = 0;
 float flag_ttf_aspect_ratio = 1.0;
 tex_format_t flag_bmfont_format = FMT_NONE;
 std::unordered_set<uint32_t> flag_charset;
+std::unordered_map<uint32_t, float> variable_axis_values;
```
rasky: We call `flag_` all variables set via command line.

tools/mkfont/mkfont.cpp
```diff
@@ -659,7 +680,7 @@ int main(int argc, char *argv[])
         
         int ret;
         if (strcasestr(infn, ".ttf") || strcasestr(infn, ".otf")) {
-            ret = convert_ttf(infn, outfn, flag_ranges);
+            ret = convert_ttf(infn, outfn, flag_ranges, variable_axis_values);
```
rasky: In general I prefer not to pass variables set via flag around but keep them global, otherwise basically 99% of the code in a tool gets littered with flag passing.

Yes I now this contradicts flag_ranges being passed; that is a mistake that should also be fixed. All other flags don't get passed around.

rasky (commented):
Thanks! mkfont is really growing lots of TTF features, approaching professional tools!

tools/mkfont/mkfont.cpp
sp1187: Renamed to --var-axis.

PR #764, opened 2025-09-06
joybus: refactor to split off detection system from joypad

rasky:
This PR splits off the background detection system from joypad, and move it into joybus.

This creates a clear, lower-level API to detect plugging / unplugging of any joybus devices to the 5 ports (the fifth channel has only been used until now for fixed devices like eeprom or RTC, it doesn't mean that future flashcart couldn't allow to hotplug those or other devices). 

Somebody willing to develop a new joybus device (not joypad related) can now use the joybus API as a base layer to implement a module to drive it. Using detection callbacks, the library can also be automatically informed automatically whenever the device is plugged/unplugged, without having to implement its own polling.

This also means that the joypad module is now more clearly only related to joypads or joypad-like devices (mouse), plus of course the whole pak ecosystem that builds upon the N64 controller.

Followups:
 • We should really come up with a nice API to build and parse PIF packets (multiple channels). The structures used by joypad.c somehow simplify building a single joybus command but not assembling or parsing the whole PIF packet. That would remove even more code from the joypad library, making it available.
 • Should we port also eeprom.c to the new joybus detection? Currently, eeprom_status() issues its own identify command but it could probably just rely on `joybus_get_identifer()` now. Also eeprom.c could use the previous API once it exists.
 • Should we deprecate `joypad_get_identifier`? It seems the main difference compared to `joybus_get_identifier` is that it returns the cold state instead of the hot state. But this said, now that there is a joybus API to read the low-level identifier, maybe `joypad_get_style()` is sufficient.

### DIFF
--- a/include/joybus.h
+++ b/include/joybus.h
@@ -101,13 +101,9 @@ typedef uint16_t joybus_identifier_t;
  * @{
  */
 /**
- * @brief Joybus identifier for an unknown or malfunctioning device.
- */
-#define JOYBUS_IDENTIFIER_UNKNOWN               0x0000
-/**
  * @brief Joybus identifier for a port with no device connected.
  */
-#define JOYBUS_IDENTIFIER_NONE                  0xFFFF
+#define JOYBUS_IDENTIFIER_NONE                  0x0000
 /**
  * @brief Joybus identifier for the Nintendo 64 voice recognition peripheral (NUS-020).
  * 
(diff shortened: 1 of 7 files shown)


meeq:
Another logical follow-up would be to implement a similar detection callback for accessories, but that's obviously a much bigger effort.

Hot-pluggable EEPROM and other "port 5" devices is an interesting use-case. There is currently no demand for it, but it's technically possible. The EEPROM code is quite old and definitely needs to be refreshed: it pre-dates asynchronous commands!

I'm in favor of deprecating `joypad_get_identifier` in favor of `joybus_get_identifier`: in general, game developers should be using `joypad_get_style` due to the complexity of GameCube joybus device identifiers.

PR #765, opened 2025-09-07
build: Support custom target triples with N64_TARGET variable

kaylagator:
This makes it possible to use libdragon with additional toolchains installed alongside the officially supported one. I have been using this to develop/test some GCC patches without clobbering my main libdragon install.

Also refactored n64dso to use the util functions instead of using its own copy.

### DIFF
--- a/boot/Makefile
+++ b/boot/Makefile
@@ -11,7 +11,8 @@ else
 endif
 
 N64_GCCPREFIX ?= $(N64_INST)
-N64_GCCPREFIX_TRIPLET = $(N64_GCCPREFIX)/bin/mips64-elf-
+N64_TARGET ?= mips64-elf
+N64_GCCPREFIX_TRIPLET = $(N64_GCCPREFIX)/bin/$(addsuffix -,$(N64_TARGET))
 N64_CC = $(N64_GCCPREFIX_TRIPLET)gcc
 N64_AS = $(N64_GCCPREFIX_TRIPLET)as
 N64_LD = $(N64_GCCPREFIX_TRIPLET)ld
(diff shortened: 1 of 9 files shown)


boot/Makefile
```diff
@@ -11,7 +11,8 @@ else
 endif
 
 N64_GCCPREFIX ?= $(N64_INST)
-N64_GCCPREFIX_TRIPLET = $(N64_GCCPREFIX)/bin/mips64-elf-
+N64_TARGET ?= mips64-elf
```
rasky, line 14: Maybe `N64_TRIPLET` is a better name? `N64_TARGET` seems correct with respect to the toolchain itself (build/host/target), but outside of that context the name "target" doesn't immediately suggest me that a triplet will be found here.

tools/common/utils.h
```diff
@@ -75,6 +75,35 @@ static const char *n64_toolchain_dir(void)
...
+        return NULL;
+    const char *target = getenv("N64_TARGET");
+    if (!target)
+        target = "mips64-elf";
+
+    int ret;
+    if (target[0])
+        ret = asprintf(&n64_gccprefix_triplet, "%s/bin/%s-", inst, target);
+    else
+        ret = asprintf(&n64_gccprefix_triplet, "%s/bin/", inst);
```
rasky, line 98: Is this fallback path needed, given the above default to `mips64-elf`?

rasky (commented):
Thanks patch looks ok, I wanted to just comment on naming for future readability.

boot/Makefile
kaylagator: My reasoning was clang and rustc do expose the target triples as a `--target=` option at runtime that matches the GCC build option.

tools/common/utils.h
kaylagator: This is for the case when the target is an empty string (which could possibly be used if the host is also a MIPS CPU)

kaylagator:
Also updated the gitignore because of some pesky binaries getting checked in

PR #768, opened 2025-09-11
wav64: fix VADPCM stereo playback when huffman is disabled; close #767

### LINKED ISSUE #767: Stereo VADPCM playback is distorted with huffman=false
When using `--wav-compress vadpcm,huffman=false` as opposed to `--wav-compress vadpcm,huffman=true` (or in short `--wav-compress 1`), playback for stereo files is distorted. It seems like the playback rate is halfed and/or samples play twice!? Hard to tell what's really going on.

Tested in `examples/mixertest/` by replacing `monosample8.wav` with [this one](http://phoboslab.org/crap/monosample8.wav) and changing compression to `--wav-compress vadpcm,huffman=false` in the makefile.

Worth noting that the same file plays just fine with `--wav-compress vadpcm,huffman=true` or `--wav-compress none`.

phoboslab:
Not completely sure that is the right thing to do, but accounting for stereo channels for `lseek()` seems to fix the distortion issue when huffman is disabled.


rasky:
thanks!

PR #776, opened 2025-09-27
cpakfs: fix get serial

bsmiles32:
* prefer returning 0 on success and negative value on error like other functions and suggested by its description
* fix comparison on fsid_read return value
* fix documentation of serial argument which is 24 bytes (and not 20)


rasky:
Thanks!

PR #777, opened 2025-09-29
cpakdemo: add a basic demo of new cpakfs API

bsmiles32:
This demo tries to demonstrate usage of most of the new cpakfs API. It is not meant to be a proper cpak manager.
It should be suitable as a replacement for both cpaktest and cpak-utest examples. These examples have not been removed, but can be removed in a following commit / PR depending on your feedback.
I've tried to make it handle not just regular 32K cpak but also those that support bank switching.
It has only been tested with ares for now.


examples/cpakdemo/cpakdemo.c
```diff
@@ -0,0 +1,630 @@
...
+		uint8_t buf[8];
+		char buf_str[20];
+		FILE* f;
+
+		// read first few bytes of file content and print some hex dump of it
+		sprintf(path, "%s%s", prefix, dir.d_name);
+		memset(buf, 0, sizeof(buf));
+		f = fopen(path, "rb");
+		if (f == NULL) {
+			strncpy(buf_str, "???", sizeof(buf_str));
```
rasky: I would suggest to assert here -- there is no reason why the filename returned by directory iteration shouldn't be reachable via fopen, it would point to a bug in cpakfs.

examples/cpakdemo/cpakdemo.c
```diff
@@ -0,0 +1,630 @@
...
+		return STATE_WAIT_ACTION;
+	}
+	printf("Found %d banks\n", banks);
+	console_render();
+
+	// Copy loop done in chunks that don't cross the bank boundary.
+	// Could be optimized further if we can guarantee that the number
+	// of bytes read is always equal to the number passed as argument (and not less).
+	// This is the case in the current cpak_write implementation but
+    // is not documented as such in the public API documentation.
```
rasky, line 316: Ah right this must be an oversight. There is no reason why cpak_write/cpak_read should return a number smaller than the requested amount.

examples/cpakdemo/cpakdemo.c
```diff
@@ -0,0 +1,630 @@
...
+		printf("(%d) %s\n", errno, strerror(errno));
+		print_actions();
+		console_render();
+		return STATE_WAIT_ACTION;
+	}
+
+	char path[255 + 8];
+	sprintf(path, "%s%s", prefix, fname);
+
+	err = remove(path);
```
rasky: Missing error check here

examples/cpakdemo/cpakdemo.c
```diff
@@ -0,0 +1,630 @@
...
+
+	// Build proper UTF-8 file name for our test file
+	const cpakfs_path_t fcomponents = {
+		.gamecode = "NXXX",
+		.pubcode  = "XX",
+		.filename = "\x2d\x1e\x2c\x2d", // "TEST" in N64 code page
+		.ext      = "\x1A",             // "A" in N64 code page
+	};
+	char fname[64];
+    cpakfs_path_format(&fcomponents, fname, sizeof(fname));
```
rasky: This function is for advanced situations where you need to get access to the low level representation of the filenames. For this specific example, you're just making up a name, so you can just do:

```
char *fname = "NXXX.XX-TEST.A";
```

rasky (commented):
Thanks! This is a great improvement over the current examples, and showcases the new API. I have a few comments

examples/cpakdemo/cpakdemo.c
bsmiles32: I've updated the copy loop to assume that we can copy a full bank at once.

examples/cpakdemo/cpakdemo.c
bsmiles32: I've removed the cpakfs_path_format as suggested. I wanted to showcase both cpakfs_path_{parse,format} functions, but didn't find a really convincing simple use case: I've thought of overriding the gamecode part of the filename with the value from the ROM header by doing a cpakfs_path_parse(fname, ...) ; path.gamecode = header.gamecode; cpakfs_path_format(path, newfname, ...). But I didn't felt like adding this would be worth, so I've kept it simple as suggested

bsmiles32:
Thanks for your feedback, I've updated the PR to take into account the various comments.
Should I also remove the other cpak examples to not confuse new users with "deprecated" functions / examples ?

examples/cpakdemo/cpakdemo.c
```diff
@@ -0,0 +1,620 @@
...
+
+    // Recommended prefix is 'cpakN:/' with N being the player number (1..4)
+    char prefix[8];
+    sprintf(prefix, "cpak%d:/", port + 1);
+
+    // Full name must respect the GAME.PU-filename.ext format
+    // To make gamecode match the one from ROM Header (NED\x00)
+    // we use the hex encoding of gamecode because NED\x00
+    // contains a \x00 which is not useable in C strings.
+    const char* fname = "4e454400.XX-TEST.A";
```
bsmiles32: I've changed the fname to make gamecode match the one from the ROM header.
Given that it contains a \x00 character, I've resorted to use the hex encoding for gamecode.
I've tried "NED\x01.XX-TEST.A" but that didn't seem to work (I thought it would because of \x01 special meaning in utf8/n64 conversion).

I think it is somewhat important to showcase proper gamecode usage for homebrews which will almost all use NED\x00 game code. What do you think ?

examples/cpakdemo/cpakdemo.c
rasky: I've been on the edge of removing them from the public API. cpaktool uses `cpakfs_path_format` just to obtain a human readable error message in case the filename is not formatted correctly, because `fopen()` doesn't have a way to go into details and just returns `EINVAL` for invalid filenames. Maybe I should more explicitly mark them as low level or something.

rasky:
> Thanks for your feedback, I've updated the PR to take into account the various comments. Should I also remove the other cpak examples to not confuse new users with "deprecated" functions / examples ?

Yes, please remove them

examples/cpakdemo/cpakdemo.c
rasky: As discussed on Discord, it doesn't seem a good idea to encourage developers to use the same gamecode for all homebrew games. It's true that homebrew games will use `NED` because of the homebrew header, so maybe we can just decouple that this time. 

In this specific case, you can use `DRAG.ON-cpakdemo` as filename (and using extension `.1`, `.2` etc. if you need multiple files), and the comment might mention to use a custom gamecode/pubcode for your own game.

bsmiles32:
PR updated with requested changes :) Thanks for your valuable input !

rasky (approved):
Thanks for the submission!

PR #782, opened 2025-10-14
Kernel ksemaphore and kqueue wait & isr

jkent:
I've implemented some features for ksemaphore and kqueue.  They have not been tested completely, but work for me in my use cases.  Feedback is appreciated.


jkent:
Sorry for the noise, I created the PR against the wrong branch.

PR #783, opened 2025-10-14
Kernel ksemaphore and kqueue wait & isr

jkent:
I've implemented some features for ksemaphore and kqueue. They have not been tested completely, but work for me in my use cases. Feedback is appreciated.

### DIFF
--- a/include/kqueue.h
+++ b/include/kqueue.h
@@ -63,7 +63,7 @@ void kqueue_put(kqueue_t *queue, void *element);
  *
  * @return true if the element was added to the queue, false otherwise
  */
-bool kqueue_put_isr(kqueue_t *queue, void *element);
+bool kqueue_try_put_isr(kqueue_t *queue, void *element);
 
 /**
  * @brief Try to add an element to the queue with timeout
--- a/include/kqueue.h
+++ b/include/kqueue.h
@@ -55,6 +55,16 @@ void kqueue_destroy(kqueue_t *queue);
 void kqueue_put(kqueue_t *queue, void *element);
 
 /**
+ * @brief Attempt to add an element to a queue from an interrupt context
+ *
+ * @param[in] queue 	Pointer to the queue structure
+ * @param[in] element	Pointer to the element to add
+ *
+ * @return true if the element was added to the queue, false otherwise
+ */
+bool kqueue_put_isr(kqueue_t *queue, void *element);
+
+/**
  * @brief Remove an element from the queue
  * 
  * This function removes an element from the queue. If the queue is empty,
--- a/include/kqueue.h
+++ b/include/kqueue.h
@@ -6,6 +6,7 @@
 #define LIBDRAGON_KERNEL_KQUEUE_H
 
 #include <stdbool.h>
+#include <stdint.h>
 
 /** 
  * @brief A thread-safe FIFO queue 
@@ -65,6 +66,21 @@ void kqueue_put(kqueue_t *queue, void *element);
 bool kqueue_put_isr(kqueue_t *queue, void *element);
 
 /**
+ * @brief Try to add an element to the queue with timeout
+ *
+ * This function adds an element to the queue. If the queue is full, the
+ * function will block until there is space in the queue or the timer
+ * expires.
+ *
+ * @param[in] queue 	Pointer to the queue structure
+ * @param[in] element	Pointer to the element to add
+ * @param[in] ticks 	Number of hardware ticks to wait
+ *
+ * @return true if the element was added to the queue, false on timeout
+ */
+bool kqueue_try_put(kqueue_t *queue, void *element, uint32_t ticks);
+
+/**
  * @brief Remove an element from the queue
  * 
  * This function removes an element from the queue. If the queue is empty,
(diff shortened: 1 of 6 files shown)


include/kqueue.h
```diff
@@ -54,6 +55,31 @@ void kqueue_destroy(kqueue_t *queue);
...
 
+/**
+ * @brief Attempt to add an element to a queue from an interrupt context
+ *
+ * @param[in] queue 	Pointer to the queue structure
+ * @param[in] element	Pointer to the element to add
+ *
+ * @return true if the element was added to the queue, false otherwise
+ */
+bool kqueue_put_isr(kqueue_t *queue, void *element);
```
rasky: In general, I'm a bit worried of adding explicit ISR variants APIs. It makes the API surface larger and more difficult to audit. I would prefer to just have the function do the right thing under interrupt, by switching codepath if it is currently running under interrupt (`exception_is_running()`).

src/kernel/kqueue.c
```diff
@@ -60,6 +62,36 @@ void kqueue_put(kqueue_t *queue, void *element)
...
+{
+    if (queue->mutex.counter == 0 && queue->count < queue->size) {
+        queue->buffer[queue->tail] = element;
+        queue->tail = (queue->tail + 1) % queue->size;
+        queue->count++;
+        __kcond_signal_isr(&queue->not_empty);
+        return true;
+    }
+    return false;
+}
```
rasky, line 78: Unless I'm missing something, this is not safe if an interrupt is triggered within a `kqueue_put`.

src/kernel/ksemaphore.c
```diff
@@ -55,3 +55,10 @@ void ksemaphore_post(ksemaphore_t *sem)
     kcond_signal(&sem->cond);
     kmutex_unlock(&sem->mutex);
 }
+
+void __kcond_signal_isr(kcond_t *cond);
+void ksemapore_post_isr(ksemaphore_t *sem)
+{
+    sem->count++;
+    __kcond_signal_isr(&sem->cond);
+}
```
rasky, line 66: This one instead seems safe because `ksemapore_post()` and `kcond_signal()` both work in an atomic context (with interrupts disabled).

rasky (changes requested):
Thanks! I have a few comments

include/kqueue.h
jkent: The counter argument to this is that with -ffunction-sections and -fdata-sections the linker can remove unused/dead code from the resulting binary, reducing its size.

src/kernel/kqueue.c
jkent: Its been a while I wrote this, but I believe queue->mutex.counter was how I made it ~~thread~~ interrupt safe.  I will review it again and report back.  Pretty sure is why the true/false is returned, since you can't lock a mutex within an ISR context without having deadlocks.

src/kernel/ksemaphore.c
jkent: 👍

src/kernel/kqueue.c
jkent: After looking at how [kmutex_lock](https://github.com/DragonMinded/libdragon/blob/16c20f6b4a1b7549828c52f32e867637cabd162d/src/kernel/kernel.c#L776) and [kmutex_unlock](https://github.com/DragonMinded/libdragon/blob/16c20f6b4a1b7549828c52f32e867637cabd162d/src/kernel/kernel.c#L889) work, it appears okay. It should never run code in the `if` statement unless the queue's mutex is in the unlocked state.

jkent (commented):
Thanks for your feedback.  I've responded to your inline comments.

src/kernel/kqueue.c
jkent: To make this more clear, the code within a [kqueue_put](https://github.com/DragonMinded/libdragon/blob/f1fbb26ef597bbe16b7dbb25a219ff780ec17245/src/kernel/kqueue.c#L51) is guarded by a mutex.

include/kqueue.h
rasky: We value correctness more than smaller codesize. Multithreading programming is extremely hairy and N64 will never require an incredible performance in this area: making more than a handful of threads is mostly useless.

Another thing that I've noticed is that `kqueue_put_isr` should instead be `kqueue_try_put_isr` because it can't obviously block.

One thing we can do for now is to keep the API separated, but add an assert at the top of each function to verify the usage wrt interrupt status. We can experiment with this and see how it goes.

src/kernel/kqueue.c
rasky: Thanks, this makes sense

rasky:
Build fails on Doxygen errors.

You should probably declare __kcond_signal_isr in kernel_internal.h with a small doxygen line, and use the declaration from that header where you need it.

include/kqueue.h
jkent: Understood.

It appears only `kqueue_try_put_isr` needs the assert since the mutexes themselves assert if within an ISR context, and the rest of the non-atomic functions are wrapped properly.

rasky (approved):
Thanks for your contribution, I'm merging this

PR #786, opened 2025-11-04
dragonfs: add a real stat() function instead of falling back to fstat

zeesworth:
Up until now, DragonFS did not have a proper implementation of the stat syscall, instead falling back to fstat. I've implemented a proper stat function for DragonFS with the goal of being able to query directories as well - this can be useful to query for the type of directory entry a given path is (directory, regular file)

This will also result in the stat call being slightly less wasteful, as it will no longer always call an allocation for the file handle.

To demonstrate this, I've updated dfsdemo to add a small info display for the currently selected file/directory, showing the absolute path, the kind of file (directory, regular file) and the file size.

image
image

### DIFF
--- a/examples/dfsdemo/dfsdemo.c
+++ b/examples/dfsdemo/dfsdemo.c
@@ -5,6 +5,7 @@
 #include <string.h>
 #include <malloc.h>
 #include <libdragon.h>
+#include <sys/stat.h>
 
 #define MAX_LIST            20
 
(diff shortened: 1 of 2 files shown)


rasky (approved):
Thanks this seems fine. The dragonfs syscalls have various issues and don't consistently set errno on error (they should). For instance in the stat() call you add, you shouldn't return DFS internal error codes, but rather then expected POSIX interface (either 0, or negative value -- but always set errno when we do). fat.c is a more modern, and correct implementation that can be taken as reference.

This said, this can be fixed in a followup that fixes all syscalls. I'll merge this once CI passes

PR #790, opened 2025-11-23
Introduce rdpq_xform API

gamemasterplc:
This is an API for 2D transformations and transformed primitives. This API is presently done CPU-side. The API can be optionally used for sprites by setting allow_xform to true in rdpq_blitparms_t during a call to rdpq_tex_blit or rdpq_sprite_blit.

### DIFF
--- a/include/rdpq_xform.h
+++ b/include/rdpq_xform.h
@@ -3,7 +3,7 @@
  * @brief RDP Command queue: 2D transformation API.
  * @ingroup rdpq
  * 
- * This file contains functions for 2D transfomations and transformed primitives.
+ * This file contains functions for 2D transformations and transformed primitives.
  */
 
 #ifndef LIBDRAGON_RDPQ_XFORM_H
--- a/include/rdpq_xform.h
+++ b/include/rdpq_xform.h
@@ -1,0 +1,156 @@
+/**
+ * @file rdpq_xform.h
+ * @brief RDP Command queue: 2D transformation API.
+ * @ingroup rdpq
+ * 
+ * This file contains functions for 2D transfomations and transformed primitives.
+ */
+
+#ifndef LIBDRAGON_RDPQ_XFORM_H
+#define LIBDRAGON_RDPQ_XFORM_H
+
+#include <stdint.h>
+#include "fgeom2d.h"
+#include "rdpq_tri.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Replace current transformation matrix with another 3x3 matrix.
+ *
+ * @param mtx   Pointer to 3x3 matrix to load
+ */
+void rdpq_xform_mtx_load(fm_mat3_t *mtx);
+/**
+ * @brief Multiply current transformation matrix by another 3x3 matrix.
+ *
+ * @param mtx   Pointer to 3x3 matrix to multiply by
+ */
+void rdpq_xform_mtx_mult(fm_mat3_t *mtx);
+
+/**
+ * @brief Replace current transformation matrix with the identity matrix.
+ *
+ * This clears any transformations that may have been applied.
+ */
+void rdpq_xform_identity(void);
+/**
+ * @brief Multiplies current transformation matrix by rotation matrix.
+ *
+ * @param theta     Counterclockwise angle to rotate by in radians
+ */
+void rdpq_xform_rotate(float theta);
+/**
+ * @brief Multiplies current transformation matrix by translation matrix.
+ *
+ * @param x     X component of translation vector
+ * @param y     Y component of translation vector
+ */
+void rdpq_xform_translate(float x, float y);
+/**
+ * @brief Multiplies current transformation matrix by scaling matrix.
+ *
+ * @param x     Scaling factor along X axis
+ * @param y     Scaling factor along Y axis
+ */
+void rdpq_xform_scale(float x, float y);
+/**
+ * @brief Push current matrix to matrix stack.
+ */
+void rdpq_xform_push(void);
+/**
+ * @brief Pop from matrix stack.
+ *
+ * This replaces the current matrix with the matrix one below it on the stack.
+ */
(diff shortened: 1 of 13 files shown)


src/rdpq/rdpq_tri.c
```diff
@@ -531,6 +532,77 @@ void rdpq_triangle_rsp(const rdpq_trifmt_t *fmt, const float *v1, const float *v
         (fmt->tex_tile & 7));
 }
 
+/** @brief Internal helper function for rendering transformed triangle */
+void __rdpq_triangle_rsp_xform(const rdpq_trifmt_t *fmt, const float *v1, const float *v2, const float *v3, fm_mat3_t *mtx)
```
rasky: Instead of duplicating `rdpq_triangle_rsp`, add the `mtx` argument to it, and then pass it NULL in the existing uses.

The function is already declared in `rdpq_internal.h` so you can then use it in rdpq_xform passing a matrix instead.

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
+/**
+ * @file rdpq_text.c
```
rasky: rdpq_form

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
...
+
+#include "debug.h"
+#include "utils.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdarg.h>
+
+//List of transformation types
+//This is ordered from most strict to least strict
+typedef enum {
```
rasky, line 26: Please us doxygen format here. Also maybe explain what "strict" means in this context.

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
...
+    //Initialize matrix stack to point to top
+    xform_stack_depth = 0;
+    curr_xform = &xform_stack[xform_stack_depth];
+    rdpq_xform_identity();
+}
+
+void __rdpq_xform_close(void)
+{
+    //Make sure that trying to push/load matrices causes crashed
+    curr_xform = NULL;
```
rasky, line 61: ... though `rdpq_xform_push` doesn't crash if this is NULL. Maybe add an explicit assert there to avoid misuses.

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
...
+}
+
+static void transform_point(fm_vec2_t *out, float x, float y)
+{
+    fm_vec2_t in = {{x, y}};
+    fm_vec3_t temp;
+    fm_mat3_mul_vec2(&temp, &curr_xform->mtx, &in);
+    out->x = temp.x;
+    out->y = temp.y;
+}
```
rasky, line 187: I think you can change this to return `fm_vec2_t` by copy, to make it a bit more readable. It shouldn't matter much performance wise.

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
...
+#include <stdlib.h>
+#include <stdarg.h>
+
+//List of transformation types
+//This is ordered from most strict to least strict
+typedef enum {
+    RDPQ_XFORM_IDENTITY, //No transformation
+    RDPQ_XFORM_RECT, //Transformation with only translation
+    RDPQ_XFORM_SCALERECT, //Any transformation that has no rotation or skew
+    RDPQ_XFORM_TRI //Any 2D transformation
```
rasky: These namings are derived from the type of primitive you will be able to emit depending on the transformation. I think this is a bit indirect way of thinking, as you should be describing the type of xform by itself.

Maybe rename as: NONE, TRANSLATE, SCALE, ALL

src/rdpq/rdpq_xform.c
```diff
@@ -0,0 +1,364 @@
...
+            transform_point(&p0, x0, y0);
+            transform_point(&p1, x1, y0);
+            transform_point(&p2, x1, y1);
+            transform_point(&p3, x0, y1);
+            
+            //Generate vertex buffers for pair of triangles
+            float v0[2] = { p0.x, p0.y };
+            float v1[2] = { p1.x, p1.y };
+            float v2[2] = { p2.x, p2.y };
+            float v3[2] = { p3.x, p3.y };
```
rasky: Can you use directly pN.v here, instead of declaring a different array?

src/math/fgeom2d.c
```diff
@@ -12,7 +12,7 @@ void fm_mat3_rotate(fm_mat3_t *out, float rotation)
     fm_mat3_mul(out, &rotation_matrix, out);
 }
 
-void fm_mat3_from_srt(fm_mat3_t *out, const fm_vec2_t *scale, float rotation, const fm_vec2_t *translate)
+void fm_mat3_from_rst(fm_mat3_t *out, float rotation, const fm_vec2_t *scale, const fm_vec2_t *translate)
```
rasky, line 15: Does this naming matches Tiny3D?

rasky (approved):
Thanks, this is a great contribution to libdragon! Looking forward to the RSP version later on :)

include/rdpq_xform.h
```diff
@@ -0,0 +1,156 @@
+/**
+ * @file rdpq_xform.h
+ * @brief RDP Command queue: 2D transformation API.
+ * @ingroup rdpq
+ * 
+ * This file contains functions for 2D transfomations and transformed primitives.
```
sp1187: transformations

src/rdpq/rdpq_tri.c
gamemasterplc: Function has been merged into rdpq_triangle_rsp.

src/rdpq/rdpq_xform.c
gamemasterplc: Documentation changed to use doxygen format and to use alternative phrasing.

PR #792, opened 2025-12-05
Portability fixes for BSD (and some less common Linux) systems

goshhhy:
this PR contains a few portability fixes made during an attempt to build libdragon's toolchain on FreeBSD.

---

bash cannot always be found at /bin/bash.

/usr/bin/env is commonly present on most unix systems. using it to
invoke bash is more portable, and is already used elsewhere in this
project. this is necessary on systems like NixOS and the BSDs.

---

the second commit adds a variable that is used to store how to invoke
GNU Make. this is necessary for systems such as the BSDs which
typically use their own make implementation that is not compatible
with the same extensions as GNU Make, and often install the GNU
one as "gmake".

for existing users, nothing changes. BSD users can now invoke with:
MAKE=gmake ./build-toolchain.sh

---

these are not enough to allow the toolchain to build to completion, but it does now get quite rather late into the build, rather than immediately failing.

### DIFF
--- a/src/rdpq/mkfontbuiltin.sh
+++ b/src/rdpq/mkfontbuiltin.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 # cd to the script directory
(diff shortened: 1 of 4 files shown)


rasky:
Thanks, I'll keep this pending because we have a large refactoring to build-toolchain.sh that is almost ready to be merged. We can apply the $MAKE change on top of that.

PR #804, opened 2026-01-02
Minor Refactor rtctest code style

networkfusion:
Ensures consitency of code style when using if statements.

- Adds curly braces to ensure meaning.
- aligns spacing between if and bracket in some circumstances.

makes it easier for AI and compilers.

### DIFF
--- a/examples/rtctest/rtctest.c
+++ b/examples/rtctest/rtctest.c
@@ -154,12 +154,12 @@ int main(void)
             if( pad_pressed.d_left )
             {
                 edit_mode = edit_mode << 1;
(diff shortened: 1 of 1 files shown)


networkfusion:
@meeq

@rasky why is it not possible to assign reviewers/labels!

rasky:
There are many one-line if statements in libdragon, and we're not going to add curly braces to all of them. That would be a coding style change to be more widely discussed. Libdragon doesn't even have a single coding style, and coding style discussions are always a mess. So I don't want to start here, from a random example.

PR #813, opened 2026-01-20
sram: Create module for helper functions to access cartridge SRAM save memory

thekovic:
I put together a simple module to help access (read/write) SRAM saves to improve SRAM "support" from "it's easy to do but figure it out yourself" to "just call these".

Untested so far; please help me test before merging.

Side note: I think it would be nice to move PI register definitions from dma.h into its own pi.h header and properly define all the registers, including DOM1/DOM2 setup registers. Ideally, we would also use the "access via volatile struct" pattern over preprocessor definitions but that would require a lot of refactoring all over the place so perhaps that ship has sailed. Let's discuss here or on Discord.

### DIFF
--- a/include/sram.h
+++ b/include/sram.h
@@ -13,13 +13,14 @@
 #define LIBDRAGON_SRAM_H
 
 #include <stddef.h>
+#include <stdbool.h>
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
-#define SRAM_ADDRESS 0x08000000
-#define SRAM_SIZE    0x00008000  // 32 KiB
+#define SRAM_ADDRESS 0x08000000 ///< Base address of SRAM in PI address space.
+#define SRAM_SIZE    0x00008000 ///< Size of SRAM (32 KiB).
 
 /**
  * @brief Initialize the SRAM subsystem
@@ -28,6 +29,15 @@ extern "C" {
  * in the cartridge. It must be called before any other SRAM functions are used.
  */
 void sram_init(void);
+
+/**
+ * @brief Detect if SRAM is present in the cartridge
+ * 
+ * This function checks if the cartridge has SRAM available for use.
+ * 
+ * @return true if SRAM is detected, false otherwise.
+ */
+bool sram_detect(void);
 
 /**
  * @brief Read data from SRAM
--- a/include/sram.h
+++ b/include/sram.h
@@ -20,7 +20,6 @@ extern "C" {
 #endif
 
 #define SRAM_ADDRESS 0x08000000 ///< Base address of SRAM in PI address space.
-#define SRAM_SIZE    0x00008000 ///< Size of SRAM (32 KiB).
 
 /**
  * @brief Initialize the SRAM subsystem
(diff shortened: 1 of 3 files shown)


include/sram.h
```diff
@@ -0,0 +1,74 @@
...
+void sram_init(void);
+
+/**
+ * @brief Detect if SRAM is present in the cartridge
+ * 
+ * This function checks if the cartridge has SRAM available for use.
+ * 
+ * @return true if SRAM is detected, false otherwise.
+ */
+bool sram_detect(void);
```
rasky: Like eeprom_detect, I would make this function return the size of SRAM. IIRC Everdrive has a special "1 Mbit SRAM" mode, so maybe we could find a way to detect that as well, and return the proper size. But even if we postpone this, the API is more general if we immediately return the size here.

include/sram.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include <stddef.h>
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define SRAM_ADDRESS 0x08000000 ///< Base address of SRAM in PI address space.
+#define SRAM_SIZE    0x00008000 ///< Size of SRAM (32 KiB).
```
rasky: I would maybe remove this because, while this is the size used by SRAM in official cartridges, I don't think we should hardcode this too much (Everdrive also has a 1 Mbit mode).

src/sram.c
```diff
@@ -0,0 +1,93 @@
...
+    // Convert to virtual address to attempt to perform direct I/O to SRAM.
+    volatile uint32_t* sram_ptr = VirtualUncachedAddr(SRAM_ADDRESS);
+    // Test read, save the existing data to restore it later.
+    uint32_t data_old = *sram_ptr;
+    // Write a test pattern to SRAM and check it.
+    *sram_ptr = test_pattern;
+    uint32_t data_new = *sram_ptr;
+    bool is_detected = (data_new == test_pattern);
+    // Restore the old data.
+    *sram_ptr = data_old;
```
rasky: You should actually use `io_read` / `io_write` here. They're the counterpart of `dma_read` / `dma_write` after all, they are defined in the same header file, and they're both missing a `pi_` prefix for historical reasons.

The code happens to work because of the specific pattern you are using (read / write / read / write) but it's fragile. I'd rather use the official PI I/O functions.

src/sram.c
```diff
@@ -0,0 +1,93 @@
...
+int sram_read(void* dst, size_t offset, size_t len)
+{
+    assertf(__sram_inited, "sram accessed, but sram_init() hasn't been called yet");
+
+    // Check if the read operation is within bounds.
+    if (offset + len > SRAM_SIZE)
+    {
+        errno = EINVAL;
+        return -1;
+    }
```
rasky, line 70: You can keep this (defining SRAM_SIZE within the .c file) until we get around implementing the Everdrive support.

rasky (approved):
I think this is a very good starting point. I have left a few comments for improvements

include/sram.h
thekovic: Addressed in https://github.com/DragonMinded/libdragon/pull/813/commits/9d1aa5abc94bf4775ae6949e6088557e579f2f96

include/sram.h
thekovic: Size removed from public API in https://github.com/DragonMinded/libdragon/pull/813/commits/9d1aa5abc94bf4775ae6949e6088557e579f2f96

src/sram.c
thekovic: Switched to libdragon's `io_` functions in https://github.com/DragonMinded/libdragon/pull/813/commits/3e2782224ae0ec68edb0d4906502d6dfacbc63ec

PR #820, opened 2026-01-26
Remove all deprecations warnings from existing examples

0xSmartis:
- Updating `controllertest.c` to latest joypad API
- Replaced deprecated joypad identifier call with joybus API in `microuiN64.c`

### DIFF
--- a/examples/controllertest/Makefile
+++ b/examples/controllertest/Makefile
@@ -4,6 +4,8 @@ all: controllertest.z64
 BUILD_DIR = build
 include $(N64_INST)/include/n64.mk
 
+CFLAGS += -Wno-deprecated-declarations
+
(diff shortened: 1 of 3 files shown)


rasky:
Thanks!

`controllertest.c` is designed to verify that we don't break the old controller.c API which is now deprecated. So it must continue using the old deprecated API. To remove warnings during builds, I think the best course of action would be to disable the deprecation warning while building this example, in case you want to attempt this. We should also probably add a comment in the example itself explaining this.

0xSmartis:
Thank you for the feedback.
I changed the controllertest example back to the Controller API and added a note at the beginning of the code.

I also extended the makefile with the flag `CFLAGS += -Wno-deprecated-declarations`. This should now only ignore the deprecation warnings but still showing errors when something else breaks.

Issue #561, opened 2024-05-28
Missing examples (good first issue!)


rasky:
If you want to help libdragon, both the trunk and preview branches are missing work on the examples that should be updated to either migrate them to the new modern APIs (like rdpq) or even rewritten / changed to totally new examples.

This is a list of things we would like to have, but feel free to experiment in these ares:

 * [x] Remove all deprecations warnings from existing examples
 * [ ] New example: minimal 2D game loop (eg: one sprite driven by the controller that can collect coins). Max 200 lines of code (keep it simple!)
 * [ ] New example: minimal initial menu of a game. A background with a logo, a simple 2/3 level menu with eg: audio settings, and a simple navigation with the controller. If possible, background music. Max 200 lines of code
 * [x] New example: simple model64 mesh viewer. Allow to browse models from ROM or SD (or at least, switch between them with L/R for instance), display the current model self-rotating on the screen, allow to switch between animations in the mesh using a simple onscreen text.

For all the examples, please make sure to either use original assets made by you, or assets licensed under CC0 or similar no-attribution license.

ghost:
I have a first attempt on the minimal 2D game loop here: https://github.com/kannoneer/libdragon/blob/gameloop/examples/gameloop/gameloop.c

I didn't submit a PR because a spritesheet didn't work with flipped rects due to https://github.com/DragonMinded/libdragon/issues/562

rasky:
> I didn't submit a PR because a spritesheet didn't work with flipped rects due to #562

This is fixed in the preview branch now

zk-tarts:
Hi! If I am wanting to try out making of one these examples, (lets say the mesh viewer), should I be forking from the trunk branch or the preview branch?

I'm brand new to this project so I am interested in creating some kind of example to help learn my way around this code.
Thanks!

rasky:
> Hi! If I am wanting to try out making of one these examples, (lets say the mesh viewer), should I be forking from the trunk branch or the preview branch?
> I'm brand new to this project so I am interested in creating some kind of example to help learn my way around this code.

For the model viewer, you have to use preview. Otherwise, you could use trunk. Either works anyway: if you send to preview, it will land on trunk when the feature it depends on are stabilized.
Thanks1

0xSmartis:
Hi everyone!
Quick question, Is this ticket still available, given that it hasn’t been updated in about a year?

[...]

PR #821, opened 2026-01-26
Add a simple model64 Mesh viewer

0xSmartis:
This PR adds basic Mesh Viewer, addressing the task from issue #561. 

It uses the OpenGL Implementation for rendering and has been tested on Ares (v147) emulator and on a SummerCart64, where SD card mesh file reading was verified.

If the 200 lines of code are a hard limit, I could reduced the code size probably by roughly 25 lines and removing the Shadow and Outline pass during the rending.

All  the `*.glb` models were made by me in Blender 4.5.2 LTS.

I'd appreciate any feedback you have!
_(I'm an OpenGL beginner, sorry)_

---

**Ares emulator (loading from ROM)**

---

**Real hardware with SummerCard64 (loading from SD)**


0xSmartis:
Also the build rom here: [meshviewer.z64.zip](https://github.com/user-attachments/files/24867587/meshviewer.z64.zip)

rasky:
Should we review this or are you planning to change it soon?

0xSmartis:
> Should we review this or are you planning to change it soon?

Sorry, this PR got a bit of feature creep for me and I started second-guessing whether it’s good enough 😅.

I fixed a few minor things to make GLTF + BIN work as well, but I think this should be sufficient as an example of how to load and play `*.model64` files.

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+        // Shadow-Pass
+        glPushMatrix();
+        glTranslatef(0.0f, -0.7f, 0.0f);
+        glScalef(1.1f, 0.02f, 1.1f);
+        glDisable(GL_CULL_FACE);
+        glEnable(GL_BLEND);
+        glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+        glTexEnvi(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_MODULATE);
+        glColor4f(0.0f, 0.0f, 0.0f, 0.35f);
+        model64_draw(model);
```
snacchus: All accesses to `model` are surrounded by null checks except this one. I assume that's accidental?

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+
+        glMatrixMode(GL_MODELVIEW);
+        glLoadIdentity();
+        glTranslatef(0.0f, 0.0f, -3.0f);
+        glRotatef(rotation, 0.0f, 1.0f, 0.0f);
+
+        // Shadow-Pass
+        glPushMatrix();
+        glTranslatef(0.0f, -0.7f, 0.0f);
+        glScalef(1.1f, 0.02f, 1.1f);
```
snacchus: If you disable the depth test for the shadow pass, you can even make the shadow completely flat using 0 as the Y-scale

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+        glDisable(GL_CULL_FACE);
+        glEnable(GL_BLEND);
+        glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA);
+        glTexEnvi(GL_TEXTURE_ENV, GL_TEXTURE_ENV_MODE, GL_MODULATE);
```
snacchus: This call to `glTexEnvi` is not necessary since that's the default setting anyway (same applies to the other calls, obviously)

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+        model64_free(model);
+        model = NULL;
+    }
+
+    const model_entry_t *entry = &models[model_index];
+    const char *path = entry->rom_path;
+    active_is_sd = false;
+
+    if (use_sd && sd_available && file_exists(entry->sd_path)) {
+        path = entry->sd_path;
```
snacchus: I was wondering what the intention is behind the SD-loading feature? It seems to be expecting the same list of files as in ROM. Is it just for demonstration purposes (as in, to show that loading from SD is possible)?

I'm fine with leaving it like this for now, but an idea for a later addition would be to actually iterate the contents of SD dynamically (perhaps in a certain directory only).

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+
+        glClearColor(0.243f, 0.25f, 0.33f, 1.0f); // BG color
+        glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
+
+        glMatrixMode(GL_MODELVIEW);
+        glLoadIdentity();
+        glTranslatef(0.0f, 0.0f, -3.0f);
+        glRotatef(rotation, 0.0f, 1.0f, 0.0f);
+
+        // Shadow-Pass
```
snacchus: Rendering a shadow and outline is interesting, but it might not work for all models. I would consider making it optional, either per model, or toggleable at runtime. But it's not required, since the list of models is fixed right now and they are known to work well with it.

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,273 @@
...
+            if (models[model_index].tint_mesh) {
+                if (active_is_sd) {
+                    glColor4f(0.30f, 0.454f, 0.6f, 1.0f);
+                } else {
+                    glColor4f(0.2313f, 0.427f, 0.384f, 1.0f);
+                }
+            } else {
+                glColor4f(1.0f, 1.0f, 1.0f, 1.0f);
+            }
+            model64_draw(model);
```
snacchus: As a suggestion: Have you tried playing around with lighting? For a example a three point lighting setup could help making details of the model more visible.

snacchus (changes requested):
Thank you for taking the time to contribute this!

I have some comments, most of which are more suggestions than anything. This is mostly solid already!

examples/meshviewer/meshviewer.c
0xSmartis: Yes, my bad. I overlooked this one. I’ll add the null check in the next commit.

examples/meshviewer/meshviewer.c
0xSmartis: Still getting up to speed with OpenGL, so this is helpful. Thanks!

examples/meshviewer/meshviewer.c
0xSmartis: > Allow to browse models from ROM or SD (or at least, switch between them with L/R for instance)

You are right. checking it dynamically and iterate it would probably a nicer way to implement it.
I took a very quick approach to see if its working.

If its okay, i would re-work this too?

examples/meshviewer/meshviewer.c
0xSmartis: I’m currently reading up on OpenGL’s lighting concepts and will try a three-point setup.

0xSmartis:
@snacchus Thanks for the review. 
I’ll aim to get these changes into my branch until the end of next week.

examples/meshviewer/meshviewer.c
snacchus: Sure, if you want to, go ahead!

0xSmartis: Stupid question @snacchus, but when I do everything dynamically i would need access to the internal values of `model64`, or? I see no other access to available animations and if a texture is loaded yet in the public api?
Is it a no-go when I reference `model64_internal.h` for this?

examples/meshviewer/meshviewer.c
0xSmartis: bevor all calls, not null check is done by wrapping it in `if(model)`

examples/meshviewer/meshviewer.c
0xSmartis: updated the code according to that

examples/meshviewer/meshviewer.c
0xSmartis: Reworked it to dynamic loading with `model64_internal.h`

examples/meshviewer/meshviewer.c
0xSmartis: Put basic three point lighting in [3f0665d](https://github.com/DragonMinded/libdragon/pull/821/commits/3f0665dd85054621115452b4dd4b34d633bde7f3)

examples/meshviewer/meshviewer.c
0xSmartis: Made it optional with C button

examples/meshviewer/meshviewer.c
0xSmartis: did remove the calls

0xSmartis:
Finished the recommend the changes from the code-review

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,409 @@
...
+    glMatrixMode(GL_MODELVIEW);
+    glLoadIdentity();
+
+    setup_lighting();
+    glDisable(GL_TEXTURE_2D);
+    glEnable(GL_DEPTH_TEST);
+    glEnable(GL_CULL_FACE);
+}
+
+int main(void)
```
snacchus: The main method has become quite long. For readability's sake I would split it up into multiple functions, like `init`, `update`, `render` etc.

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,409 @@
...
+{
+    debug_init_isviewer();
+    debug_init_usblog();
+
+    dfs_init(DFS_DEFAULT_LOCATION);
+    display_init(RESOLUTION_320x240, DEPTH_16_BPP, 2, GAMMA_NONE,
+                 FILTERS_DISABLED);
```
snacchus: `FILTERS_DISABLED` is buggy on NTSC hardware and will trigger an assert (I assume you tested on PAL hardware). I'd recommend using `FILTERS_RESAMPLE` instead for example.

examples/meshviewer/meshviewer.c
```diff
@@ -0,0 +1,409 @@
...
+
+static model_list_t *active_model_list(void)
+{
+    return (current_source == MODEL_SOURCE_SD) ? &sd_models : &rom_models;
+}
+
+static void load_current_model(void)
+{
+    if (model) {
+        model64_free(model);
```
snacchus, line 207: Because rendering happens asynchronously in the background, the current model might still be in use for drawing the previous frame when this code is run. This could cause visual glitches or worse, crashes. To avoid this, there are two possibilities:
1. Use `rdpq_call_deferred`. This will call `model64_free` asynchronously as soon as all previous rendering is done. You would use it like this: `rdpq_call_deferred((void(*)(void*))model64_free, model);`
2. Call `rspq_wait()` before this line. This has a similar effect, but synchronously waits until all previous rendering is finished, blocking the CPU.

I tested both variants, and it seems there is some problem with the first one (it causes a crash). Not sure what it is, but in theory it should work. So I suggest using the second method for now. I wanted to mention the first one anyway for you information.

examples/meshviewer/Makefile
snacchus, line 1: There's a .gitignore file missing to ignore the filesystem directory that gets created during build.

snacchus (changes requested):
Some more things I noticed on my second round of review. Sorry I didn't catch them the first time!

examples/meshviewer/meshviewer.c
snacchus: The code looks unchanged. Is this intentional?

0xSmartis:
> Some more things I noticed on my second round of review. Sorry I didn't catch them the first time!

All good. I'll probably Sunday find some time to re-work it.

snacchus (approved):
Looks good, thank you!

examples/meshviewer/Makefile
0xSmartis: Added .gitignore in latest commit.

examples/meshviewer/meshviewer.c
0xSmartis: Thanks for the good explanation.
I added the call.

examples/meshviewer/meshviewer.c
0xSmartis: Yes, I have a PAL Nintendo 64.
Good to know.

examples/meshviewer/meshviewer.c
0xSmartis: Refactored it into sub methods.

PR #823, opened 2026-02-03
n64.mk: Add support for native Windows-style path in N64_INST

thekovic:
### Motivation
In anticipation of Pyrite 64 (which I would like to run as native Windows binary) and with growing frustration over how slow I/O between WSL filesystem and NTFS filesystem is, I have decided to switch to native Windows toolchain and libdragon install. However, I don't like the MSYS terminal and bash in general; and I love my riced Powershell setup. So I wanted to make libdragon work and build from Powershell. The Makefile is too tied to bash, though, so I decided to take the path of least resistance and just invoke MSYS's bash in-place from Powershell to execute problematic commands for me with

```
function Invoke-Bash {
    if (-not (Get-Command bash -ErrorAction SilentlyContinue)) {
        Write-Error "'bash' was not found. Please install MSYS2 and ensure bash.exe is on PATH."
        return
    }

    $convertedArgs = $args | ForEach-Object {
        '"' + ($_ -replace '\\', '/') + '"'
    }

    Start-Process bash -NoNewWindow -Wait -ArgumentList ($convertedArgs -join ' ') -Environment @{
        MSYSTEM = 'UCRT64'
    }
}

Set-Alias msys Invoke-Bash
```

This PR is the set of changes necessary to make my setup work and I believe it doesn't hurt other users so it would be nice if they could be upstreamed.

### Change notes
~~I removed the block with MSYS-specific checks from `build.sh` because both checks became problematic. The check for MinGW64 environment is too restrictive~~ and the check for backwards slashes is obsolete because backwards slashes are now handled and properly converted in `n64.mk`. Now users can set `N64_INST` to a format expected by Windows (and applied by default by the Windows Environment Variables tool) such as `C:\n64`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,23 +19,17 @@ if [[ -z ${N64_INST-} ]]; then
   exit 1
 fi
 
-# Run sanity checks for MinGW environment on Windows. There are a few common things
+# Run sanity checks for MSYS2 environments on Windows. There are a few common things
 # that can go wrong, and it's better to explicitly diagnose them here rather than
 # have the build fail mysteriously later.
-# Msys identifies itself via the OSTYPE variable as msys (before Feb 2025)
-# and "cygwin" (after Feb 2025). We want to detect both.
-if [[ $OSTYPE == 'msys'* || $OSTYPE == 'cygwin'* ]]; then
-  if [ "${MSYSTEM:-}" != "MINGW64" ]; then
-    # We only support building host tools via mingw-x64 at the moment, so
-    # enforce that to help users during installation.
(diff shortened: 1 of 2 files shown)


thekovic:
~~Since @rasky felt that allowing all MSYS2 environments is too lax, I opted for a blocklist check to throw out all the environments targetting Clang (that is CLANG32, CLANG64, CLANGARM64), and MSYS environment because the Cygwin C runtime hasn't been tested.~~

EDIT: Now changed back to a blocklist. UCRT64 is newly allowed. Users spinning up bash on their own (like I do in Powershell) need to spoof MSYSTEM environment variable to prove they know what they're doing.

PR #842, opened 2026-03-13
Restrict msym to exported symbols if externs file is provided

gamemasterplc:
This shrinks the msym file by a considerable amount especially in libdragon overlay examples.

### DIFF
--- a/n64.mk
+++ b/n64.mk
@@ -258,10 +258,16 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.cpp
 	@echo "    [DSOEXTERN] $@"
 	$(N64_DSOEXTERN) -o $@ $^ 
 	
-%.msym: %.elf
+%.msym:
 	@echo "    [MSYM] $@"
-	$(N64_DSOMSYM) $< $@
-    
+	EXTERNS_FILE="$(filter %.externs, $^)"; \
+	INPUT_FILE="$(filter %.elf, $^)"; \
+	if [ -z "$$EXTERNS_FILE" ]; then \
+		$(N64_DSOMSYM) "$$INPUT_FILE" $@; \
+	else \
+		$(N64_DSOMSYM) -e "$$EXTERNS_FILE" "$$INPUT_FILE" $@; \
+	fi
+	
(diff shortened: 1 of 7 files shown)


tools/n64dso/n64dso-msym.c
```diff
@@ -187,6 +193,30 @@ void process(char *infn, char *outfn)
...
+{
+    char *line_buf = NULL;
+    size_t line_buf_size = 0;
+    
+    FILE *file = fopen(infn, "r");
+    
+    
+    //Read symbol table output from readelf
+    while(getline(&line_buf, &line_buf_size, file) != -1) {
+        char *start_ptr = strstr(line_buf, "EXTERN(");
```
rasky, line 206: I think this should be fixed to avoid parsing comments. Linker scripts do support C style comments at least (`/* */`) and we should avoid parsing something like `/* EXTERN(foo) */`.

To avoid implementing a full linker script parser of course, maybe we could just check if the first non whitespace string is `EXTERN(`?

n64.mk
```diff
@@ -258,10 +258,16 @@ $(BUILD_DIR)/%.o: $(SOURCE_DIR)/%.cpp
...
+%.msym:
 	@echo "    [MSYM] $@"
-	$(N64_DSOMSYM) $< $@
-    
+	EXTERNS_FILE="$(filter %.externs, $^)"; \
+	INPUT_FILE="$(filter %.elf, $^)"; \
+	if [ -z "$$EXTERNS_FILE" ]; then \
+		$(N64_DSOMSYM) "$$INPUT_FILE" $@; \
+	else \
+		$(N64_DSOMSYM) -e "$$EXTERNS_FILE" "$$INPUT_FILE" $@; \
```
rasky, line 268: Seems like this wouldn't work if there are multiple external files?

gamemasterplc: The main elf linking would also break with multiple extern files.

tools/n64dso/n64dso-msym.c
gamemasterplc: That would not handle the multi-line comment case

PR #843, opened 2026-03-13
Eeprom async + eepfs integrity

### LINKED ISSUE #841: Add EEPROM async API
Since recent changes in Ares emulating EEPROM timings and the historical libdragon fix to implement correct EEPROM busy waiting (https://github.com/DragonMinded/libdragon/commit/159acf0fc76d01581b4776c24d73c32dbd0635aa), writing to EEPROM has become very slow, like it's always have been on a real repro cartridge. Games that were previously saving to EEPROM freely like every second or so are now faced with huge lags.

To provide a stopgap solution, we can add an asynchronous EEPROM writing API to both `eeprom.c` and `eepfs.c`.

Notice that reading is very fast and doesn't require a non blocking API.

rasky:
This PR implements:

* An eventual consistency implementation for the EEPROM API. The eeprom module now stores an internal cache of EEPROM contents (which is lazily populated). Reads are returned from the cache; writes update the cache and then start an asynchronous write to persist them. The API behaves just like a blocking API, so it's not changed, and is still quite fast even with new Ares with real eeprom timings. From an UX perspective, it's important to inform the user when a save is in progress to prevent an accidental shutdown. This can be done with the new `eeprom_is_busy` API.
* Fixes an hideous bug in EEPFS signature generation that caused the filesystem to report corrupted/invalid eeprom contents after minor random changes to the ROM.
* Adds two opt-in features to increase EEPFS integrity: 
  * Checksum: automatically calculate a checksum of each file on the filesystem on reads and writes, reporting errors in case of failure.
  * Backup: keep two copies of each file on the fileystem, alternating the one being written. In case of accidental corruption, the other copy will likely still be valid and not too old.

Fixes #841

### DIFF
--- a/examples/eepromfstest/eepromfstest.c
+++ b/examples/eepromfstest/eepromfstest.c
@@ -32,7 +32,7 @@ typedef struct
     int16_t map_pos_y;
     uint16_t max_hp;
     uint16_t current_hp;
-    uint8_t inventory[256];
+    uint8_t inventory[64];
 } game_save_state_t;
 
 static void press_a_to_continue(void)
@@ -133,7 +133,7 @@ static int validate_game_high_scores(char * path)
     int result;
 
     result = read_game_high_scores(path);
-    if ( result != 0 ) return result;
+    if ( result != 0 ) printf("Ivalid high scores!\n");
     press_a_to_continue();
 
     result = write_game_high_scores(path);
@@ -200,7 +200,7 @@ static int validate_game_settings(char * path)
     int result;
 
     result = read_game_settings(path);
-    if ( result != 0 ) return result;
+    if ( result != 0 ) printf("Invalid settings! (error: %d)\n", result);
     press_a_to_continue();
 
     result = write_game_settings(path);
@@ -268,7 +268,7 @@ static int write_game_save_state(char * path)
         -120,
         40,
         36,
-        { 1, 22, [254] = 52 }
+        { 1, 22, [63] = 52 }
     };
 
     printf( "Writing '%s'\n", path );
@@ -291,7 +291,7 @@ static int validate_game_save_state(char * path)
     int result;
 
     result = read_game_save_state(path);
-    if ( result != 0 ) return result;
+    if ( result != 0 ) printf("Invalid save state! (error: %d)\n", result);
     press_a_to_continue();
 
     result = write_game_save_state(path);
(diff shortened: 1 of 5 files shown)


examples/eepromfstest/eepromfstest.c
```diff
@@ -133,7 +133,7 @@ static int validate_game_high_scores(char * path)
     int result;
 
     result = read_game_high_scores(path);
-    if ( result != 0 ) return result;
+    if ( result != 0 ) printf("Ivalid high scores!\n");
```
meeq: Typo: “Invalid”

include/eepromfs.h
```diff
@@ -53,7 +55,7 @@ typedef struct eepfs_entry_t
      * 
      * A leading '/' is optional and will be ignored if set.
      * 
-     * The filesytem does not support entries for directories,
+     * The filesystem does not support entries for directories,
```
meeq, line 58: 😅 👍🏻

meeq (approved):

include/eeprom.h
```diff
@@ -78,26 +115,28 @@ size_t eeprom_total_blocks( void );
...
  * @brief Write a block to EEPROM.
  * 
- * Once a block is written, the EEPROM will be busy for up to 6 milliseconds.
- * This operation will wait for the EEPROM busy bit to clear before writing;
- * you may want to pause audio before calling this to prevent stuttering.
+ * This function writes a block of data to EEPROM (8 bytes). Most users
+ * will want to use #eeprom_write_bytes instead, which is more flexible.
+ *
+ * @note Writes are eventually consistent to the EEPROM, so they will be
+ *       persisted in background to the actual EEPROM. THe written data is
```
meeq: Minor typo: "THe" should be "The"

include/eeprom.h
```diff
@@ -30,8 +30,45 @@
...
+ * and flashcarts. To do so, add <code>N64_ROM_SAVETYPE=eeprom4k</code> or
+ * <code>N64_ROM_SAVETYPE=eeprom16k</code> to your Makefile. 
+ *
+ * This uses the advanced homebrew header which is supported by most emulators
+ * and flashcarts compatible with libdragon. It still advised to use the
+ * #eeprom_present and #eeprom_total_blocks functions to probe the EEPROM
+ * and gracefully handles the case where no EEPROM is present, even if it
+ * was requested.
+ *
+ * Cheap chinese flashcarts might not support the advanced homebrew header,
```
meeq: Suggestion: Remove "chinese" here. The concern here is not the country of origin, it's the level of sophistication of the flashcart.

src/joybus/eeprom.c
```diff
@@ -17,132 +72,349 @@
...
+        .command = JOYBUS_COMMAND_ID_IDENTIFY,
     } };
-    memcpy( cmd.send.data, src, EEPROM_BLOCK_SIZE );
-    joybus_exec_cmd_struct( EEPROM_PORT, cmd );
+    joybus_exec_cmd_struct(EEPROM_PORT, cmd);
+    switch (cmd.recv.identifier)
+    {
+        case JOYBUS_IDENTIFIER_CART_EEPROM_16KBIT:
+            eeprom_cached_type = EEPROM_16K;
+            eeprom_num_blocks = 256;
```
meeq, line 172: Developer experience suggestion: perhaps it would be more useful to allow developers to query the byte capacity of the EEPROM (as well as the block capacity).

Practically speaking, if we're encouraging use of eeprom_read_bytes and eeprom_write_bytes then the blocks are mostly just an implementation detail.

src/joybus/eeprom.c
```diff
@@ -17,132 +72,349 @@
...
+    eeprom_flush_resume(had_flush_in_progress);
+
+    memcpy(dest, &eeprom_cache[start], len);
+}
+
+void eeprom_write_bytes( const void * src, size_t start, size_t len )
+{
+    if (len == 0) return;
+    eeprom_cache_alloc_if_needed();
+    if (!eeprom_cache) return;
```
meeq: Should this assert?

src/joybus/eeprom.c
rasky: I would agree but is it worth an additional API that returns the same number, multiplied by 8?

src/joybus/eeprom.c
rasky: yes, changed

PR #845, opened 2026-03-18
fix: queue tracking state

HailToDodongo:


src/rspq/rspq.c
```diff
@@ -1460,6 +1470,9 @@ void rspq_queue_run(rspq_queue_t* q)
     // Switch back to the recording context.
     if (prev_recording)
         rspq_queue_switch(prev_recording);
+
+    // after the queue was executed, we can take its tracking state going forward
+    rdpq_tracking = q->rdpq_tracking;
```
rasky, line 1475: I think here you should probably also reset q->rdpq_tracking in case more commands are appended to the queue without clearing it.

PR #847, opened 2026-03-25
Fix rdpq paragraph nbytes issue 702

genecro:
This PR adds two pointers to the builder struct in rdpq_paragraph.c for keeping track of the last character that fit in the rectangular area as well as the last space that fit (for when WRAP_WORD is enabled).


rasky:
Can you please add a test for this behavior?

genecro:
I missed the case where the first word doesn't fit in the width of the layout area and an ellipsis is added. I just pushed a fix for that. Now I am working on adding tests, will update again when complete.

genecro:
Tests have been added

genecro:
I implemented the changes we talked about in discord (https://discord.com/channels/205520502922543113/974342113850445874/1488185811261329498). All characters in the line after an ellipsis are consumed

rasky:
I've noticed more edge cases that didn't seem to be covered by this PR. I attempted my own fix here:
https://github.com/DragonMinded/libdragon/pull/857

The code is quite convoluted, but it seems to pass all testcases now (in my PR, the tests are based on yours but they cover more cases). Please let me know what you think. You can also run your PR against the new tests in #857 and try to fix the other cases. Maybe your approach is better than mine: I can't tell for sure.

Issue #850, opened 2026-03-26
newlib-4.6.0.20260123/libgloss/mips/hal/crt0.S:92: Error: invalid operands `mtc0 $0,C0_CAUSE


joneqdaniel:
```sh
BINUTILS_V=2.46.0
GCC_V=14-20260320
NEWLIB_V=4.6.0.20260123
GMP_V=6.3.0
MPC_V=1.4.0
MPFR_V=4.2.2
```
```sh
mips64-elf-gcc -B/usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib_compile_target/mips64-elf/newlib/ -isystem /usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib_compile_target/mips64-elf/newlib/targ-include -isystem /usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib-4.6.0.20260123/newlib/libc/include -B/usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib_compile_target/mips64-elf/libgloss/mips -L/usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib_compile_target/mips64-elf/libgloss/libnosys -L/usr/src/toolchain/mips/n64/libdragon/joneqdaniel/libdragon/tools/toolchain/newlib-4.6.0.20260123/libgloss/mips    -DHAVE_ASSERT_FUNC -O2 -fpermissive -I. -I../../../../newlib-4.6.0.20260123/libgloss/mips/.. -I./.. -idirafter ../../../../newlib-4.6.0.20260123/libgloss/mips/../../include -DHAVE_ASSERT_FUNC -O2 -fpermissive -c ../../../../newlib-4.6.0.20260123/libgloss/mips/idtmon.S
/tmp/ccLtPtHM.s: Assembler messages:
../../../../newlib-4.6.0.20260123/libgloss/mips/hal/crt0.S:92: Error: invalid operands `mtc0 $0,C0_CAUSE'
../../../../newlib-4.6.0.20260123/libgloss/mips/hal/crt0.S:147: Error: invalid operands `mtc0 $10,C0_CONFIG,5'
../../../../newlib-4.6.0.20260123/libgloss/mips/hal/crt0.S:161: Error: invalid operands `mtc0 $2,C0_SR'
[...]

Issue #709, opened 2025-04-29
GCC-15.2.0 internal compiler error


jopadan:
**Version** 
- #748 

**Errors**
- [Build-Docker / Toolchain-Library-And-Examples (pull_request)](https://github.com/DragonMinded/libdragon/actions/runs/14737934856/job/41369681756?pr=700)

```cpp
[...]

PR #851, opened 2026-03-27
version bump with snapshot support and add tools/toolchain/* to ignore

joneqdaniel:
- Use gcc mirror with snapshots and get latest version from index.html
- Update all other packages versions to latest working release.
- Add permissions to decompress flags
- Add xz to install workflow

**Related**
- #850 
- #709
- #855


sp1187:
FYI, it is planned to replace newlib with picolibc in the near future, therefore newlib should probably not be updated further.

Also I don't see the need of using the latest snapshot version of GCC, not sure what new features is critically important for libdragon development and it could also be a maintenance nightmare as there will be many more different toolchain configurations installed in the wild each with different behaviors and bugs etc.

grawlinson:
> FYI, it is planned to replace newlib with picolibc in the near future, therefore newlib should probably not be updated further.

Sorry to bump in, but is this documented anywhere?

sp1187:
> > FYI, it is planned to replace newlib with picolibc in the near future, therefore newlib should probably not be updated further.
> 
> Sorry to bump in, but is this documented anywhere?

I guess there is not much public documentation up at the moment, but there is a pull request https://github.com/DragonMinded/libdragon/pull/535 and the subject has been discussed regularly on the N64brew Discord.

rasky:
Duplicate of #917

PR #854, opened 2026-03-30
Add "ignore TMEM" flag to mksprite

SpookyIluha:
Mainly to allow mipmaps for large sprites.
With this flag, no TMEM checks are performed during to .sprite conversion


rasky:
As discussed on Discord, mksprite is a bit confused right now between the "texture" codepath (must fit TMEM) and the "sprite" codepath (no limits).

I think it would be better to be explicit here and start asking the user to specify which behavior is expected. Most of mksprite currently assumes "sprite" rather than "texture" (mipmaps being the egregious exception) so I'd rather reverse the flag compared to this PR.

So I would suggest something like:

* `--mode TEXTURE|SPRITE` with default `SPRITE`
* `--texture` (default unspecified, though I don't very much like flags that you can't revert)
* `--force-tmem` 

or whatever else. I don't like much any of the above, maybe the first is the clearest after all.

PR #856, opened 2026-04-07
Implement workarounds for 8BitDo Retro Receiver v1.00 firmware

meeq:
[Discord thread](https://discord.com/channels/205520502922543113/1485671114737979453)

We have received reports that LibDragon does not work correctly with 8BitDo N64 Retro Receiver and N64 2.4G Wireless Controller Adapter running certain firmware versions.

image

Despite 8bitdo's firmware update log claiming that the N64 Retro Receiver firmware v1.08 is the "Factory firmware", some devices are shipping with a v1.0 firmware which incorrectly implements Joybus accessory behavior. This causes accessory detection to fail in LibDragon.

image

### Known issues with buggy 8BitDo receiver firmware:

* Joybus Identify accessory status behavior is non-standard:
  * Rumble Pak mode initially reports as `0x01` "Accessory Present" but all subsequent identify statuses will be `0x00` "Accessory Unsupported"
  * Switching Rumble Pak and Controller Pak mode transitions between `0x00` "Accessory Unsupported" and `0x01` "Accessory Present" status
* Accessory Write command data CRC responses is always `0x00`

These issues have been fixed in later versions, so the "best" solution here is to [update your receiver's firmware using the 8BitDo Ultimate Software V2](https://app.8bitdo.com/Ultimate-Software-V2/) but this PR attempts to provide a working experience for receivers running the outdated firmware.

### Mitigations

* Gracefully handle these specific nonsensical Joybus Identify status transitions
* During accessory detection, if an invalid data CRC of 0x00 is found, set a bit to ignore all write CRC errors for the device.
* Adjust randomly-generated CPak Probe Label so that its data CRC will never be `0x00`
* Improve handling of accessory disconnect to cancel pending read/write operations

**Test ROM:** [joypadtest.z64.zip](https://github.com/user-attachments/files/26522813/joypadtest.z64.zip)

### DIFF
--- a/src/joybus/joypad_accessory.c
+++ b/src/joybus/joypad_accessory.c
@@ -150,6 +150,13 @@ static bool joypad_accessory_check_write_crc_error(
         }
         case JOYBUS_ACCESSORY_IO_STATUS_BAD_CRC:
         {
+            if (accessory->ignore_write_crc)
+            {
+                // Special case: some buggy accessories always return an invalid data CRC.
+                // Ignore the CRC error and treat the write operation as successful.
+                accessory->error = JOYPAD_ACCESSORY_ERROR_NONE;
+                return false;
+            }
             size_t retries = accessory->retries;
             if (retries < JOYPAD_ACCESSORY_RETRY_LIMIT)
             {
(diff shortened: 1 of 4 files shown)


src/joybus/joypad_accessory.c
```diff
@@ -590,6 +597,14 @@ static void joypad_accessory_detect_write_callback(uint64_t *out_dwords, void *c
 
     const joybus_cmd_n64_accessory_write_port_t *cmdw =
         (void *)&out_bytes[port + JOYBUS_COMMAND_METADATA_SIZE];
+
+    if (cmdw->recv.data_crc == 0x00 && joybus_accessory_calculate_data_crc(cmdw->send.data) != 0x00)
+    {
+        // Special case: some accessories (notably 8BitDo receivers on old firmware) always return an invalid data CRC of 0x00.
+        // If this buggy behavior is detected, ignore all write CRC errors for this device from now on.
+        joypad_accessories_hot[port].ignore_write_crc = true;
+    }
```
rasky, line 606: Maybe better do this when joypad_accessory_check_write_crc_error returns an error? At least we don't slow down the main codepath for everybody else with an additional checksum calculation.

If that function returns true (failure) you could just check if the received checksum is 0x00 and try to switch to flag bit, instead of aborting.

meeq: Unfortunately, that approach is way messier. As it currently exists, `joypad_accessory_check_write_crc_error` has side-effects. On bad CRC it will return `true` and implicitly:
1. Retry the write command if the retry limit hasn't been hit
2. Reset the accessory state if retry limit is exceeded

The codepath you're questioning here will only get hit when the write data CRC is `0x00` which is not a common occurrence and should be very cheap to check.

meeq: Mitigating factors:
- `joypad_accessory_detect_write_callback` only fires for accessory writes during detection. The main codepath (`joypad_accessory_write_callback`) does not contain this check.
- C's `&&` operator is short-circuiting, so calculating the data CRC should only need to happen in the event of an expected checksum of `0x00`.
- We also now ensure that the CPak label probe will never checksum to `0x00`.

PR #857, opened 2026-04-07
rdpq_text: fix nbytes in all edge cases

### LINKED ISSUE #702: nbytes in rdpq_paragraph_build inaccurate
`nbytes` should hold the number of characters contained in the paragraph, but it seems to only get checked at escape sequences and newlines. If the string has no escape sequences or newlines, `nbytes` is equal to the length of the entire string, not just the portion that gets displayed in the paragraph.

rasky:
Fixes #702.

This is an alternative to #847 which should cover more edge cases.


rasky:
#847 is better:  less code and passes the same tests

PR #861, opened 2026-04-15
tools: Fix polyfill.h build on MSYS2 update from 2026-04-13

thekovic:
Update to MSYS2 headers from 2026-04-13 improved conformance with C23 and POSIX which means users building libdragon now get `strndup` provided by the environment. This affects both MINGW64 and UCRT64. This caused a compilation issue because `polyfill.h` defines its own implementation since it was previously missing. I removed the custom implementation to fix the issue.

> [!WARNING]
> People using MSYS2 installation older than 2026-04-13 will need to update their environments by running `pacman -Syu` from an MSYS2 terminal.

I also added a preprocessor guard around custom implementation of `tmpfile` so that when building with UCRT64, the standard implementation is used instead.

### DIFF
--- a/tools/common/polyfill.h
+++ b/tools/common/polyfill.h
@@ -40,7 +40,7 @@ extern "C" {
 // if typedef doesn't exist (msvc, blah)
 typedef intptr_t ssize_t;
 
-/* Fetched from: https://stackoverflow.com/a/47229318 */
+/* Implementation of 'getline' fetched from: https://stackoverflow.com/a/47229318 */
 /* The original code is public domain -- Will Hartung 4/9/09 */
 /* Modifications, public domain as well, by Antti Haapala, 11/10/17
    - Switched to getc on 5/23/19 */
--- a/tools/common/polyfill.h
+++ b/tools/common/polyfill.h
@@ -31,6 +31,8 @@
 #include <fcntl.h>
 #include <time.h>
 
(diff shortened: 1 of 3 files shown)


thekovic:
@rasky I did some thinking with morning and I came up with a solution to check for a pretty solid hint of the age of MSYS2 environment's age and plug it into the build process. With the latest commit added to this PR, users that haven't updated MSYS2 should still be able to compile Libdragon.

tools/common/polyfill.h
```diff
@@ -91,24 +93,35 @@ static ssize_t getline(char **lineptr, size_t *n, FILE *stream) {
     return pos;
 }
 
-/* This function is original code in libdragon */
```
Dragorn421, line 94: (this comment is relevant for copyright reasons, I wouldn't remove it)

thekovic: What copyright reasons? Can you explain please?

Dragorn421: For attribution, see the comments above `getline` detailing where it comes from, this comment here makes it clear that this strndup implementation is not covered by the getline comments

thekovic: A clarification to make sure that code in public domain is not misinterpreted as code... in public domain?

thekovic:
Please squash before merging

PR #870, opened 2026-04-26
Fix missing stdlib.h include

andmatand:
This fixes an error I was getting when trying to build pyrite64 on linux:
```
libdragon/include/debug.h:284:17: error: ‘abort’ was not declared in this scope
```

### DIFF
--- a/include/debug.h
+++ b/include/debug.h
@@ -9,6 +9,7 @@
 
 #include <stdbool.h>
 #include <stdio.h>
+#include <stdlib.h>
 
 #ifdef __cplusplus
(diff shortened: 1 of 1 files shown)


thekovic:
Confirmed this is an issue. It occurs when building some sort of Pyrite64's C++ wrappings. I think it's worth investigating where this should be fixed so I will send a report downstream.

thekovic:
Fixed in https://github.com/DragonMinded/libdragon/commit/b1011fe317767b1e959c9b0df2cdb1b559892dc9 , please close @rasky

PR #874, opened 2026-04-29
WIP: Lossy Sprite Compression

meeq:
Rasky and I have been working on implementing support for H.264-based lossy compression of sprites. I have a working example that demonstrates lossy-compressing a PNG at build-time and decompressing it at run-time into a LibDragon Sprite.

###  Public API
  - `lossysprite_load` — loads/decodes an LSPR file (an H.264 intra slice) into a regular `sprite_t`
  - `sprite_load` / `sprite_load_buf` sniffs the LSPR magic and delegates to `lossysprite_load`, so existing call sites work transparently
  - `rdpq_sprite_blit` knows how to draw the result
  - `rdpq_sprite_upload` asserts (single-TMEM upload is impossible for semi-planar YUV)

###  Decoder (`src/lossysprite.c`)
  - Decompresses the LSPR file into YUV planes at run-time using the existing H.264 decoder
  - Emits a semi-planar layout: a full-res Y plane (`FMT_I8`) followed by a half-res UV plane (`FMT_IA16`, U high / V low — matches what the RSP UV interleaver normally produces).

###  YUV blit path (`include/yuv.h`, `src/video/yuv.c`)
  - `yuv_tex_blit_semiplanar` skips RSP YUV interleaving and goes straight to RDP load+draw

###  Sprite integration (`src/sprite.c`, `src/sprite_internal.h`, `src/rdpq/rdpq_sprite.c`)
  - Added support for semi-planar YUV sprites to `rdpq_sprite_blit` to auto-configure YUV render mode
  - New flag `SPRITE_FLAG_YUV_SEMIPLANAR` (`0x0080`) used during sprite blitting
  - Repurposed the spare padding byte in `sprite_ext_t` for YUV colorspace

###  Example (examples/lossysprite/)
  - Minimal ROM: load a 640×480 background as .sprite, blit it via rdpq_sprite_blit

## Outstanding concerns
1. Allow opt-in for lossy decoding with `sprite_load` / `sprite_load_buf` so that the linker can strip out H264 code if unused.
2. `rdpq_set_mode` functions should not be called in `rdpq_sprite_blit`
    * Due to implementation details of lossy compression with H.264, it is more efficient to use YUV16 format for blitting, but this introduces a conflict: the RDP needs to be in YUV mode, but usage of `rdpq_sprite_blit` is not intended to touch render modes.


include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+ * `mksprite --lossy [--format=<X>]`. At load time the H.264 slice is fully
+ * decoded on RSP+CPU and the YUV reconstruction is converted into the
+ * requested target format, so the returned #sprite_t looks and renders like
+ * a normal sprite of that format.
+ *
+ * Target formats:
+ *
+ * - default / `--format=RGBA16` -> #FMT_RGBA16 sprite
+ * - `--format=RGBA32`           -> #FMT_RGBA32 sprite
+ * - `--format=UYVY`             -> #FMT_YUV16 sprite (packed 4:2:2)
```
rasky, line 17: FMT_YUV16 is a standard RDP texture format, and follows RDP format conversions of "NAME<bitsize>". I might be biased because I'm used to it, but I think it is preferable to introduce a new naming `UYVY` for a standard format.

include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+ * - `--format=NV12`             -> semi-planar 4:2:0 sprite: Y plane
+ *                                  (#FMT_I8) + interleaved UV plane
+ *                                  (#FMT_IA16, half height), tagged as
+ *                                  #FMT_YUV16 with the semi-planar layout
+ *                                  recorded in #sprite_ext_t::yuv_attrs
```
rasky, line 22: Do not put implementation details in lossysprite.h such as referring to how `#sprite_ext_t::yuv_attrs` is encoded or the fact that it even exists. You should put them in lossysprite.c. lossysprite.h must only comment the public API (and tools).

include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+ * through #yuv_tex_blit using the colorspace stored in the file header.
+ * Because semi-planar sprites split the image across two TMEM banks, they
+ * are not compatible with #rdpq_sprite_upload — they must be drawn via
+ * #rdpq_sprite_blit.
+ *
+ * Typical usage goes through #sprite_load, which sniffs the LSPR magic and
+ * dispatches here automatically once #lossysprite_init has registered the
+ * decoder. Call #lossysprite_decode_buf directly only when the encoded
+ * bytes are already in memory (e.g. embedded as a DSO data blob).
+ */
```
rasky, line 41: As mentioned above this whole initial doc is highly skewed towards implementation details. A few questions the initial docs should answer:

* What is a lossy sprite and what is its relation to a normal sprite? When is it useful to use it? Is it OK for 32x32 textures? Better for larger files?
* Does it use the same file .sprite or is it a different file?
* What's the relationship between the sprite.h API and the lossysprite.h API? Do I need to use both? One or the other? If they are alternative, there are pros and cons? More cases supported by one respected to the other?
* What are the performance characteristics of a lossysprite? Do I expect it to be slow at loading, slow at drawing, both or neither? Does it use RSP during loading or drawing? Is the loading blocked or asynchronous?
* What are the memory characteristics of a lossysprite? Is it fully decompressed at loading? Is it decompressed into RGBA or YUV at load time?

include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+
+
+/**
+ * @brief Register the LSPR decoder with the sprite loader.
+ *
+ * After this call, #sprite_load recognizes the `LSPR` magic and routes
+ * matching files through #lossysprite_decode_buf. Until then, loading an
+ * LSPR file via #sprite_load fails. Safe to call multiple times.
+ */
+void lossysprite_init(void);
```
rasky, line 67: Shoudl we account for multiple algorithms here? Should there be one init per algorithm eg: `lossysprite_init_lvl3()`?

include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * The buffer is only read during the call; the caller retains ownership
+ * and may free or reuse it as soon as this function returns. The returned
+ * sprite owns its pixel storage and must be released with #sprite_free.
+ *
+ * @param buf   Pointer to the encoded LSPR file contents
+ * @param sz    Size of @p buf in bytes
+ * @return      The decoded sprite (free with #sprite_free)
+ */
+sprite_t *lossysprite_decode_buf(const void *buf, int sz);
```
rasky, line 95: Maybe call this `lossysprite_load_buf` for symmetry with `sprite_load_buf`

include/lossysprite.h
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * Cheap header-only check: validates the LSPR magic and version. Returns
+ * false for non-LSPR data without asserting, so it is safe to use as a
+ * dispatch test on a buffer that may or may not be LSPR-encoded.
+ *
+ * @param buf   Pointer to the buffer to inspect (may be NULL)
+ * @param sz    Size of @p buf in bytes
+ * @return      true if @p buf is a recognized LSPR file, false otherwise
+ */
+bool lossysprite_is_encoded(const void *buf, int sz);
```
rasky, line 125: Is this function needed in the public API? What's the use case?

tools/mksprite/mksprite_lossy.cpp
```diff
@@ -419,6 +470,23 @@ extern "C" int mksprite_convert_lossy(
         return 1;
     }
 
+    // Snapshot the parameters as adjusted by x264_encoder_open. The "stillimage"
+    // tune drives f_psy_rd/f_psy_trellis up, which causes the encoder to silently
+    // shift i_chroma_qp_offset (encoder.c:1226-1230). For CRF mode without
+    // b_stitchable, x264 also writes pic_init_qp = SPEC_QP(i_qp_constant) — not
+    // the libdragon-conventional 26. Both values must travel out-of-band so the
+    // runtime decoder can apply the same dequant scale x264 used.
```
rasky, line 478: I wouldn't call 26 "libdragon-conventional". That was just one hardcoded value that I dumped from a single file that I had encoded and I was running test with. So my test file wa using QP=26 which is why the value was hardcoded there. I knew in the back of my mind it would have to be dynamic like this commit does, but didn't get around doing it. There was no intention of hardcoding 26 as some sort of good default.

include/sprite.h
```diff
@@ -289,6 +290,38 @@ bool sprite_fits_tmem(sprite_t *sprite);
...
+ *
+ * Semi-planar sprites cannot be uploaded to TMEM as a single texture
+ * (the Y and UV halves load into separate TMEM banks); use
+ * #rdpq_sprite_blit, which dispatches to #yuv_tex_blit internally.
+ *
+ * @param sprite        The sprite to inspect
+ * @return              True if the sprite is NV12 or NV16; false otherwise
+ *                      (including non-YUV sprites and packed UYVY)
+ */
+bool sprite_is_yuv_semiplanar(sprite_t *sprite);
```
rasky, line 323: On one hand, this seems a duplicate of sprite_get_yuv_format, since it basically tests two values of that list. On the other hand, there are many reasons why #rdpq_sprite_upload can fail (the biggest being that the sprite is too large). So I'm not sure if this function is really useful to add.

include/sprite.h
```diff
@@ -289,6 +290,38 @@ bool sprite_fits_tmem(sprite_t *sprite);
...
+ *
+ * Today this returns one of:
+ * - #YUV_UYVY for packed sprites (`--format=YUV16` in mksprite)
+ * - #YUV_NV12 for semi-planar 4:2:0 sprites (`--format=NV12` in mksprite_lossy)
+ * - #YUV_NV16 for semi-planar 4:2:2 sprites (`--format=NV16` in mksprite_lossy)
+ *
+ * @param sprite        The sprite to inspect (must be #FMT_YUV16)
+ * @return              The #yuv_format_t identifying the on-disk layout
+ */
+yuv_format_t sprite_get_yuv_format(sprite_t *sprite);
```
rasky, line 309: So to clarify, sprite_get_format() would always return `FMT_YUV16` as a way to say "this is a YUV sprite", and then you need `sprite_get_yuv_format` to extract the exact layout of the YUV bytes? 

I think the complexity here, semantically speaking, is that `FMT_YUV16` is actually a specific packed format supported by RDP (UYVU). So reusing it as a generic placeholder for "generic YUV" might be a bit confusing. I don't have a better suggestion.

Maybe we should mirror this in the mksprite command line though. Instead of having `--format=NV12`, we could have something like `--format=YUV16 --yuv=NV12` or `--yuv=UYVY`. Or maybe even a special syntax like `--format=YUV16:NV12`, with `--format=YUV16` being the same of `--format=YUV16:UYVY`.

src/rdpq/rdpq_sprite.c
```diff
@@ -114,11 +168,17 @@ int __rdpq_sprite_upload(rdpq_tile_t tile, sprite_t *sprite, const rdpq_texparms
     }
 
     if (__builtin_expect(set_mode, 1)) {
+        // For YUV sprites, configure the RDP YUV render mode + colorspace
+        // from the sprite's metadata. This must happen before mipmap/tlut
+        // tweaks below, since rdpq_set_mode_yuv resets the SOM.
+        if (sprite_get_format(sprite) == FMT_YUV16)
+            rdpq_sprite_set_yuv_mode(sprite);
+
```
rasky, line 176: I would avoid this part. I think rdpq_sprite_upload() shouldn't reset the reader mode like this, especially since you can't then do the counterpart for the standard mode anyway. I think this is something that should  be documented. and left as-is-

src/rdpq/rdpq_sprite.c
```diff
@@ -13,6 +13,59 @@
...
+    if (fmt == YUV_UYVY) {
+        // Packed: a single FMT_YUV16 surface from the sprite's pixel data.
+        frame.y = sprite_get_pixels(sprite);
+    } else if (sprite_is_yuv_semiplanar(sprite)) {
+        // Semi-planar: split the pixel buffer into Y + interleaved UV. The
+        // padded plane dimensions live in texparms (set by the producer).
+        int padded_w = (int)sx->texparms.s.translate;
+        int padded_h = (int)sx->texparms.t.translate;
+        uint8_t *base = (uint8_t*)sprite + sx->data_ptr;
+        size_t y_bytes = (size_t)padded_w * padded_h;
```
rasky, line 45: They are set by lossysprite.c, not mksprite. So that's not the producer.

include/yuv_format.h
```diff
@@ -0,0 +1,74 @@
...
+ * @brief YUV chroma subsampling and memory layout.
+ *
+ * Selects how the chroma planes of a #yuv_frame_t are laid out in memory.
+ * This determines what #yuv_tex_blit (and #yuv_blitter_run) expects to find
+ * in the frame's surfaces and whether an RSP-side interleave pass is needed
+ * before the RDP draw.
+ *
+ * The default value (0) matches the historical behavior, so existing callers
+ * that build a #yuv_frame_t with designated initializers do not need to set
+ * @ref yuv_frame_t::format explicitly.
```
rasky, line 27: This paragraph has a bit of recentism. In mksprite, our default is UYVY which is the hardware format; maybe we should set that as default value (0) even though existing source code might break (though it would break with an assert, and that's probably acceptable for the preview branch).

include/yuv_format.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+typedef enum yuv_format_e {
+    /** @brief 3-plane planar 4:2:0 (I420).
+     *
+     *  Y, U, and V are three separate #FMT_I8 surfaces; U and V are at half
+     *  width and half height. RSP interleaves U+V into a temporary IA16
+     *  buffer before the RDP draw, so #yuv_init must have been called. */
```
rasky, line 34: I don't think it's now reasonable to require a user to know whether yuv_init must tbe called or not depending on the assets (considering that the YUV format is now embedded in the assets). 

I think we should always call it as part of lossysprite_init() just like fmv.c calls it.

rasky (commented):
Did a first round only on public headers, APIs, and docs.

include/rdpq_sprite.h
```diff
@@ -122,6 +123,26 @@ int rdpq_sprite_upload(rdpq_tile_t tile, sprite_t *sprite, const rdpq_texparms_t
...
+ */
+yuv_blitter_t rdpq_sprite_yuv_blitter_new(sprite_t *sprite, float x0, float y0, const rdpq_blitparms_t *parms);
+
+/**
+ * @brief Perform a YUV blit of an NV12 sprite using a blitter
+ * 
+ * @see #yuv_blitter_run
+ */
+void rdpq_sprite_yuv_blitter_run(yuv_blitter_t *blitter, sprite_t *sprite);
+
```
rasky, line 145: `yuv_blitter_t` is a recorded block that allows to blit *arbitrary* YUV surfaces (of the given size/format) by using placeholders. The idea is to prepare a block once for eg drawing 320x240x2 NV12 frames, and then draw any frame during the playback (considering MPEG1/H264 will use different surface_t for different frames).

So I don't understand very well what these functions do. Seems like a bit of a semantic confusion: they accept a specific, single `sprite_t` as input, but then proceed to create a generic blitter for all `sprite_t` of the same size? That seems a bit confusing. I assume you are using the sprite as a template to say "create a blitter to draw sprites identical to this one". Still, I find it a bit confusing.

Also whatever problem you are solving here is not specific of YUV sprites at all. If you feel it's important to provide an API to do this, it would make sense for it to be available for all sprite formats. All sprites would benefit from this API.

If we want to go this way, I think we should introduce a rdpq_sprite_blitter_t, and have APIs such as:

```C
blitter = rdpq_sprite_blitter_new(format, width, height, x0, y0, parms);
rdpq_sprite_blitter_run(blitter, sprite);
```

and make it work for all formats.

I just noticed that this would also be missing the `yuv_format`... This makes me wonder whether we really want `yuv_format` to be a "subformat", or maybe upgrade them to the top-level `tex_format_t`...

include/sprite.h
rasky: I thought a bit more of this issue. I think there's too much freedom.

At yuv.h level, it makes sense for `yuv_format_t` to specify the various packing/layouts like NV12, etc.  yuv.h is an utility library and should support many different formats that can commonly be produced by different image/video pipelines. So it makes sense to just support all of the equally behind an unified API.

On the other hand, mksprite should be much more strict. Why should a mksprite user want to request a lossysprite which is exactly NV12 or NV16? The pixel layout is mandated by the *codec* being used. rsph264 currently emits a full planar format. Tomorrow, it could be optimized to directly emit a semi-planar format (that would be a good optimization to have one day). This should be left as an implementation detail to the codec. 

The only possible input that a user *might* want to specify at the mksprite level would be the chroma subsampling, as that might affect the quality of the image. But that is *not* the pixel layout. That input could be specified as a different pseudo texture format, for instance "YUV24" to mean YUV 4:4:4 and "YUV12" to mean YUV 4:2:0 (and maybe just also add aliases like YUV_444 and YUV_420). Currently we don't support this in rsph264 so it's even moot to discuss about the correct syntax here. Anyway this is just the chroma subsampling: the pixel layout should be chosen by the decoder, and even changing over time.

I think by doing this, we should simplify quite a bit of code in lossysprite.c. I think you're currently post-processing pixels to honor the pixel layout requested by the user in mksprite, but I think that code should go. In fact, in addition of being inefficient, I don't see any reason why an user should want that CPU-based preprocessing at runtime, when yuv.h can handle the native pixel layout anyway.

PR #879, opened 2026-05-03
[Preview] Core magma + mgfx API

snacchus:
This PR adds the following:
- magma: A low level API for RSP-accelerated 3D graphics. See magma.h for more details.
- mgfx: Stands for "Magma fixed function" and offers a basic general purpose vertex shader to be used with magma.
- mghello: A new example that showcases the minimal setup required to use magma.


src/magma/magma.c
```diff
@@ -0,0 +1,986 @@
...
+#include "magma_constants.h"
+#include "rspq.h"
+#include "rdpq.h"
+#include "utils.h"
+#include "rsp_magma.h"
+#include "../rdpq/rdpq_internal.h"
+#include "rsp_asm.h"
+#include <malloc.h>
+
+// TODO: Documentation on how magma works internally
```
Dragorn421: was about to point out how the difference between "loaders" and "patches" (in `mg_meta_attribute_t`) could be clearer (from what I gathered "loaders" refer to patching immediates pointing to vertex attributes, and "patches" refer to changing whole instruction words in case an attribute isn't used at all)
but then I saw this TODO and yeah

Dragorn421 (commented):
(the glhello example needs adding to the examples/Makefile)

include/mgfx.h
```diff
@@ -0,0 +1,162 @@
...
+    float end;      ///< Distance from the eye position where fog ends. Geometry farther away than this distance will be fully fog-colored.
+} mgfx_fog_parms_t;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// @cond
+extern rsp_ucode_t rsp_mgfx;
+extern rsp_ucode_t rsp_mgfx_env;
```
Dragorn421: you can move these externs to inside `mgfx_get_shader_ucode`, this is done plenty in libdragon to hide symbols. for example in `rdpq_clear`

include/mgfx_constants.h
```diff
@@ -0,0 +1,52 @@
...
+#define __MGFX_CONSTANTS
+
+/** @brief Input number of the "Position" vertex input. */
+#define MGFX_ATTRIBUTE_POSITION         0
+
+/** @brief Input number of the "Normal" vertex input. */
+#define MGFX_ATTRIBUTE_NORMAL           1
+
+/** @brief Input number of the "Color" vertex input. */
+#define MGFX_ATTRIBUTE_COLOR            2
```
Dragorn421: should this mention taking "RGBA"

In general, should these mention the type they expect? like, `uint8_t[3]`, `int16_t[3]`, ...

include/mgfx.h
```diff
@@ -0,0 +1,168 @@
...
+
+/** @brief Data structure of a single light in the lighting uniform. */
+typedef struct
+{
+    int16_t position[4];            ///< The light's position.
+    int16_t color[3];               ///< The light's color.
+    int16_t intensity;              ///< The light's intensity.
+} __attribute__((packed, aligned(16))) mgfx_light_t;
+
+/** @brief Data structure of the lighting uniform. */
```
Dragorn421, line 59: I would add static asserts here for `MGFX_LIGHT_POSITION` and co, like
`_Static_assert(offsetof(mgfx_light_t, position) == MGFX_LIGHT_POSITION);`

include/mgfx.h
```diff
@@ -0,0 +1,168 @@
...
+} mgfx_lighting_parms_t;
+
+/** 
+ * @brief Fog parameters. 
+ * 
+ * Fog is disabled when start == end.
+ * When fog is enabled, vertex alpha is always overwritten by the computed fog factor.
+ * Note that fog must also be enabled in rdpq to work properly (see #rdpq_mode_fog).
+ */
+typedef struct
```
Dragorn421, line 122: curious, what does start > end do? does it do the expected thing of fully fogged up to end, then decreasing fog, then no fog from start and beyond?

snacchus: I just tried and it does seem to work as expected.

include/mgfx.h
snacchus: They're there already in mgfx.c. Or did you mean they should be moved here?

Dragorn421: I did not notice them in mgfx.c, my bad
I think it'd make more sense near the structs since you would see them when possibly modifying the struct, but it's not a big deal

include/mgfx_macros.h
```diff
@@ -0,0 +1,29 @@
...
+#define MGFX_S8_8(v)        ((v)*(1<<MGFX_VTX_TEX_SHIFT))
+
+/** @brief Construct a vertex position in a format supported by the mgfx shader. */
+#define MGFX_POS(x, y, z)   { MGFX_S10_5(x), MGFX_S10_5(y), MGFX_S10_5(z) }
+
+/** @brief Construct a texture coordinate in a format supported by the mgfx shader. */
+#define MGFX_TEX(s, t)      { MGFX_S8_8(s), MGFX_S8_8(t) }
+
+/** @brief Construct a vertex normal in a format supported by the mgfx shader. */
+#define MGFX_NRM(x, y, z)   ((((x) & 0x1F)<<11) | \
```
Dragorn421, line 23: this macro stands out as not taking floats as input compared to `MGFX_POS` and `MGFX_TEX`, would it be better by taking -1 - 1 input like
```suggestion
#define MGFX_NRM(x, y, z)   ((((int)(x * 0xF) & 0x1F)<<11) | \
```
but then there's the usual caveat that you can't represent the full range of values (eg 1 gets encoded as 0xF (15/16 instead of 1)) which gets hidden instead of possibly being dealt with explicitly outside the macro...

examples/mghello/mghello.c
```diff
@@ -0,0 +1,134 @@
...
+        });
+
+        // Set the culling mode.
+        mg_set_culling(&(mg_culling_parms_t) {
+            .cull_mode = MG_CULL_MODE_NONE
+        });
+
+        // Configure the type of triangles that should be emitted to the RDP.
+        // We just want vertex colors, so use SHADE.
+        mg_set_geometry_flags(0);
```
Dragorn421:
```suggestion
        mg_set_geometry_flags(0);
```

snacchus: Well spotted! Fixed.

include/mgfx_constants.h
snacchus: I've added more detailed docs to all attributes.

include/mgfx.h
snacchus: Moved them.

src/magma/magma.c
snacchus: I'll get around to this soon hopefully

include/mgfx_macros.h
snacchus: Yeah to properly convert from floats you have to do something like this:
```c
    int16_t x = CLAMP(roundf(src[0] * 15.5f), -16.0f, 15.0f);
    int16_t y = CLAMP(roundf(src[1] * 31.5f), -32.0f, 31.0f);
    int16_t z = CLAMP(roundf(src[2] * 15.5f), -16.0f, 15.0f);
    int16_t packed = MGFX_NRM(x, y, z);
```

I guess that could be baked into the macro as well? What do you think?

Dragorn421: `roundf` can't be baked into the macro as it's a function and the macro may be used in data outside functions, I guess we could come up with a macro version of it but it's starting to sounds complicated

also the macro should remain a simple value that can be assigned to a struct member in data, so `({/* code */; return_value;})` is out of question

but at the same time ideally you want to evaluate the argument to a macro once, but it's going to be evaluated many times with a CLAMP wrapping a ROUNDF...

all in all I definitely would like to see a `MGFX_NRM(0.42f, 0.42f, 0.0f)` macro so maybe we do go the multi-eval route and mention in the docs that arguments will be evaluated several times (so don't pass in function calls, which sounds unlikely anyway)?

src/magma/mgfx.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+void mgfx_get_matrices(mgfx_matrices_t *dst, const mgfx_matrices_parms_t *parms)
+{
+    mgfx_convert_matrix(&dst->mvp, parms->model_view_projection);
+    mgfx_convert_matrix(&dst->mv, parms->model_view);
+    
+    for (uint32_t i = 0; i < 16; i++)
+    {
+        dst->normal[i] = parms->normal[i] * (1<<15);
```
Dragorn421: take as normal matrix the identity, then this overflows where normal[i]==1
That sounds like a problem

snacchus: The values are now being clamped to avoid overflows

include/mgfx_macros.h
snacchus: As discussed on N64brew: I'll leave this as is for now since there doesn't seem to be a catch-all solution. We can revisit this later.

src/magma/magma.c
snacchus: Docs are done. Let me know whether they're actually helpful or if there's something missing.

src/magma/magma.c
```diff
@@ -1,7 +1,156 @@
...
+ * Drawing geometry roughly happens in two steps:
+ * 1. The vertex pipeline produces vertex data and stores it in the vertex cache. This step is highly customisable and by far the most complex.
+ * 2. Drawing commands then read the stored vertex data, assemble RDP triangle commands and submit them into the RDP queue. This step is mostly fixed function.
+ * 
+ * ## Vertex pipeline
+ * The vertex pipeline (or just "pipeline" for short) is invoked by enqueueing the command `MG_CMD_LOAD_VERTICES`.
+ * However, the core magma ucode doesn't actually contain the pipeline at all.
+ * Instead, that command jumps to a small bootstrapping segment which will load the actual pipeline into IMEM first.
+ * More precisely, it will load the shader code (see below) of the currently bound pipeline.
+ * It is loaded at the same address as the bootstrapper, ovewriting it.
```
Dragorn421:
```suggestion
 * It is loaded at the same address as the bootstrapper, overwriting it.
```
GeWellActually

examples/mghello/mghello.c
```diff
@@ -0,0 +1,135 @@
...
+        .vertex_layout.attribute_count = sizeof(vertex_attributes)/sizeof(vertex_attributes[0]),
+        .vertex_layout.attributes = vertex_attributes,
+        .vertex_layout.stride = sizeof(vertex)
+    });
+
+    // Shader uniforms are not initialized to defaults automatically, so we need to create
+    // a uniform buffer that sets everything to sane values. 
+    uniforms *uniform_data = malloc_uncached(sizeof(uniforms));
+
+    mgfx_get_fog(&uniform_data->fog, &(mgfx_fog_parms_t) {});
```
Dragorn421, line 43: imo worth adding a comment here that this effectively disables fog by passing start=end=0

src/magma/magma.c
snacchus: fixed!

examples/mghello/mghello.c
snacchus: done

PR #883, opened 2026-05-11
Workaround/fix -Wextra warnings

Dragorn421:
Most changes are required for me to build tiny3d, which uses `-Wextra -Wdouble-promotion`
https://discord.com/channels/205520502922543113/1262534624689651832/1503307355197280337

Note I'm using an older toolchain. I assume recent toolchains don't exhibit this problem

The changes in .c files were not required, but while I was at it...

I will point out the debatable changes in a self-review. But something should be picked between
- merging this PR to libdragon
- removing `-Wextra -Wdouble-promotion` from tiny3d and fixing `-Wstrict-aliasing` in `color_to_packed32`
- not supporting older toolchains

### DIFF
--- a/include/graphics.h
+++ b/include/graphics.h
@@ -142,7 +142,10 @@ inline uint16_t color_to_packed16(color_t c) {
 
 /** @brief Convert a #color_t to the 32-bit packed format used by a #FMT_RGBA32 surface (RGBA 8888) */
 inline uint32_t color_to_packed32(color_t c) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
     return *(uint32_t*)&c;
+#pragma GCC diagnostic pop
 }
 /** @brief Create a #color_t from the 16-bit packed format used by a #FMT_RGBA16 surface (RGBA 5551) */
 inline color_t color_from_packed16(uint16_t c) {
--- a/include/graphics.h
+++ b/include/graphics.h
@@ -142,10 +142,7 @@ inline uint16_t color_to_packed16(color_t c) {
 
 /** @brief Convert a #color_t to the 32-bit packed format used by a #FMT_RGBA32 surface (RGBA 8888) */
 inline uint32_t color_to_packed32(color_t c) {
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-    return *(uint32_t*)&c;
-#pragma GCC diagnostic pop
+    return (c.r << 24) | (c.g << 16) | (c.b << 8) | (c.a << 0);
 }
 /** @brief Create a #color_t from the 16-bit packed format used by a #FMT_RGBA16 surface (RGBA 5551) */
 inline color_t color_from_packed16(uint16_t c) {
--- a/include/rdpq.h
+++ b/include/rdpq.h
@@ -680,7 +680,6 @@ inline void rdpq_load_tlut_raw(rdpq_tile_t tile, int color_idx, int num_colors)
  */
 inline void rdpq_set_tile_size_fx(rdpq_tile_t tile, uint16_t s0, uint16_t t0, uint16_t s1, uint16_t t1)
 {
-    assertf((s0) >= 0 && (t0) >= 0 && (s1) >= 0 && (t1) >= 0, "texture coordinates must be positive");
     assertf((s0) <= 1024*4 && (t0) <= 1024*4 && (s1) <= 1024*4 && (t1) <= 1024*4, "texture coordinates must be smaller than 1024");
 
     extern void __rdpq_write8_syncchange(uint32_t, uint32_t, uint32_t, uint32_t);
@@ -1013,7 +1012,7 @@ inline void rdpq_set_prim_color(color_t color)
 inline void rdpq_set_detail_factor(float value)
 {
     // NOTE: this does not require a pipe sync
-    int8_t conv = (1.0 - value) * 31;
+    int8_t conv = (1.0f - value) * 31;
     extern void __rdpq_fixup_write8_syncchange(uint32_t, uint32_t, uint32_t, uint32_t);
     __rdpq_fixup_write8_syncchange(RDPQ_CMD_SET_PRIM_COLOR_COMPONENT, ((conv & 0x1F) << 8) | (2<<16), 0, 0);
 }
--- a/include/rdpq.h
+++ b/include/rdpq.h
@@ -680,6 +680,7 @@ inline void rdpq_load_tlut_raw(rdpq_tile_t tile, int color_idx, int num_colors)
  */
 inline void rdpq_set_tile_size_fx(rdpq_tile_t tile, uint16_t s0, uint16_t t0, uint16_t s1, uint16_t t1)
 {
+    assertf((s0) >= 0 && (t0) >= 0 && (s1) >= 0 && (t1) >= 0, "texture coordinates must be positive");
     assertf((s0) <= 1024*4 && (t0) <= 1024*4 && (s1) <= 1024*4 && (t1) <= 1024*4, "texture coordinates must be smaller than 1024");
 
     extern void __rdpq_write8_syncchange(uint32_t, uint32_t, uint32_t, uint32_t);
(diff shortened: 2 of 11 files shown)


include/graphics.h
```diff
@@ -142,7 +142,10 @@ inline uint16_t color_to_packed16(color_t c) {
 
 /** @brief Convert a #color_t to the 32-bit packed format used by a #FMT_RGBA32 surface (RGBA 8888) */
 inline uint32_t color_to_packed32(color_t c) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
     return *(uint32_t*)&c;
+#pragma GCC diagnostic pop
 }
```
Dragorn421, line 146: idk. I tried a memcpy but that triples the instructions count.

include/rdpq.h
```diff
@@ -680,7 +680,6 @@ inline void rdpq_load_tlut_raw(rdpq_tile_t tile, int color_idx, int num_colors)
  */
 inline void rdpq_set_tile_size_fx(rdpq_tile_t tile, uint16_t s0, uint16_t t0, uint16_t s1, uint16_t t1)
 {
-    assertf((s0) >= 0 && (t0) >= 0 && (s1) >= 0 && (t1) >= 0, "texture coordinates must be positive");
```
Dragorn421, line 683: the arguments *are* unsigned so this assert is redundant but still...

src/utils.h
```diff
@@ -28,7 +28,10 @@
 /** @brief Round n down to the previous multiple of d */
 #define ROUND_DOWN(n, d) ({ \
     typeof(n) _n = n; typeof(d) _d = d; \
+    _Pragma("GCC diagnostic push") \
+    _Pragma("GCC diagnostic ignored \"-Wtype-limits\"") \
     ((_n >= 0) ? ((_n) / (_d) * (_d)) : -(((-_n + (_d) - 1) / (_d)) * (_d))); \
+    _Pragma("GCC diagnostic pop") \
 })
```
Dragorn421, line 32: When n is of unsigned type then `-Wtype-limits` is unhappy about the `_n >= 0`
again idk

thekovic:
Can you document what version your toolchain is (specifically, GCC version in this case)?

Dragorn421:
> Can you document what version your toolchain is (specifically, GCC version in this case)?

```
$ $N64_GCCPREFIX/bin/mips64-elf-gcc --version
mips64-elf-gcc (GCC) 13.2.0

$ ls -la $N64_GCCPREFIX/bin/mips64-elf-gcc
-rwxr-xr-x 2 root root 1690728 sept.  3  2023 /opt/n64/libragon_gcc/bin/mips64-elf-gcc
```

include/graphics.h
Dragorn421: https://godbolt.org/z/eKMcd1jcz

Dragorn421: the following compiles to the same instructions as the current implementation but I have no idea if it's any better C-standard-wise
```c
uint32_t color_to_packed32_union(color_t c) {
    union {
        color_t c;
        uint32_t i;
    } v = { .c = c };
    return v.i;
}
```

rasky: It's OK for me to fix this issue, but I would prefer finding C code that is strict-alias safe and does not emit diagnostics.

I don't know why memcpy would increase instruction count, it doesn't elsewhere. Maybe if you can post some example code to use as reference, we can try some alternatives. For instance, I would try `__builtin_memcpy` and also just the standard shift+or sequence that might be optimized away.

include/rdpq.h
rasky: I would prefer to keep this.

src/utils.h
rasky: I would prefer not to add this source-level overhead for valid code

rasky (commented):
Double promotion things are OK and even useful. I'm a bit doubtful about the rest

include/graphics.h
Dragorn421: `__builtin_memcpy` is the same as `memcpy`
"the standard shift+or sequence" is indeed optimized away, I'll switch to that

[godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYM9DgDJ4GmADl3ACNMYhBJAGZSAAdUBUJbBmc3Dz04hJsBX38gllDwqItMKyyGIQImYgIU908uYtKkiqqCHMCQsIjohUrq2rSG3tb2vILugEoLVFdiZHYOAFIAJki/ZDcsAGpFyMde/EEAOgRd7EWNAEEVtYYN123d/YJiP2ATs4vrq4IATxjMFgqFtesRXNYtqYmAQXnhgq4CJhTFAYkxkABrQETCZfRYAdgAQl8tiStgB6ABUFK2AAFgq9MMCAEqArZoFhxfyCLYUsnE0muPwEAAcxgIW2IuyJV1J5KptPpeEZWwA4sRMIw2agOQJGOLefySYLBKLxcApYa5dS6QzgQS3JgtTqufq%2BTKBULTVtghb3STKdbFcrLrQYggmE7OXqeW7LrLjSKxVsmL7rniACJa2gkMWpr4JyLLJNobPEMWoYyojGAwvGPx4CAlnPi5ATHaEy3qgizBg8iAFosECk45YANmQeYzuKuA%2BLLmbFarmPQtbYLGQMV%2BjfnZZbbfx0rjHsEtfFeFTsrXG63K1HeFIOzH5i2khxkUPsq7Pa25/fuKnVz5kKp5ZgulZosutauAwSTbqWxb7h2fpbNBSTth%2Bsqkk2u5shemFGsBg4/nhJL4pmABuOyRJmB5bEcyBUZmDFkSREqYN2xC9uRRy/oeLGATOhFzvBRDgdWK5FvCdA2AwxhXpucHNmyiEYQRJ5EbxlqmFJtAyXJmDrgpt73o%2B44Pq%2BrFfpxxF/oBAHfJcs4tju5ZiZBkmEKgMQKIpOGtuhnbsd%2BjZHMQVGOE8WzLK%2B7aOFsIXAOFkVcKOiFxSFwRJXsWzCml8XIEcEZPJFGhvnx9kcFMtCcAArLwngcFopCoJwcUKDMcyOjcPCkAQmiVVM6IgDVGj6Jwki8CwEgaKNDVNS1HC8AoICjX1jWVaQcCwEg7IxHQYTkJQu37eEyAsHg3mjpIwApVwo00LpYTLRAwT9aQwR%2BFUvycD1H3MMQvwAPLBNomDWD9vDsmwgiAwwtDfetpBYPCwCOGItDLdwvBYCwhjAOIiP4Oq1h4ORmCY01mCqGDCILD1QolG9tBwsQX3OFgb2wlNWOkGTxDBPEmDpgZePM0Y/VTFQBjAAoABqSoAO6AwCDU9fwggiGI7BSDIgiKCo6iI7oDQGOLpjGOYzPBMtkBTF5ZSYwAtIDyxbI7uNzKc1HkcQUQzW7LDdv4uzpj7fsaAHTAfSHqBXbwqB868WA2xAUyWGDZT2AwTguHUXjZ6MnThA0GSJAIAz1LE8Rlwwhf5F0QwlBnzR9DUueDI0zcCC01R1%2BMQytxXejDL3fgdPXxdpx18wSFVtX1W9C1bOdl2SFswDIAxKVHFwRwRxAuCeWF3UTLwa1aNipAIJgTBYOEqekENI1jRwE2kHN8ecEtK29RLc8cMsBeiMFpnz/rzJ6SQIhAA%3D%3D%3D)

Dragorn421:
I agree with Rasky's last two comments, but let me reiterate that

> something should be picked between
> 
> * merging this PR to libdragon
> * removing `-Wextra -Wdouble-promotion` from tiny3d and fixing `-Wstrict-aliasing` in `color_to_packed32`
> * not supporting older toolchains

thekovic:
Since this is a PR to libdragon, I think we should focus on issues that only come up when compiling libdragon by itself. Any issues that arise due to building from Tiny3D's Makefiles should be resolved by Tiny3D. This can be either modifying their Makefile to placate older toolchains, or declaring old toolchains as unsupported. I think the 2nd choice might be @HailToDodongo 's preferred considering T3D already leans towards using latest C++ (23 and above) features in its code.

PR #890, opened 2026-05-26
Simplified Lossysprite support + test ROM

meeq:
A simplified take on #874 that strips out all of the advanced functionality.

A lossy sprite will be decompressed into an RGBA16 sprite with hard-coded chroma subsampling and colorspace. Once decoded, it works just like a regular sprite. See the new `lossysprite` example for sample usage:

image

Lossy sprite support requires an opt-in with `lossysprite_init` so that software which doesn't need lossy compression can strip out this functionality. Attempting to load a lossy sprite without registering the decoder will trigger an assertion.

image


src/rdpq/rdpq_tex.c
```diff
@@ -363,7 +363,22 @@ int tex_loader_calc_max_height(tex_loader_t *tload, int s0, int s1)
...
+    // counts only one half. A single LOAD_TILE/LOAD_BLOCK still transfers
+    // (s1-s0)*2 RDRAM bytes per line; cap the strip so the total transfer
+    // stays within the 2KB-per-load limit that LOAD_BLOCK enforces and
+    // some emulators sanity-check on LOAD_TILE.
+    if (fmt == FMT_YUV16) {
+        int rdram_pitch = TEX_FORMAT_PIX2BYTES(fmt, s1 - s0);
+        int max_h_rdram = 2048 / rdram_pitch;
+        if (max_h_rdram < max_h) max_h = max_h_rdram;
+        if (max_h < 1) max_h = 1;
+    }
```
rasky: RGBA32 has the same split-TMEM layout, it is also tested by the testrom and it doesn't require any tweak in this function. I'm confused on why this would be required here. What happens if you don't add this code? Can you maybe compare RGBA32 with YUV16 and check what happens in the example?

meeq: I’ll pull this out of this PR: this was a leftover from the previous branch that I intended to submit as a separate PR if it was actually necessary.

include/lossysprite.h
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * The leading four zero bytes are a safety pad: if an LSPR file reaches
+ * #sprite_load without #lossysprite_init having registered the decoder,
+ * the buffer is reinterpreted as a #sprite_t — the pad forces width=height=0
+ * so the situation is detectable and a clear assertion can fire instead of
+ * silently producing a corrupt sprite.
+ */
+#define LSPR_FILE_MAGIC      "\0\0\0\0LSPR"
+/** @brief Size in bytes of #LSPR_FILE_MAGIC. */
+#define LSPR_FILE_MAGIC_SIZE 8
```
rasky: I would move these macros to lossysprite.c, they don't seem to belong to the public API.

include/lossysprite.h
```diff
@@ -0,0 +1,144 @@
...
+ * with #sprite_free.
+ *
+ * Prefer #sprite_load when the same code path needs to handle both
+ * regular and lossy sprites: it sniffs the LSPR magic and delegates here
+ * automatically once #lossysprite_init has registered the decoder.
+ *
+ * @param fn    Path to the LSPR file
+ * @return      The loaded sprite (free with #sprite_free)
+ */
+sprite_t* lossysprite_load(const char *fn);
```
rasky: With the current state of the PR (with fewer features) I don't think there is a use case where calling lossysprite_load or lossysprite_load_buf provide a benefit over the sprite equivalent. These APIs were designed to maybe provide lower-level knobs like eg. do asynchronous load, or YUV stuff. In their current design, they are identical in behavior to the sprite counterparts.

Maybe we can remove them from this header file and readd them later if we add maybe some extra parms to them?

include/lossysprite.h
```diff
@@ -0,0 +1,144 @@
...
+ * The @p out buffer must be at least #lossysprite_decoded_size_buf bytes
+ * long and aligned to #LOSSYSPRITE_DECODE_ALIGN bytes.
+ *
+ * @param buf       Pointer to the encoded LSPR file contents
+ * @param sz        Size of @p buf in bytes
+ * @param out       Pointer to the output buffer (64-byte aligned)
+ * @param out_sz    Size of @p out in bytes
+ * @return          The decoded sprite (lives in @p out)
+ */
+sprite_t *lossysprite_load_into(const void *buf, int sz, void *out, size_t out_sz);
```
rasky: Is lossysprite_load_into really necessary? I'm always a bit doubtful about these APIs as in general, as if you have an already allocated buffer which is 320*240, the difference between reusing it and freeing it and then letting lossysprite_load_buf reallocate it, should be almost none (especially for a CPU-intenstive function like lossy sprite loading).

src/lossysprite.c
```diff
@@ -0,0 +1,362 @@
...
+
+#include "lossysprite.h"
+
+///@cond
+// Match the layout used inside the h264_decoder TU (h264_decoder.c defines
+// this before including all internal .c files). Without it, mbStorage_t here
+// gains an extra u32 `decoded` field, shifting mbA/mbB/mbC/mbD off the offset
+// h264bsdInitMbNeighbours writes to — neighbour pointers come back NULL.
+#define OPTIMIZE_NO_DECODED_FLAG
+///@endcond
```
rasky: Good point but maybe we could just move this to video/h264_decoder.h?

src/lossysprite.c
```diff
@@ -0,0 +1,362 @@
...
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <malloc.h>
+
+/** @brief LSPR version number. */
+#define LSPR_VERSION 4
+
+///@cond
+#define ALIGN64(n) (((n) + 63) & ~63)
```
rasky: FWIW we have the ROUND_UP macro for this kind of alignments

src/lossysprite.c
```diff
@@ -0,0 +1,362 @@
...
+
+    // Rotating ring of mbLayer instances. Each MB writes its CAVLC packed
+    // delta records into mbLayer.residual.posCoefBuf, which the RSP later
+    // DMAs asynchronously. Using a single mbLayer would race MB N+1's CAVLC
+    // writes against MB N's pending DMA; rotating gives the RSP enough lag
+    // to consume each MB's buffer before its slot is reused. This mirrors
+    // the video-player path (h264bsdDecodeSliceData uses the same idiom
+    // with NUM_PARALLEL_MACROBLOCKS).
+    ///@cond
+    #define LSPR_MB_RING_SIZE 32
```
rasky: I would attempt lowering this to just 2, and then using syncpoints to pace CPU/RSP. I think this could almost match the performance while greatly reduce memory occupation (also you can allocate 2 macroblockLayer_t directly on the stack).

src/lossysprite.c
```diff
@@ -0,0 +1,362 @@
...
+                (unsigned long)currMbAddr);
+        currMbAddr++;
+        if (!h264bsdMoreRbspData(&strm))
+            break;
+    }
+
+    // Drain the RSP queue once at slice end. The YUV→target conversion
+    // below reads `pic`, which is the destination of all queued residual /
+    // intra-pred tasks; we must wait for them before the CPU reads it.
+    rsph264_sync();
```
rasky, line 209: I think the comment is a leftover since there's no CPU code accessing the YUV buffer in sight. The sync is still required though because you're freeing mbLayers/mb immediately after. Maybe update the comment?

src/lossysprite.c
```diff
@@ -0,0 +1,362 @@
...
+            // combiner state leaks back to the caller's render mode, which makes
+            // the next non-YUV draw on the screen surface emit magenta/purple
+            // pixels until the caller next sets a render mode.
+            rdpq_mode_push();
+            {
+                yuv_tex_blit(&frame, 0, 0, NULL, &YUV_BT709_FULL);
+            }
+            rdpq_mode_pop();
+        }
+        rdpq_detach_wait();
```
rasky, line 276: I would like to eventually grow an async loading, where sprite_load() returns while RSP/RDP are still crunching. This could be a feature for the lossysprite loading API, to use in alternative to sprite_load for the cases where you know what you're doing.

The implementation should be to remove the sync above, use deferred frees for temporary buffers, and use rdpq_detach here. I think this could help saving some loading time as you could parallelize some portion of it.

rasky (commented):
First pass!

PR #899, opened 2026-06-07
examples: Make sure all examples' Makefiles clean up their DFS files

nadiaholmquist:
Currently some of the examples do not delete converted files in their `filesystem` directories when you run `make clean`. This can cause problems because of stale assets built with old tool versions in your local libdragon copy if you've built the examples a long time ago. Here's a small PR to fix that.


rasky:
Thanks!

rasky:
Thanks!

PR #906, opened 2026-06-13
GL: re-implement the RSP pipeline on top of magma

snacchus:
This PR completely replaces the RSP pipeline of the OpenGL implementation with an improved version that is based on magma. This results in overall greatly improved performance (up to ~4,6x less RSP time required has been measured).

## Overview
The rspq overlay `rsp_gl_pipeline` has been removed and replaced with a magma vertex shader of the same name. Like the old overlay, the shader expects vertices in a fixed format, but flexible layout. Since magma works most efficiently when loading large batches of vertices, the new implementation therefore has to convert all vertex data into internal buffers before handing them off to magma. The data is converted on demand when draw calls are dispatched.

In some cases, whenever GL is able to track changes to vertex layout and vertex data, converted data may be retained and reused across frames. This is the case when using vertex buffer objects and vertex array objects in conjunction. This is consequently the most optimal way to render geometry now, as opposed to display lists with the old implementation.

Because best practices for optimal performance are changing with this re-implementation, drawing models with model64 will not be optimal for now. The performance benefits from the new implementation still outweigh this pessimization, however (see measurements below). An improved version of model64 will follow in a later PR.

## Performance comparison
The following measurements were made on a real PAL N64.

This is the baseline, using the old RSP pipeline:
Profiler results: baseline

Compare with the new, magma based implementation without further modifications to model64:
Profiler results: new implementation with unoptimized mesh

And finally, the new implementation using an optimized version of model64:
Profiler results: new implementation with optimized mesh

## Breaking changes
- The currently bound `GL_ELEMENT_ARRAY_BUFFER_ARB` is no longer a global state and instead now scoped to the current vertex array object. This now correctly implements the GL specification.
- `glDrawElements` only supports `GL_UNSIGNED_SHORT` as index type for now. This will be fixed in the future.
- Texture coordinates with 3 or more components don't work correctly in the new implementation. They are effectively treated as 2 components. Depending on demand, this will be fixed in the future.
- Tex gen with `GL_OBJECT_PLANE` and `GL_EYE_PLANE` is no longer supported by the RSP pipeline. The CPU pipeline will be used instead.

## Other changes
- The new packed format `GL_SHORT_5_6_5_N64` has been added and is available for use in `glNormalPointer`.
- The custom hashtable implementation in `obj_map.h` has been removed and usages migrated to `hashtable_internal.h`.


Dragorn421 (commented):
(this visually passes my triangle tests at https://github.com/Dragorn421/n64homebrew/tree/main/gltest (on ares))

src/GL/rsp_pipeline.c
```diff
@@ -157,426 +188,506 @@ const rsp_read_attrib_func rsp_read_funcs[ATTRIB_COUNT][ATTRIB_TYPE_COUNT] = {
...
+        .count = count
+    };
+    
+    prepare_drawing(state->array_object, range);
+
+    mg_draw_begin();
+    mg_input_assembly_parms_t input_assembly_parms = array_object_get_input_assembly_parms(state->array_object, mode, range);
+    mg_ex_draw(&input_assembly_parms, count, first, mode);
+    mg_draw_end();
 }
```
rasky, line 617: If you create the array object already offseted by `first`, why you need to pass `first` to `mg_ex_draw`? Is that a bug? Did you test `glDrawArray` with `first != 0`?

src/GL/list.c
```diff
@@ -94,7 +94,7 @@ void glCallList(GLuint n)
     // During display list recording, we cannot anticipate whether it will be called within a glBegin/glEnd pair or not.
     assertf(!state->begin_end_active, "glCallList between glBegin/glEnd is not supported!");
 
-    rspq_block_t *block = obj_map_get(&state->list_objects, n);
+    rspq_block_t *block = hashtable_lookup(&state->lists, n);
```
rasky, line 97: Switching from obj_map to hashtable seems to have removed the NULL check. I think both 0 and TOMBSTONE_KEY are reserved values that should return `GL_INVALID_VALUE` instead of crashing

src/GL/rsp_pipeline.c
```diff
@@ -157,426 +188,506 @@ const rsp_read_attrib_func rsp_read_funcs[ATTRIB_COUNT][ATTRIB_TYPE_COUNT] = {
...
+            offsetof(native_vertex_t, texcoord),
+            offsetof(native_vertex_t, mtx_index),
+        },
+        .stride = sizeof(native_vertex_t)
+    };
+
+    for (array_type_t i = 0; i < ARRAY_COUNT; i++)
+    {
+        parms->arrays[i] = &state->array_object->arrays[i];
+    }
```
rasky, line 521: Shouldn't this be filtered by `is_array_used`? We have a similar filter in data_source.c. It would seem that arrays could easy have NULL pointers otherwise and lead to crashes?

src/GL/rsp_gl_pipeline.rspl
```diff
@@ -0,0 +1,573 @@
...
+                load_mat4x4(vmvp0, vmvp1, vmvp2, vmvp3, addr);
+                #endif
+            }
+            
+            vlpos *= vnorm:sfract;
+            VTEMP = vlambient * vambient;
+            VTEMP = vlcol:sfract +* vlpos:sfract.xxxxXXXX;
+            VTEMP = vlcol:sfract +* vlpos:sfract.yyyyYYYY;
+            vlcol = vlcol:sfract +* vlpos:sfract.zzzzZZZZ;
+            vlcol = max(vlcol, VZERO);
```
rasky, line 431: I think the N*L dot product should be clamped (like we do `gl_clamped_dot` in the CPU pipeline. Also I'm not sure about how the attenuation is handled. Should it be only applied to `vlcol` (which is the diffuse term if I understand correctly?)

src/GL/rsp_pipeline.c
```diff
@@ -157,426 +188,506 @@ const rsp_read_attrib_func rsp_read_funcs[ATTRIB_COUNT][ATTRIB_TYPE_COUNT] = {
...
+
+    mg_draw_begin();
+    mg_input_assembly_parms_t input_assembly_parms = array_object_get_input_assembly_parms(state->array_object, mode, range);
+    mg_ex_draw_indexed(&input_assembly_parms, indices, count, -range.first, mode);
+    mg_draw_end();
+}
+
+static void gl_rsp_draw_elements(GLenum mode, uint32_t count, const void* indices, GLenum type)
+{
+    assertf(type == GL_UNSIGNED_SHORT, "Index type must be GL_UNSIGNED_SHORT");
```
rasky, line 671: We were handling different types before. I guess this is not a huge issue, but maybe it should be documented as a change.

src/GL/cpu_pipeline.c
```diff
@@ -97,7 +102,15 @@ static void read_u32_i(GLubyte *dst, const uint32u_t *src, uint32_t count)
...
 }
 
-const cpu_read_attrib_func cpu_read_funcs[ATTRIB_COUNT][ATTRIB_TYPE_COUNT] = {
+static void read_packed565n(GLfloat *dst, const uint16u_t *src, uint32_t count)
+{
+    uint16_t v = *src;
+    dst[0] = ((v>>11)&0x1F) / 15.5f;
+    dst[1] = ((v>>5)&0x3F) / 31.5f;
+    dst[2] = ((v>>0)&0x1F) / 15.5f;
+}
```
rasky, line 111: I don't recall: did we have GL_SHORT_5_6_5_N64 before? And was it being used only for colors? I assume this can't be used for normals or other things that need signed numbers?

src/GL/rsp_gl2.rspl
```diff
@@ -0,0 +1,411 @@
...
+                m1:ufract = load(MATRICES_MV, 0x30);
+
+                MakeNormalMatrix(m0, m1);
+
+                // TODO: reuse modelview matrix already loaded into registers
+                GL_MtxMult();
+                dma_out_async(dmem_ptr, rdram_ptr, MATRICES_SIZE);
+                rdram_ptr += MATRICES_SIZE;
+                i += 1;
+            }
```
rasky, line 229: Isn't this one quite expensive to run on each draw call? we did have GL_STATE_PALETTE_DIRTY at least before but it seems not being used anymore.

src/GL/draw_call_cache.c
```diff
@@ -0,0 +1,121 @@
+#include "draw_call_cache.h"
+#include "gl_internal.h"
+#include "array_object.h"
+#include "fnv1a.h"
+#include "mg_ex.h"
+
+#define MAX_CACHED_DRAW_CALL_COUNT  8
```
rasky, line 7: Unused?

src/GL/draw_call_cache.h
```diff
@@ -0,0 +1,47 @@
+#ifndef __DRAW_CALL_CACHE_H
+#define __DRAW_CALL_CACHE_H
+
+#include <stdint.h>
+#include "rspq.h"
+#include "GL/gl.h"
+#include "hashtable_internal.h"
+#include "indices.h"
+
+#define MAX_CACHED_DRAW_CALLS   8
```
rasky, line 10: Unused?

src/GL/rsp_pipeline.c
```diff
@@ -157,426 +188,506 @@ const rsp_read_attrib_func rsp_read_funcs[ATTRIB_COUNT][ATTRIB_TYPE_COUNT] = {
...
+    mg_draw_begin();
+    draw_call_run(draw_call);
+    mg_draw_end();
 }
 
-static void gl_rsp_draw_elements(uint32_t count, const void* indices, read_index_func read_index)
+static void draw_elements_from_pointer(GLenum mode, uint32_t count, const void* indices)
 {
-    gl_fill_all_attrib_defaults(state->array_object->arrays);
+    index_bounds_t range = find_index_bounds(indices, count);
```
rasky, line 659: A linear scan of the whole index buffer on each draw call seems quite expensive? Can it be avoided in some way? I guess VBO is one but maybe there's a way to avoid it even without?

src/GL/indices.c
```diff
@@ -0,0 +1,15 @@
...
+    uint16_t min = USHRT_MAX;
+    uint16_t max = 0;
+
+    for (size_t i = 0; i < count; i++)
+    {
+        if (indices[i] < min) min = indices[i];
+        if (indices[i] > max) max = indices[i];
+    }
+
+    return (index_bounds_t) { min, max - min + 1 };
```
rasky, line 14: Potential surprising result for count=0, not guarded with an assertion

src/GL/rsp_gl_pipeline.rspl
```diff
@@ -0,0 +1,573 @@
+include "rsp_magma.inc"
+#include "../../include/magma_constants.h"
+#include "gl_constants.h"
```
rasky, line 3: Maybe add a comment explaining how to regenerate the various .S files. I guess the variants are generated via ENABLE_ENV_MAP / ENABLE_ENV_MAP, but maybe having here a guide would help people to tinker with the RSPL source.

src/GL/gl.c
rasky, line 182: `mg_close()`?

src/GL/data_cache.c
```diff
@@ -0,0 +1,91 @@
...
+static void update_stride_from_data_source(data_cache_t *cache)
+{
+    cache->stride = data_source_get_stride(cache->source);
+}
+
+static void reallocate_buffer(data_cache_t *cache, uint32_t new_size)
+{
+    free_buffer(cache);
+    // TODO: use allocator abstraction?
+    cache->buffer = malloc_uncached(new_size);
```
rasky, line 54: If I understand correctly, this is done at each draw call that doesn't use a VBO (is that what data_source_is_fully_internal means?)?

It seems like pretty expensive and fragmentation prone, especially with the release being done as a deferred call.

I guess this means that non-VBO draw calls are really not to be used?

src/GL/pipelines.c
```diff
@@ -0,0 +1,82 @@
...
+    mg_pipeline_t **pipelines = calloc(PIPELINE_COUNT, sizeof(mg_pipeline_t*));
+
+    // This will iterate over all possible combinations of features
+    for (size_t i = 0; i < PIPELINE_COUNT; i++)
+    {
+        pipelines[i] = mg_pipeline_create(&(mg_pipeline_parms_t) {
+            .vertex_shader_ucode = get_pipeline_ucode(i),
+            .vertex_layout = layout->vertex_layout
+        });
+    }
```
rasky, line 37: Did you evaluate lazy pipeline creation? Or maybe it's not a big issue?

src/GL/pipelines.c
```diff
@@ -0,0 +1,82 @@
...
+    &rsp_gl_pipeline,
+    &rsp_gl_pipeline_env,
+    &rsp_gl_pipeline_nrm,
+    &rsp_gl_pipeline_env_nrm
+};
+
+static rsp_ucode_t *get_pipeline_ucode(uint32_t features)
+{
+    return pipeline_ucodes[features];
+}
```
rasky, line 24: Maybe an assert here would help even that `features` comes from some non-trivial codepath

src/GL/ringbuffer.c
```diff
@@ -0,0 +1,35 @@
...
+    buf->entry_count = entry_count;
+    buf->current_index = 0;
+    buf->buffer = malloc_uncached(entry_size * entry_count);
+    buf->syncpoints = calloc(entry_count, sizeof(rspq_syncpoint_t));
+}
+
+void ringbuffer_free(ringbuffer *buf)
+{
+    free(buf->syncpoints);
+    free_uncached(buf->buffer);
```
rasky, line 15: Do these need deferred calls? Maybe not because it's only run after rspq_wait()?

src/GL/vertex_layout.h
```diff
@@ -0,0 +1,48 @@
...
+
+inline void vertex_layout_append(vertex_layout_t *vl, uint32_t input, uint32_t size)
+{
+    vertex_layout_add(vl, input, vl->vertex_layout.stride, size);
+}
+
+inline void vertex_layout_set_stride(vertex_layout_t *vl, uint32_t stride)
+{
+    vl->vertex_layout.stride = stride;
+}
```
rasky, line 35: I'm a bit confused by vertex_layout_append reading the stride, but then also writing it (within vertex_layout_add), plus there is a vertex_layout_set_stride? The API seems a bit confused to me... maybe it just needs some comments

src/GL/rsp_pipeline.c
```diff
@@ -7,102 +7,129 @@
...
+DEFINE_READ_FUNC(vtx_read_f64,  int16_t, doubleu,   POS_CONVERT, 3, vtx_default)
+DEFINE_FIXED_READ_FUNC(vtx_read_x16, int16_t, state->vertex_halfx_precision, 3, vtx_default)
+
+DEFINE_NORMAL_INT_READ_FUNC(nrm_read_i8,  int8_t,    3)
+DEFINE_NORMAL_INT_READ_FUNC(nrm_read_i16, int16u_t,  11)
+DEFINE_NORMAL_INT_READ_FUNC(nrm_read_i32, int32u_t,  27)
+DEFINE_NORMAL_FLT_READ_FUNC(nrm_read_f32, floatu)
+DEFINE_NORMAL_FLT_READ_FUNC(nrm_read_f64, doubleu)
+
+static void nrm_read_packed565(int16_t *dst, const int16_t *src, uint32_t count)
```
rasky, line 94: Shouldn't this use unaligned pointers?

src/GL/vertex_layout.c
```diff
@@ -0,0 +1,34 @@
...
+void vertex_layout_init(vertex_layout_t *vl)
+{
+    vl->vertex_layout.attributes = vl->attributes;
+    vl->vertex_layout.attribute_count = 0;
+    vl->vertex_layout.stride = 0;
+}
+
+void vertex_layout_add(vertex_layout_t *vl, uint32_t input, uint32_t offset, uint32_t size)
+{
+    mg_vertex_attribute_t *attr = &vl->attributes[vl->vertex_layout.attribute_count++];
```
rasky, line 13: Maybe this warrants an assert on the array size?

src/GL/mg_ex.c
```diff
@@ -0,0 +1,82 @@
...
+    const uint32_t prim_count = count / 4;
+    for (uint32_t prim_index = 0; prim_index < prim_count; prim_index += MAX_QUADS_IN_BATCH_COUNT)
+    {
+        uint32_t prims_in_batch_count = MIN(prim_count - prim_index, MAX_QUADS_IN_BATCH_COUNT);
+        for (uint32_t i = 0; i < prims_in_batch_count; i++)
+        {
+            for (uint32_t j = 0; j < 4; j++) 
+                batch_indices[i*5 + j] = first + (prim_index+i)*4 + j;
+            
+            batch_indices[i*5 + 4] = -1;
```
rasky, line 20: maybe comment that this is the sentinel to terminate the primitive (or use a macro for -1 / 0xFFFF)

src/GL/mg_ex.c
```diff
@@ -0,0 +1,82 @@
...
+    uint16_t batch_indices[QUAD_BATCH_SIZE];
+
+    const uint32_t prim_count = count / 4;
+    for (uint32_t prim_index = 0; prim_index < prim_count; prim_index += MAX_QUADS_IN_BATCH_COUNT)
+    {
+        uint32_t prims_in_batch_count = MIN(prim_count - prim_index, MAX_QUADS_IN_BATCH_COUNT);
+        for (uint32_t i = 0; i < prims_in_batch_count; i++)
+        {
+            memcpy(batch_indices + i*5, indices + (prim_index+i)*4, sizeof(uint16_t)*4);
+            batch_indices[i*5 + 4] = -1;
```
rasky, line 58: Same here

src/GL/rsp_pipeline.c
```diff
@@ -7,102 +7,129 @@
...
-        for (uint32_t i = 0; i < count; i++) {
-            int16_t value = src[i];
-            assertf(value <= SHRT_MAX>>shift && value >= SHRT_MIN>>shift, "Fixed point overflow: %d << %d", value, shift);
-            gl_cmd_stream_put_half(s, value << shift);
-        }
-    } else {
-        for (uint32_t i = 0; i < count; i++) {
-            gl_cmd_stream_put_half(s, src[i] >> -shift);
-        }
+#define GLP_NRM(x, y, z) MGFX_NRM(x, y, z);
```
rasky, line 56: Trailing semicolon in macro seems it can create problems later on

src/GL/data_cache.c
```diff
@@ -0,0 +1,91 @@
...
+{
+    *cache = (data_cache_t){0};
+    cache->source = source;
+    data_cache_invalidate(cache);
+}
+
+static void free_buffer(data_cache_t *cache)
+{
+    if (cache->buffer != NULL) {
+        rspq_call_deferred(free_uncached, cache->buffer);
```
rasky, line 15: maybe set cache->buffer=NULL here to avoid bugs causing double frees

src/GL/ringbuffer.h
```diff
@@ -0,0 +1,34 @@
...
+#include <stdint.h>
+#include <rspq.h>
+
+typedef struct {
+    uint8_t *buffer;
+    rspq_syncpoint_t *syncpoints;
+    uint32_t entry_size;
+    uint32_t entry_count;
+    uint32_t current_index;
+} ringbuffer;
```
rasky, line 17: `ringbuffer_t`

src/GL/fnv1a.h
```diff
@@ -0,0 +1,17 @@
...
+inline uint32_t fnv1a_init()
+{
+    return 0x811c9dc5; // FNV offset basis
+}
+
+inline void fnv1a_step(uint32_t *hash, uint32_t v)
+{
+    *hash ^= v;
+    *hash *= 0x01000193; // FNV prime
+}
```
rasky, line 15: no matching extern inline for these functions

src/GL/array.h
```diff
@@ -0,0 +1,81 @@
...
+            return 2;
+        case ARRAY_COLOR:
+            return 4;
+        case ARRAY_TEXCOORD:
+            return 4;
+        case ARRAY_MTX_INDEX:
+        default:
+            return 1;
+    }
+}
```
rasky, line 75: missing extern inline

src/GL/indices.h
```diff
@@ -0,0 +1,32 @@
...
+{
+    return bounds.first + bounds.count;
+}
+
+inline bool are_bounds_included(index_bounds_t a, index_bounds_t b)
+{
+    return bounds_get_lower_inclusive(a) <= bounds_get_lower_inclusive(b) &&
+        bounds_get_upper_exclusive(a) >= bounds_get_upper_exclusive(b);
+}
+
```
rasky, line 31: missing extern inline

src/GL/vertex_layout.h
```diff
@@ -0,0 +1,48 @@
...
+
+void vertex_layout_finalize(vertex_layout_t *vl);
+
+inline uint32_t vertex_layout_get_hash(const vertex_layout_t *vl)
+{
+    return vl->hash;
+}
+
+#ifdef __cplusplus
```
rasky, line 44: missing extern inline

src/GL/vertex_layout_cache.h
```diff
@@ -0,0 +1,30 @@
...
+
+void vertex_layout_cache_init(vertex_layout_cache_t *cache);
+void vertex_layout_cache_invalidate(vertex_layout_cache_t *cache);
+void vertex_layout_cache_update(vertex_layout_cache_t *cache, const array_t *arrays);
+
+inline const vertex_layout_t *vertex_layout_cache_get_layout(const vertex_layout_cache_t *cache)
+{
+    assertf(!cache->dirty, "Vertex layout cache is dirty!");
+    return &cache->layout;
+}
```
rasky, line 24: missing extern inline

src/GL/array_object.h
```diff
@@ -0,0 +1,45 @@
...
+typedef struct draw_call_cache_s draw_call_cache_t;
+
+typedef struct gl_array_object_s {
+    array_t arrays[ARRAY_COUNT];
+    gl_buffer_object_t *element_array_buffer;
+    vertex_layout_cache_t layout_cache;
+    data_source_t vertex_data_source, mtx_index_data_source;
+    data_cache_t vertex_data_cache, mtx_index_data_cache;
+    draw_call_cache_t *draw_call_cache;
+} gl_array_object_t;
```
rasky, line 19: I think this part is the most complicated for me to understand. It would be help a block comment explaining which field invalidates which cache and why.

src/GL/data_source.h
```diff
@@ -0,0 +1,44 @@
...
+    ARRAY_MASK_NORMAL = 1 << ARRAY_NORMAL,
+    ARRAY_MASK_COLOR = 1 << ARRAY_COLOR,
+    ARRAY_MASK_TEXCOORD = 1 << ARRAY_TEXCOORD,
+    ARRAY_MASK_MTX_INDEX = 1 << ARRAY_MTX_INDEX,
+} array_mask_t;
+
+typedef struct {
+    uint32_t offsets[ARRAY_COUNT];
+    uint32_t stride;
+} data_layout_t;
```
rasky, line 20: It seems that you're then not using array_type_t to index offsets[], but rather a compact index. This maybe also warrants various comments as it can be tricky to get right.

src/GL/array_convert.h
```diff
@@ -0,0 +1,26 @@
...
+#include "array.h"
+#include "data_source.h"
+
+typedef struct array_convert_parms_s {
+    const array_t *arrays[ARRAY_COUNT];
+    uint32_t array_count;
+    const data_layout_t *out_layout;
+    index_bounds_t range;
+    void *out_buffer;
+} array_convert_parms_t;
```
rasky, line 14: I could use some documentation of this structure and how the various fields relate.

src/GL/draw_call_cache.h
```diff
@@ -0,0 +1,47 @@
...
+
+typedef struct draw_call_parms_s {
+    uint32_t offset;
+    uint32_t count;
+    GLenum mode;
+} draw_call_parms_t;
```
rasky, line 16: Can you also document this structure? I think offset is a byte offset, while count is an element count. This is non-trivial.

src/GL/rsp_gl_pipeline.rspl
```diff
@@ -0,0 +1,573 @@
...
+
+//#define TRACE_VTX_LOOP          1
+
+#define SCRATCH_MAT_DIFFUSE  0x30
+#define SCRATCH_MAT_AMBIENT  0x40
+#define SCRATCH_MAT_EMISSIVE 0x50
+#define SCRATCH_RGBA_IN      0x60
```
rasky, line 10: I guess these offsets are into a scratch area? But we don't have a RSPL include file from Magma that defines the base offset for it?

src/GL/rsp_gl_pipeline.rspl
```diff
@@ -0,0 +1,573 @@
+include "rsp_magma.inc"
```
rasky, line 1: Maybe a top-level comment explaining what this ucode does, how it interacts with the others, etc.

rasky (commented):
I took the time of going through the PR. I didn't have time to read it all or rather I don't understand the full architecture. So I just jumped here and there, studied a bit of code irrespective of the context, and just browsed it in the editor leaving comments. I think I might have spotted a few bugs from patterns that looked unconvincing but again, missing context, it's a bit hard for me to make sure a bug is a bug.

src/GL/rsp_pipeline.c
snacchus: Well spotted! That was indeed a bug.

src/GL/cpu_pipeline.c
snacchus: `GL_SHORT_5_6_5_N64` is a new addition for the magma port. I'm not sure what you mean by the last question. It is currently only used for normals, which are signed.

src/GL/rsp_gl2.rspl
snacchus: You're right, I was lazy on this one. Should I reinstate it before we merge this or can it wait for a follow up PR?

src/GL/rsp_pipeline.c
snacchus: The problem is that without VBOs, there is no guarantee that the index buffer contents stay the same between frames, so we just have to re-scane every time. I can't really think of another way.

src/GL/rsp_gl_pipeline.rspl
snacchus: Maybe we could do one better and provide a bash script that will call RSPL with the right parameters. People would just need to provide their RSPL installation path. ~~Is there a good idiomatic way of doing this?~~ Nevermind, we already do this with scripts/build_mgfx.sh. Should I add a similar build_gl_pipeline.sh script?

src/GL/gl.c
snacchus: This is the old problem of initializing module dependencies again. I agree, it would be better to also close magma here, but there is still the edge case of someone, for some reason, using both GL and magma (I think it's possible in theory). As in, they init magma first, then GL. Later, they close GL and get surprised that magma is also closed. The chances of someone coming across this are small I guess. I think it makes more sense to just close magma here. What do you think?

src/GL/data_cache.c
snacchus: Yes, that's correct. I agree that this is not ideal at the moment. We could use some alternative allocation strategy instead, like re-using old buffers from previous frames, or an arena allocator etc.

src/GL/pipelines.c
snacchus: There is already lazy pipeline creation. This function is responsible for creating the pipelines for one vertex layout. It has to create multiple variants in a loop because: Which variant ends up being used depends on runtime states that might be changed outside of a display list. That's why the final pipeline to be used needs to be determined dynamically on the RSP.

src/GL/ringbuffer.c
snacchus: Yeah this is only called when GL as a module is being closed. Deferring this doesn't seem necessary to me.

src/GL/rsp_gl_pipeline.rspl
snacchus: The base offset is `MG_SCRATCH_MEM`, which is defined in rsp_magma.inc. It is included here in RSPL as `extern u16 MG_SCRATCH_MEM;`. Are you saying it should be moved into a dedicated RSPL include file?

src/GL/rsp_gl_pipeline.rspl
snacchus: I've added such a script

src/GL/rsp_gl_pipeline.rspl
snacchus: Well spotted again, I messed up implementing the lighting formula correctly somehow. To verify this more easily I plan on adding some more unit tests.

src/GL/rsp_pipeline.c
snacchus: Added the missing filtering

src/GL/vertex_layout.h
snacchus: I've added some docs. Does it make sense now?

src/GL/rsp_pipeline.c
snacchus: yes, fixed

src/GL/mg_ex.c
snacchus: The sentinel value (aka primitive restart index) is now exposed through magma's API

src/GL/array_object.h
snacchus: Added some docs. Please tell me if they're helpful!

src/GL/data_source.h
snacchus: Added docs

src/GL/array_convert.h
snacchus: Added docs

src/GL/draw_call_cache.h
snacchus: Added docs

src/GL/rsp_pipeline.c
snacchus: I've updated the list of breaking changes above.

snacchus:
@rasky I'm done going through your comments now. I fixed a bunch of them and left some question on others. There is still one TODO left on my end, which is fixing the RSP lighting formula.

src/GL/rsp_gl_pipeline.rspl
snacchus: This should be fixed now.

PR #909, opened 2026-06-16
Fix errors appearing with GCC 17

grawlinson:
I'm using a snapshot of GCC 17 (`mips64-elf-gcc (GCC) 17.0.0 20260524 (experimental)`) and I've noticed that libdragon fails to compile due to specific errors:

* `error: use of logical '&&' with constant operand`
* `error: use of logical '||' with constant operand`


rasky:
Thanks!

PR #910, opened 2026-06-17
Add helloworld example, prints "Hello World!" using rdpq

Dragorn421:
To be ported to preview too

I have no idea how many comments we want in the Makefile / in the .c


grawlinson:
More comments don't hurt! I know I find documentation helpful. :D

Issue #855, opened 2026-04-02
build toolchain workflow is missing XZ compression support


joneqdaniel:
XZ format downloads work for toolchain script, but workflow check scripts are missing dependencies.
Adding XZ compression package dependency install for different operating systems is necessary.

PR #917, opened 2026-06-26
version bump and GCC snapshot support

lndpj:
- Use gcc mirror with snapshots and get latest version from index.html
- Switch to binutils-with-gold
- Update all other packages versions to latest working release
- Use newlib-4.5.0.20241231
- Add permissions to decompress flags
- Add xz to install workflow

**Changes**
- GCC_V has to be initialized with major version number for snapshots.

**Status**
- [x] GCC 16 snapshots
- [x] binutils-with-gold
- [ ] #850
- [ ] #918 

**Outdated**
- #851
- #855
- #709

### DIFF
--- a/.github/workflows/build-toolchain-matrix.yml
+++ b/.github/workflows/build-toolchain-matrix.yml
@@ -50,6 +50,10 @@ jobs:
           sudo apt-get install texinfo
           sudo apt-get install libmpc-dev
           sudo apt-get install squashfs-tools
+          sudo apt-get install liblzma-dev
+          sudo apt-get install xz-utils
+          sudo apt-get install xzdec
+
           # If there are other dependencies, we should add them here and make sure the documentation is updated!
 
       - name: Install x-compile system build dependencies
@@ -85,6 +89,7 @@ jobs:
         with:
           path: |
             ./tools/**/*.tar.gz
+            ./tools/**/*.tar.xz
             ./tools/**/*.tar.bz2
           key: ${{ runner.os }}-dependency-downloads # TODO: concurrency errors may currently occur due to matrix, but they are the same files.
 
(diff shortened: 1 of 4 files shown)


rasky:
Hello, thanks of the PR. We don't want to switch to GCC 16 yet as we're waiting for GCC 16.2 in August that will include the patch that allows to actually bootstrap mips64-elf (it's been broken during the whole GCC 15 cycle).

Moreover, GCC 14.4 should come out any day now, which includes another important patch, which allows to bootstrap using GCC 16 as host compiler. We plan to switch to this one.

As for the other changes:
* We don't want to switch to binutils-gold. Gold is deprecated as far as I can tell.
* We don't want to automatically update version depending on an external website. We do want to pin the specific versions we want people to use and that we tested that they work.
* We are not updating newlib anymore as we plan to switch to picolibc as soon as somebody figures it out the dual-lib issue
* I prefer not to switch to xz if possible because it is an additional dependency needed on the host PC. It's true that it saves download time, but people don't rebuild toolchains that often (if even, many prefer downloading a binary one that we publish anyway).

lndpj:
Please **close** the **outdated** entries of my other non-accessible user id.

lndpj:
Current snapshots of 14 and 16 flawlessly pass building/running the toolchain along with examples in ares.
15 snapshot is broken because of missing backport of fix #709 in 17
I've only tried 16 snapshot as host for all the targets.

Adding a FORCE_SNAPSHOT_VERSION solves the `depending on an external website` problem.
```sh
FORCE_SNAPSHOT_VERSION=${FORCE_SNAPSHOT_VERSION:-""}
# Use FORCE_SNAPSHOT_VERSION if set and not 15 branch
if [[ "$FORCE_SNAPSHOT_VERSION" != "" && "${FORCE_SNAPSHOT_VERSION:0:2}" != "15" ]]; then
    GCC_V=$FORCE_SNAPSHOT_VERSION
    test -f "$DOWNLOAD_PATH/gcc-$GCC_V.tar.xz"           || download "https://gcc.gnu.org/pub/gcc/snapshots/$GCC_V/gcc-$GCC_V.tar.xz"
elif [[ "${GCC_V:0:2}" != "15" ]]; then
    test -f "$DOWNLOAD_PATH/index.html"                  || download "https://gcc.gnu.org/pub/gcc/snapshots/LATEST-${GCC_V:0:2}/index.html"
    GCC_V=`cat "$DOWNLOAD_PATH/index.html" | grep "<title>GCC ${GCC_V:0:2}-" | cut -f2 -d\ `
    test -f "$DOWNLOAD_PATH/gcc-$GCC_V.tar.xz"           || download "https://gcc.gnu.org/pub/gcc/snapshots/LATEST-${GCC_V:0:2}/gcc-$GCC_V.tar.xz"
else
    echo "ERROR: Incompatible GCC version!"
    exit 1
fi
test -d "$BUILD_PATH/gcc-$GCC_V"                     || tar -xJpf "$DOWNLOAD_PATH/gcc-$GCC_V.tar.xz" -C "$BUILD_PATH"
```

PR #925, opened 2026-07-05
Add FlashRAM save storage driver

meeq:
New flashram.c/.h module for the 128 KiB (1 Mibit) Macronix-family FlashRAM save chip found in some N64 cartridges. Unlike SRAM it is command-driven (identify / read / status / sector-erase / page-program) rather than flat memory; the protocol matches real hardware and is compatible with ares and SC64.

Two API levels: a low-level page/sector interface, and a high-level flat byte-range flashram_read/flashram_write that behaves like SRAM. flashram_write does a per-sector read-modify-write so a partial-sector write preserves the untouched bytes in that 16 KiB erase sector. The driver manages its own PI-DMA cache coherency.


include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define FLASHRAM_ADDRESS     0x08000000     ///< Base address of FlashRAM in PI address space (shared with SRAM domain).
+#define FLASHRAM_SIZE        0x00020000     ///< Total FlashRAM size (128 KiB / 1 Mibit).
+#define FLASHRAM_PAGE_SIZE   0x00000080     ///< Program granularity (128 bytes).
+#define FLASHRAM_SECTOR_SIZE 0x00004000     ///< Erase granularity (16 KiB).
+#define FLASHRAM_NUM_PAGES   (FLASHRAM_SIZE / FLASHRAM_PAGE_SIZE)     ///< Number of programmable pages (1024).
+#define FLASHRAM_NUM_SECTORS (FLASHRAM_SIZE / FLASHRAM_SECTOR_SIZE)   ///< Number of erase sectors (8).
```
rasky: API-wise I'd rather be a little more flexible for future homebrew, in case somebody wants to attempt support for flashes with different sizes. Instead of hardcoding these, we should probably have a description structure being returned by `flashram_detect()` when a flash is detected. I see it returns `FLASHRAM_SIZE` but that gives not enough information on the layout.

include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+ */
+int flashram_detect(void);
+
+/**
+ * @brief Read the FlashRAM status register
+ *
+ * @return The 32-bit status word. The low bits report program/erase busy and
+ *         completion state (see the implementation for the bit layout).
+ */
+uint32_t flashram_status(void);
```
rasky: This is maybe a bit too low-level. The exact layout of the flashram registers is probably better hidden within the module. What's the use case for having this function in the public API?

include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * @brief Read data from FlashRAM
+ *
+ * Reads a byte range from FlashRAM into @p dst. Any offset and length are
+ * accepted; the read is served entirely via PI DMA (FlashRAM array data cannot
+ * be read with CPU I/O).
```
rasky: I don't think this is generally true. Maybe this is true for word-sized flashes? Anyway it seems like an implementation detail that shouldn't be part of the API documentation.

include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * This is a blocking operation and can take on the order of tens of
+ * milliseconds per touched sector on real hardware.
+ *
+ * @param src    Source buffer containing the data to write.
+ * @param offset Byte offset in FlashRAM to write to (0 to #FLASHRAM_SIZE - 1).
+ * @param len    Number of bytes to write.
+ * @return Number of bytes written, or a negative value on error (sets @c errno).
+ */
+int flashram_write(const void* src, size_t offset, size_t len);
```
rasky, line 151: I have two main discussion points on this API:

* A blocking API for an operation that takes tens of milliseconds seems rather unfortunate. Multithreading is still a hard sell, if we could design an asynchronous API, we would probably be in a better situation.
* I understand the idea of providing a "higher level API" here but this one seems to be very hard to be successfully used:
  * With the default layout, up to 1/8th of the flash would need to be cleared for any write basically
  * This means that internally this function has to allocate a buffer of non trivial size for the RMW cycle
  * The operation is extremely slow; callers will want to minimize the number of unnecessary sector clears
  * There is no context to know whether a sector really needs to be cleared or can just be "appended" by programming more pages. So this API has to assume the worst.

It seems to me this is not really ideal. I doubt there will be many users that will want to cope with these compromises.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    if (__flashram_inited)
+        return;
+
+    // Configure PI DOM2 registers to enable access to FlashRAM (same timing as SRAM).
+    disable_interrupts();
+    *PI_BSD_DOM2_LAT = 0x5;
+    *PI_BSD_DOM2_PWD = 0xc;
+    *PI_BSD_DOM2_PGS = 0xd;
+    *PI_BSD_DOM2_RLS = 0x2;
+    enable_interrupts();
```
rasky: Unfortunately the way PI domains are designed makes them a bit unfortunate to use. Here you're basically assuming that nobody else will need to access any other peripheral in the Domain 2, during the whole run. This might be true and even likely, but it's still a very strict requirements.

In theory we should work towards a better lower-level PI API where we describe peripherals and setup domain parameters for them, and then lazily update PI registers when needed. Anyway this is probably not required to be sorted out now.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    uint32_t id[2] __attribute__((aligned(16))) = {0};
+    flashram_command(FLASHRAM_CMD_IDENTIFY_MODE);
+    data_cache_hit_writeback_invalidate(id, sizeof(id));
+    dma_read_raw_async(id, FLASHRAM_ADDRESS, sizeof(id));
+    dma_wait();
+
+    // Restore read mode so a later CPU/DMA read returns array data.
+    flashram_command(FLASHRAM_CMD_READ_MODE);
+
+    return (id[0] == FLASHRAM_IDENTIFIER) ? FLASHRAM_SIZE : 0;
```
rasky: I think here there should be multiple valid identifiers? Also there's the whole issue of byte-access vs word-access that should probably be handled. I think we can leave it unhandled if you don't want to tackle that at first, but we should then probably document in the public API that only one specific model of flash is supported for now?

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+        return -1;
+    }
+    if (len == 0)
+        return 0;
+    // PI DMA requires a 2-byte-aligned source (FlashRAM) address.
+    if (offset & 1)
+    {
+        errno = EINVAL;
+        return -1;
+    }
```
rasky: I'm a bit doubtful whether we should use errno here or use assertions. Normally we use assertions in most libdragon APIs that shouldn't fail, and only use errno in the POSIX codepaths.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+                result = -1;
+                break;
+            }
+        }
+        else
+        {
+            // Partial sector: read-modify-write to preserve untouched bytes.
+            if (sector_buf == NULL)
+            {
+                sector_buf = memalign(16, FLASHRAM_SECTOR_SIZE);
```
rasky: isn't this a good use case for scratch memory?

rasky (commented):
First round mainly focusing on public API, and just a quick look at implementation

include/flashram.h
bsmiles32: One possibility that I used during my experiments was to describe flash layout using number of bits per component: https://github.com/bsmiles32/g64drive/blob/feat/ultrasave_wip/ultrasave/flash/layout.go#L10-L45
This could describe both byte and word based flash, and could have been extended to other page/sector sizes.

include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * @brief Detect whether FlashRAM is present in the cartridge
+ *
+ * Reads the chip's silicon ID and checks it against the known FlashRAM
+ * identifier. Unlike SRAM detection this is non-destructive (it does not write
+ * to the save area).
+ *
+ * @return #FLASHRAM_SIZE if FlashRAM is detected, 0 otherwise.
+ */
+int flashram_detect(void);
```
bsmiles32: Should we also return the manufacturer / model or their ID ? and as suggested the flash layout instead of just its size ?

include/flashram.h
bsmiles32: At the hardware level, the status register is only 8bit (which happens to be accessed as 32bit because of CPU I/O). So we may return just an uin8_t here (and hope the compiler doesn't make us pay too much for the uint32_t -> uin8_t cast)

bsmiles32: Also, if you expose a low-level API for the FlashRAM you probably also want to expose a "clear status" function because in case of error, status should be cleared before attempting another program/erase operation.

include/flashram.h
bsmiles32: FlashRAM array can be accessed just fine with CPU I/O. There are 2 concerns w.r.t. read array mode: the distinction between word-based or byte-based flash layout (word based layout need address adjustment) and the 256-page boundary crossing (only happens for DMA). See https://n64brew.dev/wiki/Flash for details.

include/flashram.h
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * @brief Erase a single FlashRAM sector
+ *
+ * Erases one 16 KiB sector, setting all of its bytes to 0xFF. Blocking.
+ *
+ * @param sector Sector index (0 to #FLASHRAM_NUM_SECTORS - 1).
+ * @return true on success, false on timeout/failure (sets @c errno on bad args).
+ */
+bool flashram_erase_sector(unsigned int sector);
```
bsmiles32: At the hardware level the erase sector command takes a "page number" not a sector number. Effective sector is deduced from the page number. Do we want to use the same formalism at API level and only expose page numbers (and provide some helper functions to convert / work with sectors) ?

NOTE: I had a similar similar choice to make during my experiments in go and I went with page numbers everywhere and having some helper functions (based on chip layout) to return sector being/end pages and so on. This also led to the use of page ranges (also useful for read array operation) which I decided to encode as [begin,end[. I felt that using page numbers everywhere was matching the hardware better.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
+/**
+ * @file flashram.c
+ * @brief FlashRAM access functions for N64 cartridges
+ *
+ * FlashRAM is a 1 Mibit (128 KiB) Macronix-family NOR flash used as save
```
bsmiles32: Do we want to support all the various models of manufactured FlashRAM and detect at runtime which model it is and adapt behavior at runtime (some models have quirks, some are word-based, ...) or just support a "single" variant that ares and flashcarts will support ?

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+ * storage on some N64 cartridges. It shares the PI DOM2 window with cartridge
+ * SRAM (a cart has one or the other, never both), but is driven through a
+ * command state machine rather than being flat memory:
+ *
+ *   - Command register at #FLASHRAM_ADDRESS | 0x10000: 32-bit command words are
+ *     written here as (opcode << 24) | argument, where the argument (for
+ *     erase/program) is a *page number*, not a byte offset.
+ *   - Data window at #FLASHRAM_ADDRESS: the status/ID readback and the array
+ *     read/write buffer. Array data can only be moved via 16-bit PI DMA; a CPU
```
bsmiles32: Array data can be accessed with CPU I/O not just DMA.
See my previous remark

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+ *     read of this window returns the status/ID latch, never array data.
+ *
+ * The command protocol implemented here matches real hardware (as exercised by
+ * libultra and homebrew such as N64-SwapDumper) and is compatible with both
+ * SC64 and the ares emulator.
```
bsmiles32: This remark is a bit misleading I think. If you make such claim, that means that you've tested your code against a real flash chip (eg. a MN63F8MPN for instance), not the flash emulation of your flash cart or ares (which are known for not being fully accurate).

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+/// Silicon ID word reported by FlashRAM in identify mode.
+#define FLASHRAM_IDENTIFIER      0x11118001
+
+// Command opcodes (written as (opcode) or (opcode | page_number)).
+#define FLASHRAM_CMD_STATUS_MODE     0xD2000000  ///< Enter status mode.
+#define FLASHRAM_CMD_IDENTIFY_MODE   0xE1000000  ///< Enter identify (silicon ID) mode.
+#define FLASHRAM_CMD_READ_MODE       0xF0000000  ///< Enter array read mode.
+#define FLASHRAM_CMD_SECTOR_ERASE    0x4B000000  ///< Arm sector erase at page (| page).
+#define FLASHRAM_CMD_EXECUTE_ERASE   0x78000000  ///< Execute the armed erase.
+#define FLASHRAM_CMD_PAGE_PROGRAM    0xB4000000  ///< Enter page-program mode.
```
bsmiles32: In the wiki, this command is named "Load Byte Page" (which is also what datasheets use).

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+#define FLASHRAM_COMMAND_ADDRESS (FLASHRAM_ADDRESS | 0x00010000)
+
+/// Silicon ID word reported by FlashRAM in identify mode.
+#define FLASHRAM_IDENTIFIER      0x11118001
+
+// Command opcodes (written as (opcode) or (opcode | page_number)).
+#define FLASHRAM_CMD_STATUS_MODE     0xD2000000  ///< Enter status mode.
+#define FLASHRAM_CMD_IDENTIFY_MODE   0xE1000000  ///< Enter identify (silicon ID) mode.
+#define FLASHRAM_CMD_READ_MODE       0xF0000000  ///< Enter array read mode.
+#define FLASHRAM_CMD_SECTOR_ERASE    0x4B000000  ///< Arm sector erase at page (| page).
```
bsmiles32, line 54: You could have added the full chip erase command (even if unused for now)

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+#define FLASHRAM_CMD_EXECUTE_ERASE   0x78000000  ///< Execute the armed erase.
+#define FLASHRAM_CMD_PAGE_PROGRAM    0xB4000000  ///< Enter page-program mode.
+#define FLASHRAM_CMD_EXECUTE_PROGRAM 0xA5000000  ///< Execute page program at page (| page).
+
+// Status register bits.
+#define FLASHRAM_STATUS_PROGRAM_BUSY 0x01  ///< A page program is in progress.
+#define FLASHRAM_STATUS_ERASE_BUSY   0x02  ///< A sector/chip erase is in progress.
+#define FLASHRAM_STATUS_PROGRAM_OK   0x04  ///< The last page program completed.
```
bsmiles32: PROGRAM_OK and ERASE_OK don't just reflect if an operation completed (the _BUSY bit is used for that). But it reflect if an error occurred during the last completed operation.

See https://n64brew.dev/wiki/Flash#Status_Mode

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+/// Write a command word to the FlashRAM command register.
+static inline void flashram_command(uint32_t command)
+{
+    io_write(FLASHRAM_COMMAND_ADDRESS, command);
+}
+
+uint32_t flashram_status(void)
+{
+    assertf(__flashram_inited, "flashram accessed, but flashram_init() hasn't been called yet");
+    flashram_command(FLASHRAM_CMD_STATUS_MODE);
```
bsmiles32: If you intend to be compatible with real hardware such as MX29L1100 you may need to write to the CIR twice for the change to be effective. See https://n64brew.dev/wiki/Flash#Command_Internal_Register_(CIR) (quirk note).

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    return io_read(FLASHRAM_ADDRESS);
```
bsmiles32: You may want only return the lower 8 bits as the status because :
1. the status register is really 8 bit
2. On MX29L1100 the upper 16bits of the first IO status read comes from previous mode (this is a quirk).

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+void flashram_init(void)
+{
+    if (__flashram_inited)
+        return;
+
+    // Configure PI DOM2 registers to enable access to FlashRAM (same timing as SRAM).
+    disable_interrupts();
+    *PI_BSD_DOM2_LAT = 0x5;
+    *PI_BSD_DOM2_PWD = 0xc;
+    *PI_BSD_DOM2_PGS = 0xd;
```
bsmiles32: Using a page size of "0xd" should at least come with a comment about the 256-page DMA boundary and that we don't intent to support word based FlashRAM.
https://n64brew.dev/wiki/Flash#PI_Configuration

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+
+/// Command register address (offset 0x10000 into the FlashRAM window).
+#define FLASHRAM_COMMAND_ADDRESS (FLASHRAM_ADDRESS | 0x00010000)
+
+/// Silicon ID word reported by FlashRAM in identify mode.
+#define FLASHRAM_IDENTIFIER      0x11118001
+
+// Command opcodes (written as (opcode) or (opcode | page_number)).
+#define FLASHRAM_CMD_STATUS_MODE     0xD2000000  ///< Enter status mode.
+#define FLASHRAM_CMD_IDENTIFY_MODE   0xE1000000  ///< Enter identify (silicon ID) mode.
```
bsmiles32: In the wiki (and datasheets), this one is named SiliconID mode.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    flashram_command(FLASHRAM_CMD_READ_MODE);
+}
+
+int flashram_detect(void)
+{
+    assertf(__flashram_inited, "flashram accessed, but flashram_init() hasn't been called yet");
+
+    // Enter identify mode and DMA out the two silicon-ID words.
+    uint32_t id[2] __attribute__((aligned(16))) = {0};
+    flashram_command(FLASHRAM_CMD_IDENTIFY_MODE);
```
bsmiles32: Same as previous comment about quirks on real HW : switching to SiliconID mode may require 2 consecutive writes to CIR to be effective on some chips.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+        return 0;
+    // PI DMA requires a 2-byte-aligned source (FlashRAM) address.
+    if (offset & 1)
+    {
+        errno = EINVAL;
+        return -1;
+    }
+
+    flashram_command(FLASHRAM_CMD_READ_MODE);
+
```
bsmiles32, line 320: This only works on byte-based flash, not word-based one. Maybe add a comment about that.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    if (sector >= FLASHRAM_NUM_SECTORS)
+    {
+        errno = EINVAL;
+        return false;
+    }
+
+    unsigned int page = sector * FLASHRAM_PAGES_PER_SECTOR;
+    flashram_command(FLASHRAM_CMD_SECTOR_ERASE | page);
+    flashram_command(FLASHRAM_CMD_EXECUTE_ERASE);
+    return flashram_wait_ready(FLASHRAM_STATUS_ERASE_BUSY, FLASHRAM_STATUS_ERASE_OK,
```
bsmiles32: You should probably clear the status register before returning from this function.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    // Copy into an aligned bounce so callers can pass any alignment.
+    uint8_t buffer[FLASHRAM_PAGE_SIZE] __attribute__((aligned(16)));
+    memcpy(buffer, data, FLASHRAM_PAGE_SIZE);
+
+    flashram_command(FLASHRAM_CMD_PAGE_PROGRAM);
+    data_cache_hit_writeback(buffer, FLASHRAM_PAGE_SIZE);
+    dma_write_raw_async(buffer, FLASHRAM_ADDRESS, FLASHRAM_PAGE_SIZE);
+    dma_wait();
+    flashram_command(FLASHRAM_CMD_EXECUTE_PROGRAM | page);
+    return flashram_wait_ready(FLASHRAM_STATUS_PROGRAM_BUSY, FLASHRAM_STATUS_PROGRAM_OK,
```
bsmiles32: Same comment, you should probably clear status before returning from this function.

src/flashram.c
```diff
@@ -0,0 +1,331 @@
...
+    return io_read(FLASHRAM_ADDRESS);
+}
+
+/// Poll the status register until @p busy_mask clears; returns whether @p ok_mask is then set.
+static bool flashram_wait_ready(uint32_t busy_mask, uint32_t ok_mask, uint32_t timeout_ms)
+{
+    uint64_t start_ms = get_ticks_ms();
+    while (true)
+    {
+        uint32_t status = flashram_status();
```
bsmiles32: Calling flashram_status will cause a write to CIR to switch to status mode. In the case of erase and program operations, WSM (Write State Machine) will "automatically" switch to status mode, so you don't have to write anything to CIR to poll status register for completion/errors.

bsmiles32 (commented):
Thanks for working on this ! Just did a first round of review.

gobbledygoober:
Not necessarily critical for this PR, but I wanted to capture the suggestion from Discord to add an example that leverages this feature.

gobbledygoober:
> Not necessarily critical for this PR, but I wanted to capture the suggestion from Discord to add an example that leverages this feature.

The new savedoodle example is very cool! Nice work constructing an elegant way to exercise the various save types.

src/flashram.c
meeq: I didn't use scratch memory for this because I wasn't sure if we were going to move forward with the eventual consistency model. If the allocation needs to live longer than the function call, scratch becomes riskier.

For a short-lived allocation in a synchronous function with no other allocations it doesn't really make a difference.

meeq:
Related: #951

PR #939, opened 2026-07-10
fix/sram_main_include

gobbledygoober:
Follow-on task for https://github.com/DragonMinded/libdragon/pull/813 (initial SRAM support).

Per [discussion in Discord](https://discord.com/channels/205520502922543113/1523308991294472374/1524455789958533270), adding `sram.h` to `libdragon.h` wrapper. 

Also added a commit to sort the list of include files alphabetically for readability.

### DIFF
--- a/include/libdragon.h
+++ b/include/libdragon.h
@@ -112,5 +112,6 @@
 #include "magma.h"
 #include "magma_math.h"
 #include "mgfx.h"
+#include "sram.h"
 
 #endif


rasky:
Thanks! Can you please drop the second commit? I'd like to group the include by area rather than alphabetically. They are roughly like that already, but we could push this further. They could also logically match the structure of the `src` subdirectories (which we should add a couple). Anyway, better drop it for now.

PR #941, opened 2026-07-13
audio: add a check for whether the calculated DAC rate will fit

nadiaholmquist:
Specifying a too low frequency can lead to audio_init calculating a DAC rate value that won't actually fit in the register (the value is only 14-bit). This will only happen if someone gives a ridiculously slow audio frequency, but it's a bug nonetheless.

I ran into this because I was trying to play tones at specific frequencies without involving any resampling, and I was surprised when values that were too small would lead to completely wrong output.

### DIFF
--- a/src/audio.c
+++ b/src/audio.c
@@ -201,6 +201,7 @@ void audio_init(const int frequency, float latency)
        is also used for the AI output), divided by the requested frequency,
        rounding up. */
     int dacrate = ((2 * clockrate / frequency) + 1) / 2;
+    assertf(dacrate <= (1 << 14), "Requested frequency %d is too low", frequency);
     /* Bitrate is the half period for each bit of the sample. We need to send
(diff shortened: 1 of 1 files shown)


src/audio.c
```diff
@@ -201,6 +201,7 @@ void audio_init(const int frequency, float latency)
        is also used for the AI output), divided by the requested frequency,
        rounding up. */
     int dacrate = ((2 * clockrate / frequency) + 1) / 2;
+    assertf(dacrate <= (1 << 14), "Requested frequency is too low, a DAC rate of %d would overflow the register", dacrate);
```
rasky: Can you please show the requested frequency here rather than the calculated dacrate? The error should be readable to the programmer, who is likely not aware of what the dacrate is.

I would phrase the error something like "Requested frequency %d is too low" or something like that

rasky (commented):
Thanks, makes sense to add the assert

rasky:
Thanks!

PR #942, opened 2026-07-14
dma: allow DMA to cart offsets >=0x20000000

meeq:
`dma_read_async` takes the CPU-copy path for `io_accessible()` PI addresses, mapping the PI address through KSEG1 (`pi_address | 0xA0000000`) and then `PhysicalAddr()`, which masks to 29 bits. For a cart address >= `0x20000000` (e.g. a block streamed from a high offset of a >256MiB ROM) that silently drops the high bit and reads the wrong location. Report such addresses as not CPU-accessible so the aligned raw-DMA path is used with the full 32-bit PI address instead.

### DIFF
--- a/src/dma.c
+++ b/src/dma.c
@@ -48,6 +48,15 @@ bool io_accessible(pi_addr_t pi_address)
     if (pi_address >= 0x1FC00000 && pi_address <= 0x1FCFFFFF)
         return false;
 
+    // The CPU-copy path maps the PI address through KSEG1 (pi_address | 0xA0000000),
+    // which only reaches physical addresses below 0x20000000. A cart address at or
+    // above that (e.g. a large disc image streamed from a high cart offset on a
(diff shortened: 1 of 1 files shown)


rasky:
Physical addresses above 0x20000000 up to 0x7fffffff are I/O accessible using 64-bit pointers. See `io_read` / `io_write` that supports them. There is indeed a bug here: the misaligned path in `dma_read_async` doesn't know about this, and uses 32-bit pointers to do the CPU transfers, effectively causing the bug that you found.

The fix here is to make the misaligned path in `dma_read_async`  use 64-bit virtual pointers: `vaddr64_t`, `VirtualUncachedAddr64`, and `sys_vaddr_*()` functions.

meeq:
Waiting on #951 to revisit this approach

rasky:
Fixed in #951

PR #943, opened 2026-07-15
wav64: ensure NULL huff_tbl for vadpcm streams

meeq:
`wav64_vadpcm_init` only sets `vhead->huff_tbl` (via `init_huffman`) when the `VADPCM_FLAG_HUFFMAN` bit is set. `huff_tbl` is a runtime pointer that physically lives inside the file-loaded packed ext header, so for a non-huffman stream it retains its file bytes (which *should be* `0`/`NULL`). `wav64_vadpcm_close` then does `if (huff_tbl) free(huff_tbl)`, freeing a garbage pointer -- a wild `free()` that can corrupt the dlmalloc heap and manifests intermittently as corruption far from the source. NULL it in the non-huffman branch.

### DIFF
--- a/src/audio/wav64_vadpcm.c
+++ b/src/audio/wav64_vadpcm.c
@@ -422,6 +422,7 @@ void wav64_vadpcm_init(wav64_t *wav, int state_size)
 
     // Init huffman
     wav64_header_vadpcm_t *vhead = (wav64_header_vadpcm_t*)wav->st->ext;
+    assertf(vhead->huff_tbl == NULL, "huff_tbl must be NULL before initialization");
     if (vhead->flags & VADPCM_FLAG_HUFFMAN) {
         wav64_vadpcm_init_huffman(wav);
     }


meeq:
I'm still trying to figure out how exactly I managed to get a non-VADPCM wav64 with a non-zero value in huff_tbl -- this fix may not strictly necessary (maybe it should be an assertion instead).

rasky:
A non-VADPCM wav64 doesn't technically have a huff_tbl field at all.

A VADPCM, non-huffman wav64 should always have an empty huff_tbl pointer on disk:

```C
		w16(out, 0); // padding
		w32(out, 0); // huff_tbl_ptr
		w32(out, skip_points.size() > 0 ? CODEBOOK_SIZE*16 : 0); // skip_points_ptr
```

You can PR the assertion if you want and can help debugging your issue, but the PR as-is shouldn't be necessary.

rasky (changes requested):
You can PR the assertion if you want and can help debugging your issue, but the PR as-is shouldn't be necessary.

meeq:
Agreed that the fix is not necessary: this turned out to be caused by unrelated stale pointer corruption. I switched the path to an assertion so that it fails loudly for the next poor soul.

PR #944, opened 2026-07-15
mkmaterial: Miscellaneous fixes

### LINKED ISSUE #924: mkmaterial: bad blender handling
The mkmaterial handling of the blender is buggy

## Bug 1

`MATFLAG_BLENDER` is never set by `mkmaterial_export.cpp`

I think it should always be set, like the combiner

## Bug 2

`mkmaterial_export.cpp`: `w8(f, mode);` writes a u8
`rdpq_mat.c`: `rdpq_blender_t bl = FETCH(mat, uint32_t);` reads a u32

I think `mkmaterial_export.cpp` ought to write a u32 as the spec at the top of `rdpq_mat.c` mentions "4 bytes"

## Bug 3

`mkmaterial_export.cpp`: `int mode = bl.mode;` is an index into the blender mode enum
but it is then passed as a blender expression:
`rdpq_mat.c`: `rdpq_mode_blender(bl);`

I think `mkmaterial_export.cpp` should convert the mode enum to a blender expression and store the blender expression instead.
Note in fixing this, it would be interesting to also think of implementing a `blender.mode.raw` jmat property

snacchus:
This PR does the following:
- Removes the material keys "combiner.reg.keyscale" and "combiner.reg.keycenter", due to them not being correctly implemented and raising an assert at runtime.
- Fixes #924
- Fixes a bug with materials that only contain a blender formula but no other render mode overrides, which caused the rdpq mode not to be reset properly by `rdpq_mat_draw_end`.
- Fixes a bug with materials that don't override the combiner, which caused an invalid combiner to be written.
- Adds tests for some of the cases described above.

### DIFF
--- a/tools/mkmaterial/mkmaterial_export.cpp
+++ b/tools/mkmaterial/mkmaterial_export.cpp
@@ -184,6 +184,7 @@ void Material::write(FILE *f)
 
     uint16_t flags = MATFLAG_COMBINER; // always write the combiner
     if (tex[0] || tex[1])       flags |= MATFLAG_TEXTURE;
+    if (bl.mode > 0)            flags |= MATFLAG_BLENDER;
     if (rm.antialias >= 0)      flags |= MATFLAG_RMO_AA;
     if (rm.fog >= 0)            flags |= MATFLAG_RMO_FOG;
     if (rm.dither[0] >= 0)      flags |= MATFLAG_RMO_DITHERING;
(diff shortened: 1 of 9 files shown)


tools/mkmaterial/mkmaterial_export.cpp
```diff
@@ -184,6 +184,7 @@ void Material::write(FILE *f)
 
     uint16_t flags = MATFLAG_COMBINER; // always write the combiner
     if (tex[0] || tex[1])       flags |= MATFLAG_TEXTURE;
+    if (bl.mode >= 0)           flags |= MATFLAG_BLENDER;
```
Dragorn421: Can bl.mode be negative ? (i.e. when does a material not set the blender?)

snacchus: oh good point, that was supposed to be `>`

PR #946, opened 2026-07-15
Fix optional blender for mkmaterial materials

Dragorn421:
discussion on discord here: https://discord.com/channels/205520502922543113/1520425166482440324/1526932041621438575

This changes mkmaterial materials to not interpret the "none" blender as "unset" (current libdragon preview), but instead interpret the lack of blender property in the jmat as "unset", and "none" as the no-blending option.

### DIFF
--- a/tests/assets/basic.mat
+++ b/tests/assets/basic.mat
@@ -17,6 +17,9 @@ rm.alpha_compare = 1
 
 [empty]
 
+[blender_none]
+blender.mode = none
+
 [blender_multiply]
 blender.mode = multiply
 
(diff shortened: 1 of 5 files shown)


Dragorn421:
Note: no idea why CI is failing. it's mkmaterial failing an assert, seemingly from trying to write prim_lod_frac values (when there's nothing that uses/sets prim_lod_frac in the first place). Can't reproduce locally...

Dragorn421:
> Note: no idea why CI is failing. it's mkmaterial failing an assert, seemingly from trying to write prim_lod_frac values (when there's nothing that uses/sets prim_lod_frac in the first place). Can't reproduce locally...

fixed, turns out `CombinerRegister::is_set` was uninitialized

PR #951, opened 2026-07-20
dma: implement an async PI DMA queue

rasky:
Implement an asynchronous PI DMA queue, that is processed under interrupt. It works like this:

* `dma_read_async` and `dma_write_async` enqueue a transfer, without waiting for any existing transfer to finish. They are changed to return a 64-bit ticket, that refers to the transfer being scheduled.
* The queue is currently hardcoded to 16 slots. If more than 16 transfers are enqueued, the 17th `_async()` call will wait for the first one to finish.
* `dma_wait()` waits for all enqueue transfer to be finished, which is the most backward compatible behavior.
* `io_read()` / `io_write()` currently wait for all enqueue transfer to finish first, so they act as full barrier (they call `dma_wait()` internally, basically). This is also a very cautious backward compatibility approach. It could be revisited if needed.
* New functions: `dma_wait_started(ticket)` and `dma_wait_finished(ticket)` can be used to spinlock when needed, waiting for a certain transfer to happen.
* New function: `dma_get_progress(ticket)` returns the number of bytes currently transferred. It can be used to perform DMA racing (partial processing of data while a transfer is in progress)
* New function: `dma_finished(ticket)` is a non-blocking call to check if a certain transfer is done
* All functions can be called with interrupts enabled or disabled, and perform a best-effort eager semantic. For instance calling `dma_wait_finished()` with disabled interrupts will still work correctly and make the DMA queue proceed.

### DIFF
--- a/include/dma.h
+++ b/include/dma.h
@@ -30,6 +30,12 @@
  * manipulating registers on a cartridge such as a gameshark.  Code should never
  * make raw 32-bit reads or writes in the cartridge domain as it could collide with
  * an in-progress DMA transfer or run into caching issues.
+ *
+ * Async DMA requests are queued and executed one at a time, dequeued via the
+ * PI interrupt.  Each async function returns a ticket that can be used with
+ * #dma_get_progress and #dma_finished to query the transfer,
+ * #dma_wait_started or #dma_wait_finished to wait for that transfer, or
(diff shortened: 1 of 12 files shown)


rasky:
The second commit here contains a better fix for
https://github.com/DragonMinded/libdragon/pull/942

PR #965, opened 2026-08-02
mkmaterial: Blender-related improvements

snacchus:
This PR adds the following new properties to mkmaterial:
- `blender.mode.raw`: Similar to `combiner.*.raw`, in that it allows to specify the exact blender formula (1-cycle or 2-cycle)
- Several properties that control the `BLEND_COLOR` and `FOG_COLOR` RDP registers:
  - `blender.reg.blend.rgb`: Corresponds to the RGB portion of the `BLEND_COLOR` register. Note that the alpha portion is **not** settable through a similar property, because this is already covered by `rm.alpha_compare`.
  - `blender.reg.fog.rgb`: Corresponds to the RGB portion of the `FOG_COLOR` register.
  - `blender.reg.fog.alpha`: Corresponds to the alpha portion of the `FOG_COLOR` register. This property must not be set if `blender.mode` is set to `multiply_const`, because the constant value also takes up the fog alpha.

To make it possible to set RGB and alpha portions of RDP registers independently, some new API has been added at the rdpq-level:
- `rdpq_set_blend_color_rgb`
- `rdpq_set_blend_color_alpha`
- `rdpq_set_fog_color_rgb`
- `rdpq_set_fog_color_alpha`

These functions are implemented by tracking the current register contents in DMEM (of the rdpq overlay).


tools/mkmaterial/mkmaterial.h
```diff
@@ -159,6 +188,9 @@ struct Material {
     void write(FILE *f);
 };
 
+// mkmaterial_parse_blender.cppp
```
Dragorn421:
```suggestion
// mkmaterial_parse_blender.cpp
```

src/rdpq/rsp_rdpq.S
```diff
@@ -160,6 +186,58 @@ RDPQCmd_SetPrimColor:
...
+    # The command is patched to be one of SET_FOG_COLOR or SET_BLEND_COLOR,
+    # corresponding to the original command.
+    #############################################################
+    .func RDPQ_SetColorRegisterComponent
+RDPQ_SetColorRegisterComponent:
+    # This results in the following mapping:
+    # * Set Fog Color Component -> SET_FOG_COLOR
+    # * Set Blend Color Component -> SET_BLEND_COLOR
+    lui t0, 0xE000 ^ 0xF800
+    xor a0, t0
```
Dragorn421, line 227: just for my curiosity, how does this bit work / what does it do?
I get that a0 will be the high word in the 8 bytes command sent to the rdp, so this is tweaking the rdp command id somehow

Dragorn421 (commented):
Looks good to me! Thanks!

Regarding the weirdness around blender.mode.const and blender.reg.fog.alpha both referring to the same value in fine (which it seems you handled correctly by erroring if both are set), one thing to note is that `blender.mode.const` currently has no effect (doesn't get exported into the material)
It may be a good opportunity to either get rid of `blender.mode.const` or actually implement it too

src/rdpq/rsp_rdpq.S
snacchus: Correct, it patches the command ID so that it's accepted by the RDP:
- `0xE0` is the ID of the new command `RDPQCmd_SetFogColorComponent`
- `0xF8` is the RDP command `SET_FOG_COLOR`
- XOR'ing these two IDs together results in their bit-difference: Only the bits are set which differ between them
- XOR'ing one of the ID's with the bit-difference means that exactly the bits that were different are toggled, which in turn means we get exactly the other ID
- Because `a0` contains the ID of the current command, by doing this we get the ID of the RDP command instead, which can then be forwarded to the RDP
- Note that `SET_BLEND_COLOR` comes directly after `SET_FOG_COLOR` (`0xF9`)
- By putting `RDPQCmd_SetBlendColorComponent` directly after `RDPQCmd_SetFogColorComponent`, we can map _both_ commands in the same manner to save on IMEM.
- Note that this only works because the least significant bit of the command ID is not affected by the XOR, or in other words: both IDs share the same LSB value

snacchus:
@Dragorn421 `blender.mode.const` did indeed have no effect thus far, but I actually fixed that too in this PR. See `Blender::get_fog_alpha`

src/rdpq/rsp_rdpq.S
Dragorn421: Thanks for the details!
Would it also work to do `lui t0, 0x1800 ; or a0, t0` ?

snacchus: Sure, that would work as well in this specific case. But I'd like to keep the current code for symmetry with `RDPQCmd_SetPrimColorComponent`

Dragorn421 (approved):
:tada:

PR #966, opened 2026-08-03
Initial DESTEX implementation

SpookyIluha:
Overview

**DESTEX** means **Destination-as-Texture**. It makes the current RDP render target available as one input to the color combiner.

This is useful because the N64 blender unit cannot express every operation reliably. In particular, additive blending can wrap instead of saturating. DESTEX moves the operation into the color combiner's 9-bit internal domain, which can represent values up to roughly 1.5 before the final clamp, then writes the combined result opaquely. Transparent source pixels may still be rejected with alpha compare. The same mechanism can implement addition, subtraction, multiplication, overlay and other custom effects.

The destination may be a display framebuffer, an offscreen render target, or a sub-surface. DESTEX always refers to the color image active when the draw executes.

DESTEX is not native framebuffer fetch. The RDP cannot directly sample the pixel it is currently writing without blender unit. Instead, each draw is split into small regions:

```text
source texture ───────────────────────────────┐
                                              │ TEX0
current destination ── sync ── load to TMEM ─┼─ color combiner ── opaque write
                                              │ TEX1
                                              └───────────────────────────────
```

For every region, the RDP waits for earlier destination writes, copies that region into TMEM, samples both textures, and writes the result back. This feedback dependency is fundamental; moving command generation to the RSP reduces CPU work but cannot remove the RDP sync or copy.

Useful related terms are *destination texturing*, *destination feedback*, *framebuffer-as-texture*, *framebuffer feedback*, and *emulated framebuffer fetch*.

### TMEM layout

RGBA16 layout divides TMEM in half:

```text
0x0000..0x07ff  destination chunk, replaced for every generated chunk
0x0800..0x0fff  current source chunk, replaced by the blitter as needed
```

The lower half holds one 32x32 RGBA16 destination region at a fixed 64-byte pitch. The upper half can hold, for example, a 32x32 RGBA16 source chunk, a 32x32 IA8/I8 source chunk, or a 64x64 I4 source chunk.

For a source region that fits in the upper half, the blitter uploads it once and reuses it across every intersecting destination-feedback chunk. For a larger source, it walks source chunks and gives each upload as much reuse as possible before replacing it.

CI4 and CI8 sources are not supported. Color-indexed sampling requires the global TLUT mode which isn't possible here.

RGBA32 destinations are not supported at the moment, but could be.

### Why the 2 UVs must be packed

An RDP textured primitive supplies one S/T/W coordinate stream. TEX0 and TEX1 both derive their coordinates from that stream; they do not receive independent UV sets.

For arbitrary scaling and rotation, DESTEX needs two mappings at once:

- Source UVs follow the source surface through axis-aligned scale and flips.
- Destination UVs follow absolute screen pixels so each output pixel samples the previous value underneath it.

The workaround packs both logical coordinates into different bits of the same S/T value. Tile masks and shifts then decode different pieces for TEX0 and TEX1. Example coordinates:

```text
packed = destination * 32 + source

source tile:       mask the low five bits
destination tile:  shift right by five
```

The source tile sees `source mod 32`. The destination tile effectively sees:

```text
packed / 32 = destination + source / 32
```

This is the central DESTEX trick and its main compromise: the low source component slightly perturbs the recovered destination coordinate. Centered bilinear filtering makes the error much less noticeable, but cannot remove it.

## What DESTEX solves — and what it does not

DESTEX provides:

- Destination-aware color-combiner expressions without relying on the framebuffer blender for the arithmetic.
- Additive, subtractive, multiplicative, lerped, or custom compositing through ordinary RDPQ combiner state.
- Arbitrary axis-aligned scaling, clipping, flips, and automatic splitting of larger source surfaces and destination footprints.
- Rotation when the selected source region fits the reserved source half of TMEM and the packed-coordinate limits.
- Dynamic display framebuffers and offscreen targets.
- CPU-light batches whose RDP commands are generated by the RSP.

It does not provide:

- Native or free framebuffer fetch.
- Independent hardware UV sets.
- Free, exact source clamping under the packed-coordinate scheme.
- CI4/CI8 source textures or TLUT mode in the two-texture layout.
- Removal of the per-chunk feedback sync and destination load.
- Automatic source-chunk splitting for a rotated source that does not fit the
  reserved source half of TMEM.


src/rdpq/rsp_rdpq.S
```diff
@@ -95,6 +95,9 @@ _PADDING:               .long  0
 
 RDPQ_RDRAM_STATE_ADDR:      .word  0
 RDPQ_RDRAM_SYNCPOINT_ADDR:  .word  0
+# Most recently resolved SET_COLOR_IMAGE command. This is private to the RDPQ
+# overlay: placing it in common RSPQ state consumes DMEM in every overlay.
+RDPQ_COLOR_IMAGE:           .quad  0
```
rasky: This comment is a bit too verbose: there is no need to clarify that private DMEM doesn't affect all the other overlays

src/rdpq/rsp_rdpq.S
```diff
@@ -376,15 +379,35 @@ RDPQCmd_SetColorImage:
...
+    # image active at playback time rather than the one active at recording.
+    #############################################################
+    .func RDPQCmd_SetTextureImageColor
+RDPQCmd_SetTextureImageColor:
+    lw a0, %lo(RDPQ_COLOR_IMAGE) + 0
+    lw a1, %lo(RDPQ_COLOR_IMAGE) + 4
+    lui t0, 0x0200              # SET_COLOR_IMAGE (0xFF) -> SET_TEXTURE_IMAGE (0xFD)
+    xor a0, t0
+    jal_and_j RDPQ_Write8, RDPQ_Finalize
+    .endfunc
```
rasky: There is no check that the the last set COLOR_IMAGE has the same size of the one that was first saved when the block was recorded. This means that we get a corrupted output if you record a rdpq_blendfx_blit against a 320x240 colorbuffer and reuse it over an offscreen surface of different size. 

It would be possible to pass the expected framebuffer size to RDPQCmd_SetTextureImageColor in the rdpq_write call, and then compare it here with the saved one, with an  (RSP) assert.

Not sure if it's worth, but it costs very little.

src/rdpq/rdpq_blendfx.c
```diff
@@ -0,0 +1,975 @@
...
+
+static void destex_reset_cached_state(void)
+{
+    destex_residency.valid = false;
+    destex_residency.source_tile_ready = false;
+    destex_feedback_tile.valid = false;
+    destex_feedback_image.valid = false;
+    if (destex_source_loader.valid)
+        destex_source_loader.loader.load_mode = TEX_LOAD_UNKNOWN;
+}
```
rasky: I think this cache doesn't save the current block being recorded (or not). So there might be a cache issue if somebody does something crazy like:

```C
rdpq_blendfx_multi_begin();
rdpq_blendfx_blit(A, ...);
rspq_block_begin();
rdpq_blendfx_blit(A, ...);
rspq_block_end();
```

The recorded block would be created using the residency cache (thus skipping some uploads) but when played back, it would misbehave as that cache wouldn't be loaded.

I'd rather play safe here and detect this case. I think it's sufficient to cache the value of `rspq_block` (NULL or not) and then compare it in destex_blit_scoped. Also please add a test for this case.

rasky: I myself pushed a fix for this, please double check it.

Cyd0n1a:
Awesome. 😎🌠🚀

PR #967, opened 2026-08-03
audioconv: adjust loop offset for ULC to block boundary

phoboslab:
Embedded loop points (specified with `--wav-loop-offset`) that were not at a ULC block boundary (1024 samples) would hit the "unaligned ULC seek" assert during playback. This PR adjust the loop offset to the next block boundary and appends the skipped samples to the end of the file.

I also moved `ULC_BLOCK_SIZE` to the top of the file, since there's now multiple points where we need it.

### DIFF
--- a/tools/audioconv64/conv_wav64.cpp
+++ b/tools/audioconv64/conv_wav64.cpp
@@ -49,6 +49,7 @@ double flag_wav_seek_interval_sec = 0.0;
 const char *flag_wav_seek_file = NULL;
 bool flag_wav_mono = false;
 const int OPUS_SAMPLE_RATE = 48000;
+const int ULC_BLOCK_SIZE = 1024;
 
 static bool read_wav(const char *infn, wav_data_t *out)
 {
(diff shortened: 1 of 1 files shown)


rasky:
My only issue with this is that supporting arbitrary seeking is a good idea anyway, not only for loops where you can do this workaround, but also for skip points. Surely it can't be that hard to do?

phoboslab:
Well, we could start decoding at the previous block boundary and then throw away N samples from the beginning. That's certainly possible. This would need to happen on the RSP and we still need to force 64bit alignment for DMA. Only complication is that IMEM is pretty tight and we need this for direct mono and stereo interleave paths. I'll look into it!

On the other hand, would a max 23ms discrepancy (assuming 44100hz) that big of a deal for skip points?

phoboslab:
Given that...
1. ULC can only decode full blocks
2. we need `samplebuffer_undo()` to throw away some samples
3. we need `rspq_highpri_sync()` to use `samplebuffer_undo()` safely
4. ULC only makes sense for music, speech and longer sound effects and sample-accurate seeks are thus not that important (correct if I'm wrong here and I don't see a particular use case)

... I don't feel that implementing sample-accurate seek points is a worthwhile tradeoff.

The sync could be avoided if we decode to temp buffers and copy only the needed samples into the samplebuffer, but this copy would need to run on the RSP. So that's not a good solution either.

Maybe there's another way that I'm not seeing?

rasky:
OK let's keep it like this for now

PR #975, opened 2026-08-14
Fix tools change_ext function

Dragorn421:
It was wrongly assuming that the new extension was the same length or shorter than the previous extension


rasky:
Fixed by cherry picking the preview fix

PR #977, opened 2026-08-18
rdpq: fix triangles in RSPQ blocks

meeq:
`rdpq_triangle()` uses the RSP to assemble the final RDP triangle command.

According to `include/rdpq.h`:
> RSP commands that generate RDP commands must be enqueued using `rdpq_write` instead of `rspq_write`.

Usage of `rspq_write` allows static RDP append commands to coalesce across block-recorded triangles and leaves no reserved region for the generated triangle words. Interleaved texture uploads and triangles can execute with reordered texture state or produce a malformed RDP command stream.


src/rdpq/rdpq_tri.c
```diff
@@ -438,6 +438,8 @@ void rdpq_triangle_cpu(const rdpq_trifmt_t *fmt, const float *v1, const float *v
         cmd_id |= 0x1;
     }
 
+    if (__builtin_expect(rspq_block != NULL, 0))
+        __rdpq_block_reserve(size / 2);
```
rasky: Why is this required?

meeq: Admittedly, `size / 2` is not as clear as it could be; it happens to be correct, but it's not obvious why. I'll revise to unify the cost calculation between the CPU and RSP paths.

rasky: I think my main question was why you need to call __rdpq_block_reserve explicitly. Is that because there is no rdpq equivalent for `rspq_write_begin` like there's for `rspq_write`?

meeq: Yes, precisely.

rasky: I think it might be better to add a `rdpq_write_begin` API then; it should theoretically be a small inline wrapper over `rspq_write_begin`. That seems much clearer than calling an internal API for seemingly no reason.

PR #978, opened 2026-08-21
[CI] Set LIBDRAGON_PREVIEW=0 by default

Dragorn421:
This only affects CI, the n64.mk keeps LIBDRAGON_PREVIEW ?= 2 (where `?=` means we can override for CI by setting `LIBDRAGON_PREVIEW` as an env var)


Dragorn421:
This correctly caught the new savedoodle example not setting `LIBDRAGON_PREVIEW = 2`:
https://github.com/DragonMinded/libdragon/actions/runs/32452479752/job/96683381751?pr=978#step:14:447

PR #979, opened 2026-08-22
rspq: make lowpri buffer handoff atomic

elohmeier:
Fix a race where highpri preemption can split the lowpri buffer-done signal from its jump, allowing the CPU to clear a buffer the RSP will resume into.

Stage the next buffer in an inactive call slot and use SWAP_BUFFERS for the atomic completion-and-transfer operation. Adds regression coverage.

### DIFF
--- a/include/rsp_queue.inc
+++ b/include/rsp_queue.inc
@@ -644,9 +644,10 @@ rspq_switch_highpri:
     #############################################################
     # RSPQCmd_SwapBuffers
     #
-    # Switch between lowpri and highpri or viceversa. This is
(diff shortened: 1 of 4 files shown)


rasky:
Thanks, this is a great contribution! The race condition was indeed present and never noticed / reported before. The fix is correctly executed and wittingly avoids changing the code by reusing the atomic buffer swap command. I've just reworded documentation in a way that I feel it's more natural for first-time readers that don't know the context of the bug being fixed here.

elohmeier:
Great, thanks for your review & merging!

PR #981, opened 2026-08-26
Downgrade binutils to 2.45

Dragorn421:
reported on discord here https://discord.com/channels/205520502922543113/974342113850445874/1541235212774805536 and here https://discord.com/channels/205520502922543113/974342113850445874/1542172027597885513

Note: I do not know if this downgraded version of binutils will work with the other versions


Dragorn421:
the github actions do not like that there are several "definitions" of `BINUTILS_V` (it greps for it and finding several lines breaks stuff)

do we just globally downgrade instead of mingw-only?

Dragorn421:
I elected to just downgrade binutils globally

PR #1001, opened 2026-09-13
video: add MPEG initialization and indexed playback helpers

forkymcforkface:
Expose MPEG initialization and playback helpers so applications can use the installed SDK without compiling `mpeg1.c` themselves or accessing private decoder state.

- `mpeg1_init` registers the codec and initializes its RSP overlay before the audio mixer starts. This preserves the initialization order needed by the tested application to avoid late-overlay synchronization hangs.
- The indexed-seek callback drains RSP work, rewinds, seeks to the indexed byte offset, and sets the decoder timeline. `mpeg1_seek_next` skips up to two leading orphan B pictures and aligns the returned reference frame.
- `mpeg1_next_frame_flush` emits the retained final reference once at EOF, including when the last decoded picture is B.

This is a draft for API review. The seek helper preserves an existing converter contract of at most two leading orphan B pictures; it is not general open-GOP recovery. The EOF behavior is exposed through the helper rather than changing every `video_next_frame` caller. General decoder-backed open/closed-GOP and final-I/P/B fixtures are still needed before treating this as a general codec fix.

Validation: SDK build and PhosphorOS production matrix pass. A modeled decoder/queue fixture checks bounded output filtering, EOF, the RSP fence, rewind/offset/timeline ordering, and callback precedence. Real 4 MB N64 video pause/seek, settings return, theme changes, and terminal-frame playback pass with zero recorded audio underruns or producer overruns. Real 8 MB hardware was not tested.

Regression harnesses remain outside this SDK change.


rasky:
The change is not motivated properly. The description assumes implicit context. Please write a description of the change.

PR #1003, opened 2026-09-13
wav64: close the file after preloading

sreuter:
A wav64 opened with `WAV64_STREAMING_NONE` has all of its samples in RDRAM by the time `internal_open` returns, so the file is never read again, but the descriptor was held until `wav64_close`.

Sound effects are the usual reason to preload, and a game tends to load them all at startup and keep them for its lifetime, so this can easily be a dozen descriptors sitting open and doing nothing.

On DFS that is only waste. On other filesystems the limit on simultaneously open files can be a lot lower, and then a handful of preloaded sounds is enough to exhaust it and unrelated opens start failing.

Only the descriptor wav64 opened itself is closed. One handed in through `wav64_loadfd` is left for the caller to manage.

Tested on hardware with 13 preloaded effects.


rasky:
Thanks!

PR #1004, opened 2026-09-13
Improve ROM config option validation for RTC and REGIONFREE boolean

networkfusion:
ROM authors that use ROM config options may add "N64_ROM_RTC = 0" to their makefiles and expect it to disable the RTC, but in actual fact this will include it. This adds the `filter-out` command so that it works as expected.

The same fix was performed for N64_ROM_REGIONFREE.

Other options may benefit from the same fix, and willing to improve those as well if needed.


rasky:
Please rebase this PR

PR #1010, opened 2026-09-20
fmv: honor fmv_parms_t.buffered_pics

jshield:
Adds \`buffered_pics\` to \`fmv_parms_t\`, passed through to \`video_open\`. 0 keeps the historical default of 8 so existing callers (including \`videoplayer\`) are unchanged.

Motivation: on a constrained N64 port (OpenLara64, 8 MiB RDRAM), level arenas are kept resident across title FMVs to avoid malloc churn and preflight failures. The 8-picture H.264 DPB needs a ~3 MiB contiguous hole that no longer exists once arenas are reserved. Setting \`buffered_pics\` to 2 keeps the FMV in its scratch pool instead of forcing callers to free level memory around every cutscene.

Verified: patched SDK builds and links \`openlara-n64.z64\` with \`buffered_pics = 2\`; host \`test-passport-title\` passes. No Ares runtime row yet.


jshield:
Superseded: upstream preview already landed the same knob as `fmv_parms_t.decode_ahead_pics` (defaults 4 base / 8 Expansion Pak). Switching our port to the upstream field instead.

